PR #6, opened 2019-12-06
Make component synchronization more robust

ahmed-mez:
### What does this PR do?

- Do not create Agent DaemonSet / CLC Runner deployments until the Cluster Agent has at least 1 available replica

### Motivation

Better component synchronization: Make sure the Cluster Agent is ready before creating Agent DaemonSet / CLC Runner deployments

### Additional Notes

### DIFF
--- a/pkg/controller/datadogagentdeployment/clusteragent.go
+++ b/pkg/controller/datadogagentdeployment/clusteragent.go
@@ -60,15 +60,21 @@ func (r *ReconcileDatadogAgentDeployment) reconcileClusterAgent(logger logr.Logg
 			if errors.IsNotFound(err) {
 				logger.Info("ClusterAgent deployment not found", "name", nsName.Name, "namespace", nsName.Namespace)
 				// Create and attach a ClusterAgentDeployment
-				var result reconcile.Result
-				result, err = r.createNewClusterAgentDeployment(logger, dad, newStatus)
-				return result, err
+				return r.createNewClusterAgentDeployment(logger, dad, newStatus)
 			}
 			return reconcile.Result{}, err
 		}
 
-		result, err := r.updateClusterAgentDeployment(logger, dad, clusterAgentDeployment, newStatus)
-		return result, err
+		if result, err = r.updateClusterAgentDeployment(logger, dad, clusterAgentDeployment, newStatus); err != nil {
+			return result, err
+		}
+
+		// Make sure we have at least one Cluster Agent available replica
+		if clusterAgentDeployment.Status.AvailableReplicas == 0 {
+			return reconcile.Result{RequeueAfter: defaultRequeuPeriod}, fmt.Errorf("cluster agent deployment is not ready yet: 0 pods available out of %d", clusterAgentDeployment.Status.Replicas)
+		}
+
+		return reconcile.Result{}, nil
 	}
 	return reconcile.Result{}, nil
 }
--- a/pkg/apis/datadoghq/v1alpha1/test/new.go
+++ b/pkg/apis/datadoghq/v1alpha1/test/new.go
@@ -169,9 +169,10 @@ func NewExtendedDaemonSet(ns, name string, options *NewExtendedDaemonSetOptions)
 
 // NewDeploymentOptions set of option for the Deployment creation
 type NewDeploymentOptions struct {
-	CreationTime *time.Time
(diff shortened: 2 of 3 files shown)


pkg/controller/datadogagentdeployment/clusteragent.go
```diff
@@ -68,7 +68,15 @@ func (r *ReconcileDatadogAgentDeployment) reconcileClusterAgent(logger logr.Logg
 		}
 
 		result, err := r.updateClusterAgentDeployment(logger, dad, clusterAgentDeployment, newStatus)
-		return result, err
+		if err != nil {
+			return result, err
+		}
+
+		if clusterAgentDeployment.Status.ReadyReplicas == 0 {
+			return reconcile.Result{Requeue: true}, fmt.Errorf("cluster agent deployment is not ready yet: 0 pods ready out of %d", clusterAgentDeployment.Status.Replicas)
```
ahmed-mez: would it be better to use `RequeueAfter` instead ?

clamoriniere: I think we don't need to requeue at all since we are watching the ClusterAgent deployment resource events.

pkg/controller/datadogagentdeployment/service.go
```diff
@@ -40,7 +40,21 @@ func (r *ReconcileDatadogAgentDeployment) manageClusterAgentService(logger logr.
...
+
+	// Make sure the corresponding Endpoints object is created
+	endpoints := &corev1.Endpoints{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Namespace: dad.Namespace, Name: serviceName}, endpoints)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return reconcile.Result{Requeue: true}, fmt.Errorf("cluster agent service endpoints object %s/%s is not created yet", dad.Namespace, serviceName)
+		}
+		return reconcile.Result{}, err
+	}
```
clamoriniere: I think this is not necessary, since this should be handle automatically by the kubernetes controllers. Also you already test it in the `reconcileClusterAgent` function

pkg/controller/datadogagentdeployment/clusteragent.go
ahmed-mez: makes sense

pkg/controller/datadogagentdeployment/service.go
ahmed-mez: correct 👍

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
```diff
@@ -162,6 +162,15 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Watch for changes to secondary resource Endpoints and requeue the owner DatadogAgentDeployment
+	err = c.Watch(&source.Kind{Type: &corev1.Endpoints{}}, &handler.EnqueueRequestForOwner{
```
ahmed-mez: would it be possible to watch Endpoints only when the Cluster Agent is enabled?

ahmed-mez:
thanks for the review, yes I believe checking the Deployment status is enough

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
clamoriniere: it is not trivial to do, but it could be done.

pkg/controller/datadogagentdeployment/clusteragent.go
```diff
@@ -27,6 +27,8 @@ import (
 	"github.com/DataDog/datadog-operator/pkg/controller/utils/comparison"
 )
 
+const availabilityRequeueAfterDuration = 5 * time.Second
```
clamoriniere: maybe we can consider to reuse the default requeue duration: https://github.com/DataDog/datadog-operator/blob/24eb8b44b92fa948a6db5a4ffa5105430fa2181b/pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go#L41

ahmed-mez: indeed, thanks

PR #8, opened 2019-12-10
Send component deployment metrics to Datadog

ahmed-mez:
### What does this PR do?

- Sends metrics and events directly to Datadog to monitor the different Agent components deployment status
- Define `openmetrics` check AD annotations on `datadog-operator`

### Motivation

Metrics could be used to create alerts if `DatadogAgentDeployment` failed

### Additional Notes


pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
```diff
@@ -53,7 +56,19 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileDatadogAgentDeployment{client: mgr.GetClient(), scheme: mgr.GetScheme(), recorder: mgr.GetEventRecorderFor("DatadogAgentDeployment")}
+	reconciler := &ReconcileDatadogAgentDeployment{
+		client:   mgr.GetClient(),
+		scheme:   mgr.GetScheme(),
+		recorder: mgr.GetEventRecorderFor("DatadogAgentDeployment"),
+		datadogMetricsForwarder: &datadog.MetricsForwarder{
+			RetryInterval:        10 * time.Second,
```
xornivore: Curious if it is worth exposing these settings via pflag config parameter?

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
```diff
@@ -263,3 +279,86 @@ func (r *ReconcileDatadogAgentDeployment) updateStatusIfNeeded(logger logr.Logge
...
+// it starts sending deployment metrics once the connection is validated
+func (r *ReconcileDatadogAgentDeployment) sendDatadogMetrics(ctx context.Context) {
+	retryTicker := time.NewTicker(r.datadogMetricsForwarder.RetryInterval)
+RetryLoop:
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-retryTicker.C:
+			dad, err := r.listDatadogAgentDeployment()
```
xornivore: If you pass the `ctx` down to listDatadogAgentDeployment it might help with getting rid of `context.TODO()` in that func.

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
ahmed-mez: yes, I believe the operator configurability must be discussed at this point

ahmed-mez: regarding `RetryInterval` it might even best if we implement an exponential backoff strategy, like retry after 10 sec, 20 sec, 30 sec etc ..

pkg/controller/utils/datadog/datadog.go
```diff
@@ -0,0 +1,170 @@
...
+// SetGlobalTags adds global tags to the MetricsForwarder from DatadogAgentDeployment
+func (dd *MetricsForwarder) SetGlobalTags(dad *datadoghqv1alpha1.DatadogAgentDeployment) {
+	if dad == nil {
+		return
+	}
+	tags := []string{}
+	if dad.Spec.ClusterName != "" {
+		tags = append(tags, fmt.Sprintf(clusterNameTagFormat, dad.Spec.ClusterName))
+	}
+	// TODO add more tags
```
ahmed-mez: any suggestions of tags that we can add ?

pkg/controller/datadogagentdeployment/secret.go
```diff
@@ -149,3 +150,40 @@ func needClusterAgentSecret(dad *datadoghqv1alpha1.DatadogAgentDeployment) bool
...
+		appKey = dad.Spec.Credentials.AppKey
+	} else {
+		secretName := getAppKeySecretName(dad)
+		secret := &corev1.Secret{}
+		err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: dad.Namespace, Name: secretName}, secret)
+		if err != nil {
+			return "", "", err
+		}
+		appKey = string(secret.Data[datadoghqv1alpha1.DefaultAPPKeyKey])
+	}
```
clamoriniere: maybe you can refactor the code with a sub function that take in parameters: `dad` and the function `getAppKeySecretName ` or `getAPIKeySecretName `

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
```diff
@@ -263,3 +279,87 @@ func (r *ReconcileDatadogAgentDeployment) updateStatusIfNeeded(logger logr.Logge
 
 	return result, currentError
 }
+
+// sendDatadogMetrics establishes a connection with the Datadog API
+// it starts sending deployment metrics once the connection is validated
+func (r *ReconcileDatadogAgentDeployment) sendDatadogMetrics(ctx context.Context) {
+	retryTicker := time.NewTicker(r.datadogMetricsForwarder.RetryInterval)
+RetryLoop:
```
clamoriniere: I'm sure we can do it without it.

have you look into the package: `k8s.io/apimachinery/pkg/util/wait` it provide some usefull function that implement for you the ticker, stop. force instance the `Poll` function seams to do the work
https://github.com/kubernetes/apimachinery/blob/master/pkg/util/wait/wait.go#L313

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
```diff
@@ -53,7 +56,19 @@ func Add(mgr manager.Manager) error {
...
+		scheme:   mgr.GetScheme(),
+		recorder: mgr.GetEventRecorderFor("DatadogAgentDeployment"),
+		datadogMetricsForwarder: &datadog.MetricsForwarder{
+			RetryInterval:        10 * time.Second,
+			SendMetricsInterval:  15 * time.Second,
+			RefreshCredsInterval: 15 * time.Minute,
+			MetricsPrefix:        "datadog.operator",
+		},
+	}
+	go reconciler.sendDatadogMetrics(context.Background())
```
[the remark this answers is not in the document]
ahmed-mez: neat! thanks

ahmed-mez:
@clamoriniere @xornivore thanks for the comments, the PR is ready for a second review

pkg/controller/utils/datadog/datadog.go
```diff
@@ -0,0 +1,441 @@
...
+	var metricValue float64
+
+	// Agent deployment metrics
+	if status.Agent != nil {
+		if status.Agent.Available == status.Agent.Desired {
+			metricValue = deploymentSuccessValue
+		} else {
+			metricValue = deploymentFailureValue
+		}
+		tags := append(dd.tags, fmt.Sprintf(stateTagFormat, string(status.Agent.State)))
```
xornivore: Minor nit - since this gets repeated below, could add a helper here - e.g. `tagsWithState(..) []string`, then `tags := dd.tagsWithState(status.Agent.State)`

pkg/controller/utils/datadog/datadog.go
```diff
@@ -0,0 +1,441 @@
...
+
+// sendStatusEvents forwards events for each component deployment (agent, clusteragent, clustercheck runner)
+// based on the status of DatadogAgentDeployment
+func (dd *MetricsForwarder) sendStatusEvents(status *datadoghqv1alpha1.DatadogAgentDeploymentStatus) error {
+	if status == nil {
+		return errors.New("nil status")
+	}
+
+	// Agent deployment events
+	if status.Agent != nil && !dd.componentEnabled[agentName] {
```
xornivore: I wonder if there's a way to reorganize these if statements for better readability. There's a similar pattern but the types are different.

pkg/controller/utils/datadog/datadog.go
```diff
@@ -0,0 +1,441 @@
...
+	sendMetricsInterval time.Duration
+	metricsPrefix       string
+	tags                []string
+	componentEnabled    map[string]bool
+	delegator           delegatedAPI
+}
+
+// delegatedAPI is used for testing purpose, it serves for mocking the Datadog API
+type delegatedAPI interface {
+	delegatedSendDeploymentMetric(float64, string, []string) error
```
xornivore: I wonder if maybe it's worth dropping `delegated` on the method names?

ahmed-mez: `MetricsForwarder` already has a method named `sendDeploymentMetric`

xornivore: Ah I see, nm then 👍

pkg/controller/utils/datadog/datadog.go
ahmed-mez: good point, maybe we can add a method (`detectEvent(componentIsNil bool, componentName, componentEventName string)` for example) and we call it this way `dd.detectEvent(status.<component> == nil, <component name>, <component event name>)` 

This way the logic could be shared between the 3 components

ahmed-mez: more simply, we can just pass `componentIsNil` to `sendDeploymentEvent`

ahmed-mez: but adding logic to `delegatedSendDeploymentEvent` will prevent us from testing and asserting that we're calling the datadog api using the correct values 🤔

ahmed-mez: nm, could be solved using this method
```golang
func (dd *MetricsForwarder) getDeploymentEventType(isComponentNil bool, componentName string) *string {
	if !isComponentNil && !dd.componentEnabled[agentName] {
		dd.componentEnabled[agentName] = true
		return &deploymentCreatedEventType
	}
	if isComponentNil && dd.componentEnabled[agentName] {
		dd.componentEnabled[agentName] = false
		return &deploymentDeletedEventType
	}
	return nil
}
```

ahmed-mez: called this way
```golang
	if eventType := dd.getDeploymentEventType(status.Agent == nil, agentName); eventType != nil {
		if err := dd.sendDeploymentEvent(agentEventName, *eventType); err != nil {
			return err
		}
	}
```

pkg/controller/utils/datadog/datadog.go
```diff
@@ -275,42 +283,44 @@ func (dd *MetricsForwarder) sendStatusEvents(status *datadoghqv1alpha1.DatadogAg
...
 
 	// Agent deployment events
-	if status.Agent != nil && !dd.componentEnabled[agentName] {
-		if err := dd.sendDeploymentEvent(agentEventName, deploymentCreatedEventType); err != nil {
-			return err
-		}
-		dd.componentEnabled[agentName] = true
-	} else if status.Agent == nil && dd.componentEnabled[agentName] {
-		if err := dd.sendDeploymentEvent(agentEventName, deploymentDeletedEventType); err != nil {
+	if eventType := dd.getDeploymentEventType(status.Agent == nil, agentName); eventType != nil {
```
xornivore: 👌

pkg/controller/utils/datadog/datadog.go
```diff
@@ -0,0 +1,451 @@
...
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+const (
+	defaultSendMetricsInterval  = 15 * time.Second
+	defaultMetricsNamespace     = "datadog.operator"
+	gaugeType                   = "gauge"
+	deploymentSuccessValue      = 1.0
+	deploymentFailureValue      = 0.0
+	deploymentMetricFormat      = "%s.%s.deploymentsuccess"
```
clamoriniere:
```suggestion
	deploymentMetricFormat      = "%s.%s.deployment.success"
```
what do you think about this suggestion?

pkg/controller/utils/datadog/datadog.go
```diff
@@ -0,0 +1,451 @@
...
+}
+
+// sendStartupEvent posts the operator startup event
+// it must be called only once, when the operator starts
+func (dd *MetricsForwarder) sendStartupEvent() error {
+	event := &api.Event{
+		Time:      api.Int(int(time.Now().Unix())),
+		Title:     api.String("Datadog Operator has started"),
+		EventType: api.String("Datadog Operator Startup"),
+	}
```
clamoriniere: we should add `SourceType ` too for a better integration in the datadog website

README.md
```diff
@@ -39,6 +39,34 @@ Using the Datadog Operator requires the following prerequisites:
 
 See the [Quick Start tutorial](docs/quick-start.md).
 
+## Metrics and Events
+
+The Datadog Operator sends metrics and events to Datadog to monitor the Datadog Agent components deployment in the cluster.
+
+### Metrics
+
+|metric name   |metric type   |description   |
```
l0k0ms:
```suggestion
|Metric name   |Metric type   |Description   |
```

README.md
```diff
@@ -39,6 +39,34 @@ Using the Datadog Operator requires the following prerequisites:
...
+|---|---|---|
+|datadog.operator.agent.deployment.success   |gauge   |1 if the de desired number of Agent replicas equals the number of available Agent pods, 0 otherwise   |
```
l0k0ms:
```suggestion
| `datadog.operator.agent.deployment.success`   | gauge   | `1` if the de desired number of Agent replicas equals the number of available Agent pods, `0` otherwise.   |
```

README.md
```diff
@@ -39,6 +39,34 @@ Using the Datadog Operator requires the following prerequisites:
...
+|datadog.operator.clusteragent.deployment.success   |gauge   |1 if the de desired number of Cluster Agent replicas equals the number of available Cluster Agent pods, 0 otherwise   |
```
l0k0ms:
```suggestion
| `datadog.operator.clusteragent.deployment.success` | gauge   | `1` if the de desired number of Cluster Agent replicas equals the number of available Cluster Agent pods, `0` otherwise.   |
```

README.md
```diff
@@ -39,6 +39,34 @@ Using the Datadog Operator requires the following prerequisites:
...
+|datadog.operator.clustercheckrunner.deployment.success   |gauge   |1 if the de desired number of Cluster Check Runner replicas equals the number of available Cluster Check Runner pods, 0 otherwise   |
```
l0k0ms:
```suggestion
| `datadog.operator.clustercheckrunner.deployment.success`   | gauge   | `1` if the de desired number of Cluster Check Runner replicas equals the number of available Cluster Check Runner pods, `0` otherwise.   |
```

README.md
```diff
@@ -39,6 +39,34 @@ Using the Datadog Operator requires the following prerequisites:
...
+
+**Note:** The Datadog API and APP keys are required to forward metrics to Datadog, they must be provided in the `credentials` field in the Custom Resource definiton.
```
l0k0ms: you should add a link to the Datadog API key page (US site)

README.md
```diff
@@ -39,6 +39,34 @@ Using the Datadog Operator requires the following prerequisites:
...
+
+The Datadog Operator exposes Golang and Controller metrics in OpenMetrics format. For now they can be collected using the [OpenMetrics integration](https://docs.datadoghq.com/integrations/openmetrics/). A Datadog integration will be available soon.
```
l0k0ms:
```suggestion
The Datadog Operator exposes Golang and Controller metrics in OpenMetrics format. For now they can be collected using the [OpenMetrics integration](https://docs.datadoghq.com/integrations/openmetrics/). A Datadog integration will be available in the future.
```

README.md
```diff
@@ -39,6 +39,34 @@ Using the Datadog Operator requires the following prerequisites:
...
+
+The OpenMetrics check is activated by default via [autodiscovery annotations](./chart/datadog-operator/templates/deployment.yaml) and will be scheduled by the Agent running on the same node as the Datadog Operator Pod.
```
l0k0ms:
```suggestion
The OpenMetrics check is activated by default via [autodiscovery annotations](./chart/datadog-operator/templates/deployment.yaml) and is scheduled by the Agent running on the same node as the Datadog Operator Pod.
```

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
```diff
@@ -43,6 +45,7 @@ const (
 
 func init() {
 	pflag.BoolVarP(&supportExtendedDaemonset, "supportExtendedDaemonset", "", false, "support ExtendedDaemonset")
+	pflag.DurationVar(&metricsRetryInterval, "metricsRetryInterval", 15*time.Second, "Datadog API connection retry duration")
```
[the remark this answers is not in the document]
ahmed-mez: this is actually the retry interval for initializing the datadog api connection, I'm going to remove it based on our discussion

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
```diff
@@ -53,7 +56,17 @@ func Add(mgr manager.Manager) error {
...
-	return &ReconcileDatadogAgentDeployment{client: mgr.GetClient(), scheme: mgr.GetScheme(), recorder: mgr.GetEventRecorderFor("DatadogAgentDeployment")}
+	reconciler := &ReconcileDatadogAgentDeployment{
+		client:   mgr.GetClient(),
+		scheme:   mgr.GetScheme(),
+		recorder: mgr.GetEventRecorderFor("DatadogAgentDeployment"),
+	}
+
+	if err := mgr.Add(datadog.NewMetricsForwarder(reconciler.client, metricsRetryInterval)); err != nil {
+		log.Error(err, "cannot add Datadog metrics forwarder")
+	}
```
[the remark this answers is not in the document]
ahmed-mez: we'll need to pass `mgr.GetClient()` as argument instead of `reconciler.client`

pkg/controller/utils/datadog/datadog.go
ahmed-mez: agreed

pkg/controller/utils/datadog/datadog.go
ahmed-mez: good catch thanks

README.md
```diff
@@ -39,6 +39,34 @@ Using the Datadog Operator requires the following prerequisites:
...
+## Metrics and Events
+
+The Datadog Operator sends metrics and events to Datadog to monitor the Datadog Agent components deployment in the cluster.
+
+### Metrics
+
+|Metric name   |Metric type   |Description   |
+|---|---|---|
+| `datadog.operator.agent.deployment.success`   | gauge   | `1` if the de desired number of Agent replicas equals the number of available Agent pods, `0` otherwise.   |
+| `datadog.operator.clusteragent.deployment.success`   | gauge   | `1` if the de desired number of Cluster Agent replicas equals the number of available Cluster Agent pods, `0` otherwise.   |
```
l0k0ms:
```suggestion
| `datadog.operator.clusteragent.deployment.success`   | gauge   | `1` if the desired number of Cluster Agent replicas equals the number of available Cluster Agent pods, `0` otherwise.   |
```

README.md
```diff
@@ -39,6 +39,34 @@ Using the Datadog Operator requires the following prerequisites:
...
+| `datadog.operator.clustercheckrunner.deployment.success`   | gauge   | `1` if the de desired number of Cluster Check Runner replicas equals the number of available Cluster Check Runner pods, `0` otherwise.   |
```
l0k0ms:
```suggestion
| `datadog.operator.clustercheckrunner.deployment.success`   | gauge   | `1` if the desired number of Cluster Check Runner replicas equals the number of available Cluster Check Runner pods, `0` otherwise.   |
```

README.md
```diff
@@ -39,6 +39,34 @@ Using the Datadog Operator requires the following prerequisites:
...
 
+## Metrics and Events
+
+The Datadog Operator sends metrics and events to Datadog to monitor the Datadog Agent components deployment in the cluster.
+
+### Metrics
+
+|Metric name   |Metric type   |Description   |
+|---|---|---|
+| `datadog.operator.agent.deployment.success`   | gauge   | `1` if the de desired number of Agent replicas equals the number of available Agent pods, `0` otherwise.   |
```
l0k0ms:
```suggestion
| `datadog.operator.agent.deployment.success`   | gauge   | `1` if the desired number of Agent replicas equals the number of available Agent pods, `0` otherwise.   |
```

README.md
ahmed-mez: good catch

.github/CODEOWNERS
```diff
@@ -0,0 +1,9 @@
+# See https://help.github.com/articles/about-codeowners/ for syntax
```
clamoriniere: 👍

pkg/controller/utils/datadog/forwarders_manager.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+}
+
+// stopAllForwarders stops the running metricsForwarder goroutines
+func (f *ForwardersManager) stopAllForwarders() {
+	f.Lock()
+	defer f.Unlock()
+	for id, forwarder := range f.forwarders {
+		log.Info("Stopping metrics forwarder", "ID", id)
+		forwarder.stop <- struct{}{}
```
clamoriniere: you can also close the channel to stop the forwarder

pkg/controller/utils/datadog/forwarders_manager.go
```diff
@@ -0,0 +1,80 @@
...
+// unregisterForwarder deletes a given metricsForwarder
+func (f *ForwardersManager) unregisterForwarder(id string) {
+	f.Lock()
+	defer f.Unlock()
+	if _, found := f.forwarders[id]; !found {
+		log.Error(fmt.Errorf("%s not found", id), "cannot unregister metrics forwarder")
+		return
+	}
+	log.Info("Metrics forwarder unregistered", "ID", id)
+	delete(f.forwarders, id)
```
clamoriniere: I think you should stop it before the `delete`

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
```diff
@@ -48,12 +50,27 @@ func init() {
...
+		scheme:     mgr.GetScheme(),
+		recorder:   mgr.GetEventRecorderFor("DatadogAgentDeployment"),
+		forwarders: datadog.NewForwardersManager(mgr.GetClient()),
+	}
+	return reconciler, mgr.Add(reconciler.forwarders)
+}
+
+type metricForwardersManager interface {
+	Register(types.NamespacedName)
+	Start(stop <-chan struct{}) error
```
clamoriniere: should we have Unregister() method too?

clamoriniere: Maybe `Start()` method is not necessary in this interface

pkg/controller/utils/datadog/forwarders_manager.go
ahmed-mez: I don't think so, this is only called when the forwarder sends its ID to the manager and breaks it's infinite loop (when its `DatadogAgentDeployment` is deleted)

in the forwarder
```golang
					// DatadogAgentDeployment was deleted
					// inform the ForwardersManager before shutting down the metricsForwarder
					managerChan <- mf.id
					return
````

in the manager 
```golang
		case id := <-f.stoppedForwarder:
			f.unregisterForwarder(id)
```

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
```diff
@@ -202,6 +220,8 @@ func (r *ReconcileDatadogAgentDeployment) Reconcile(request reconcile.Request) (
 		return reconcile.Result{}, err
 	}
 
+	r.forwarders.Register(request.NamespacedName)
```
clamoriniere: Can we "Register" the resource only once? 
Maybe by call register only on Add() event in the watcher

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
ahmed-mez: currently we're letting each `metricsForwarder` handling this with the `ForwarderManager`.  the forwarder sends its id to the managetr when the CR is deleted, no need to have this in the reconcile loop

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
ahmed-mez: makes perfect sense 👍

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
ahmed-mez: it's necessary because we need the `reconcile.forwarders` to implement the `Runnable` interface here

```golang
mgr.Add(reconciler.forwarders)
```

clamoriniere: ok fair enough.
can be change by
```golang
func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
        forwarders := datadog.NewForwardersManager(mgr.GetClient())
	reconciler := &ReconcileDatadogAgentDeployment{
		client:     mgr.GetClient(),
		scheme:     mgr.GetScheme(),
		recorder:   mgr.GetEventRecorderFor("DatadogAgentDeployment"),
		forwarders: forwarders,
	}
	return reconciler, mgr.Add(forwarders)
}
```

pkg/controller/utils/datadog/metrics_forwarder.go
```diff
@@ -0,0 +1,538 @@
...
+	// Global tags need to be set only once
+	mf.initGlobalTags()
+
+	// wait.PollInfinite is blocking until mf.connectToDatadogAPI returns true
+	if err := wait.PollInfinite(mf.retryInterval, mf.connectToDatadogAPI); apierrors.IsNotFound(err) {
+		// DatadogAgentDeployment was deleted
+		// inform the ForwardersManager before shutting down the metricsForwarder
+		managerChan <- mf.id
+		return
+	}
```
clamoriniere: I'm not a big fan of having the metricForwarder responsible to stop himself.
I would prefer to have a `Unregister()` method available in the MetricsForwarders interface
and here: ```func (mf *metricsForwarder) start(stop <-chan struct{}) ```

ahmed-mez: I see your point

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
```diff
@@ -48,24 +52,56 @@ func init() {
...
+}
+
+type metricForwardersManager interface {
+	Register(types.NamespacedName)
+	Unregister(types.NamespacedName)
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
-func add(mgr manager.Manager, r reconcile.Reconciler) error {
+func add(mgr manager.Manager, r reconcile.Reconciler, registerFunc func(types.NamespacedName)) error {
```
clamoriniere: did you add this new parameters for unit-test purpose?

ahmed-mez: no, `r` is an interface not a struct, I cannot access the `forwarders` attribute to use its `Register` method

pkg/controller/utils/datadog/metrics_forwarder.go
```diff
@@ -0,0 +1,539 @@
...
+}
+
+// tagsWithState used to append the state tag
+func (mf *metricsForwarder) tagsWithState(state string) []string {
+	return append(mf.globalTags, append(mf.tags, fmt.Sprintf(stateTagFormat, state))...)
+}
+
+// sendStatusEvents forwards events for each component deployment (agent, clusteragent, clustercheck runner)
+// based on the status of DatadogAgentDeployment
+func (mf *metricsForwarder) sendStatusEvents(status *datadoghqv1alpha1.DatadogAgentDeploymentStatus) error {
```
clamoriniere: maybe we should avoid sending events for now.

pkg/controller/utils/datadog/metrics_forwarder.go
```diff
@@ -0,0 +1,539 @@
...
+	secret := &corev1.Secret{}
+	err := mf.k8sClient.Get(context.TODO(), types.NamespacedName{Namespace: dad.Namespace, Name: secretName}, secret)
+	if err != nil {
+		return "", err
+	}
+	return string(secret.Data[dataKey]), nil
+}
+
+// updateStatusIfNeeded updates the Datadog metrics forwarding status in the DatadogAgentDeployment status
+func (mf *metricsForwarder) updateStatusIfNeeded(dad *datadoghqv1alpha1.DatadogAgentDeployment, err error) error {
```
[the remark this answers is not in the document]
ahmed-mez: actually it's not the same function, not the same signature (return values), not the same arguments passed to `UpdateDatadogAgentDeploymentStatusCondition`  and not the same logger

pkg/controller/utils/datadog/forwarders_manager.go
```diff
@@ -0,0 +1,83 @@
...
+	log.Info("Metrics forwarder unregistered", "ID", id)
+}
+
+// stopAllForwarders stops the running metricsForwarder goroutines
+func (f *ForwardersManager) stopAllForwarders() {
+	f.Lock()
+	defer f.Unlock()
+	for id, forwarder := range f.forwarders {
+		log.Info("Stopping metrics forwarder", "ID", id)
+		forwarder.stop()
```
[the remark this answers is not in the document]
ahmed-mez: good catch, actually the `stop()` only closes the stop channel (non blocking), so I added a wait group so that the forwarder `start()` method decrements it before it returns.

pkg/controller/utils/datadog/metrics_forwarder.go
```diff
@@ -0,0 +1,434 @@
...
+func (mf *metricsForwarder) updateCredsIfNeeded(apiKey, appKey string) error {
+	if mf.keysHash != hashKeys(apiKey, appKey) {
+		return mf.initAPIClient(apiKey, appKey)
+	}
+	return nil
+}
+
+// validateCreds returns validates the creds by querying the Datadog API
+func (mf *metricsForwarder) validateCreds(apiKey, appKey string) (*api.Client, error) {
+	return mf.delegator.delegatedValidateCreds(apiKey, appKey)
```
clamoriniere: do we still need this `delegator` indirection? or can we call directly: 
```suggestion
	return mf.delegatedValidateCreds(apiKey, appKey)
```

PR #12, opened 2019-12-18
Add proper status in `get dad` display

vboulineau:
### What does this PR do?

Actually compute status and displays it in `kubectl get dad` output.
Currently it's implemented for the Deployments and (Extended)DaemonSets we have.

```
NAME            ACTIVE   AGENT                 CLUSTER-AGENT     CLUSTER-CHECKS-RUNNER   AGE
datadog-agent   True     Progressing (1/0/1)   Running (2/2/2)   Progressing (2/0/2)     24h
```

I've changed existing statuses and put all supported objects to have the same status:
- Progressing (is deployed, but no pod ready)
- Running (is deployed, with at least 1 pod ready)
- Updating (rolling update on-going, i.e. number of requested != number of uptodate PODs)
- Failed (creation of the target resource failed or Kubernetes reported a failed state - only for Deployments)

The numbers are Requested / Ready / Uptodate.

### Additional Notes

Flagging the PR as WIP as we need to agree on the different statuses and conditions to display them.
So feel free to make proposals if you don't agree with the statuses/conditions!

### DIFF
--- a/pkg/controller/datadogagentdeployment/utils.go
+++ b/pkg/controller/datadogagentdeployment/utils.go
@@ -14,6 +14,7 @@ import (
 
 	datadoghqv1alpha1 "github.com/DataDog/datadog-operator/pkg/apis/datadoghq/v1alpha1"
 	"github.com/DataDog/datadog-operator/pkg/kubernetes"
+	edsdatadoghqv1alpha1 "github.com/datadog/extendeddaemonset/pkg/apis/datadoghq/v1alpha1"
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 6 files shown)


pkg/controller/datadogagentdeployment/utils.go
```diff
@@ -1018,10 +1019,74 @@ func getPodAffinity(affinity *corev1.Affinity, labelValue string) *corev1.Affini
...
+	var deploymentState datadoghqv1alpha1.DatadogAgentDeploymentState
+	if dsStatus.UpToDate != dsStatus.Desired {
+		deploymentState = datadoghqv1alpha1.DatadogAgentDeploymentStateUpdating
+	} else if dsStatus.Ready == 0 {
+		deploymentState = datadoghqv1alpha1.DatadogAgentDeploymentStateProgressing
+	} else {
+		deploymentState = datadoghqv1alpha1.DatadogAgentDeploymentStateRunning
+	}
+
+	dsStatus.State = fmt.Sprintf("%v (%d/%d/%d)", deploymentState, dsStatus.Desired, dsStatus.Ready, dsStatus.UpToDate)
```
clamoriniere: could you add the pods "available" count too?

pkg/controller/datadogagentdeployment/utils.go
```diff
@@ -1018,10 +1019,74 @@ func getPodAffinity(affinity *corev1.Affinity, labelValue string) *corev1.Affini
...
```
clamoriniere: could you add the pods "available" count too?

pkg/controller/datadogagentdeployment/utils.go
vboulineau: Can do, though I believe we should pick ready or available, putting 4 indicators may be a big too much. I think ready and available are quite close so any of the 2 should be enough, no?

clamoriniere: I thought it made sense compared to the current daemonset display.
```
k get ds
NAME            DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR   AGE
datadog-agent   1638      1633      1615    1616         1622        <none>          306d
```
What I want to high light with this info is pod that start properly (ready at least one) but after in crashloop. In this case `available` will be greater than `ready`

clamoriniere: we can start with what you did. If people needs more info they can look at the DaemonSet

vboulineau: > I thought it made sense compared to the current daemonset display.
> 
> ```
> k get ds
> NAME            DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR   AGE
> datadog-agent   1638      1633      1615    1616         1622        <none>          306d
> ```
> 
> What I want to high light with this info is pod that start properly (ready at least one) but after in crashloop. In this case `available` will be greater than `ready`

That's why I kept ready over available, I guess it gives a better idea if everything is working smoothly or not.

pkg/controller/datadogagentdeployment/utils.go
```diff
@@ -1018,10 +1019,74 @@ func getPodAffinity(affinity *corev1.Affinity, labelValue string) *corev1.Affini
...
+	dsStatus.Available = eds.Status.Available
+	dsStatus.UpToDate = eds.Status.UpToDate
+
+	var deploymentState datadoghqv1alpha1.DatadogAgentDeploymentState
+	if dsStatus.UpToDate != dsStatus.Desired {
+		deploymentState = datadoghqv1alpha1.DatadogAgentDeploymentStateUpdating
+	} else if dsStatus.Ready == 0 {
+		deploymentState = datadoghqv1alpha1.DatadogAgentDeploymentStateProgressing
+	} else {
+		deploymentState = datadoghqv1alpha1.DatadogAgentDeploymentStateRunning
```
clamoriniere: what about using a `switch {}` instead of the `if else if else` ?

I think this logic can be added directly to the EDS controller. like this you just need to get the state value. 

Also here we are loosing the information when the EDS is in a Canary state. Is this wanted? Do you think it is not a relevant information?

vboulineau: Fine with a `switch{}`. We could add it to the EDS, but then we need to keep the logic for normal DS, so not sure it's worth it?

A `Canary` state existed but was never set in the code, so I am not sure we are loosing something :D. Currently it will be flagged as `Updating`, though if there is a specific way to detect `Canary` state in EDS, I can add it there.

clamoriniere: About the "Canary" state: the intension was here :) You can consider that was a bug.

To detect thea canary state you have in the [EDS.status.canary](https://github.com/DataDog/extendeddaemonset/blob/master/pkg/apis/datadoghq/v1alpha1/extendeddaemonset_types.go#L103) section that appear when a canary is ongoing. Else the [EDS.status.state](https://github.com/DataDog/extendeddaemonset/blob/master/pkg/apis/datadoghq/v1alpha1/extendeddaemonset_types.go#L101) should reflect it too

vboulineau: Added!

PR #14, opened 2019-12-20
Add possibility to provide custom ConfigMap for Agent

vboulineau:
### What does this PR do?

Add the possibility to add an inline YAML content representing the datadog.yaml file for the Agent.

### Additional Notes

Flagging as WIP as there might be several points to discuss:
- Currently only supports Agent (not DCA, CLC)
- Content is inlined in the DAD object, as I expect that most of the time, it will remain quite small
- Currently we only check the content to be YAML valid, nothing more, maybe we could improve but we would need to have access to Agent's config parser.


pkg/controller/datadogagentdeployment/agent.go
```diff
@@ -352,3 +350,34 @@ func newDaemonsetObjectMetaData(agentdeployment *datadoghqv1alpha1.DatadogAgentD
...
+	return fmt.Sprintf("%s-datadog-yaml", dad.Name)
+}
+
+func buildAgentConfigurationConfigMap(dad *datadoghqv1alpha1.DatadogAgentDeployment) (*corev1.ConfigMap, error) {
+	if dad.Spec.Agent.CustomConfig == nil {
+		return nil, nil
+	}
+
+	// Validate that user input is valid YAML
+	// Maybe later we can implement that directly verifies against Agent configuration?
```
xornivore: It's probably worth filing an issue for this right away.

pkg/apis/datadoghq/v1alpha1/datadogagentdeployment_types.go
```diff
@@ -124,6 +121,11 @@ type DatadogAgentDeploymentSpecAgentSpec struct {
 	// See https://docs.datadoghq.com/agent/guide/agent-configuration-files/?tab=agentv6 for more details.
 	// +optional
 	Checksd *ConfigDirSpec `json:"checksd,omitempty"`
+
+	// Allow to put custom configuration for the agent, corresponding to the datadog.yaml config file
+	// See https://docs.datadoghq.com/agent/guide/agent-configuration-files/?tab=agentv6 for more details.
+	// +optional
+	CustomConfig *string `json:"customConfig,omitempty"`
```
xornivore: It might be worth considering `Agent` (`DatadogAgentDeploymentSpecAgentSpec`) as the place to put this new configuration. Quite possibly the `Confd` and `Checksd` fields should be there as well. @clamoriniere what do you think?

vboulineau: All of them are already in `DatadogAgentDeploymentSpecAgentSpec`, or am I missing something?

pkg/controller/datadogagentdeployment/agent.go
vboulineau: That is something I'm wondering about. It will for sure require a dependency (maybe not code, but still) on the agent, and as we can deploy "any" version of the agent with the operator, it might not be easy, so I'm not sure if we really want to do it or not

pkg/apis/datadoghq/v1alpha1/datadogagentdeployment_types.go
xornivore: Never mind, you are right I got confused with the diff here. :) One wish from the usage stand point - do you think it is possible to define this in such a way that we just have one field called `Config`? It just feels a bit odd to have both`Config` and then `CustomConfig`, thoughts?

vboulineau: IMHO I would not put `CustomConfig` inside `Config`, but we can find another name that may be more explicit. Maybe `DatadogYaml` or `CustomAgentYaml`?

clamoriniere: is it mandatory to have a pointer here? 

```suggestion
	CustomConfig string `json:"customConfig,omitempty"`
```

pkg/controller/datadogagentdeployment/agent.go
```diff
@@ -352,3 +350,34 @@ func newDaemonsetObjectMetaData(agentdeployment *datadoghqv1alpha1.DatadogAgentD
...
+	m := make(map[interface{}]interface{})
+	if err := yaml.Unmarshal([]byte(*dad.Spec.Agent.CustomConfig), m); err != nil {
+		return nil, err
```
clamoriniere: I think it can be good to wrap this `yaml.Unmarshal` err with a bit more context, like it is related to the CustomConfig file.

vboulineau: Yep, that makes sense, will do

pkg/apis/datadoghq/v1alpha1/datadogagentdeployment_types.go
vboulineau: Yeah we can simplify, it was to aligne with others, but we also already have an optional string without pointer, so I'll update that too.

vboulineau: Fixed

PR #18, opened 2020-01-02
Reconcile metric + events

ahmed-mez:
### What does this PR do?

- Send `reconcile.success` metric to Datadog, tagged by the reconcile error
- Send Creation/Deletion/Update resource events to Datadog

### Motivation

Better monitoring of the Datadog Agent components deployment


pkg/controller/utils/datadog/metrics_forwarder.go
```diff
@@ -85,6 +100,9 @@ func newMetricsForwarder(k8sClient client.Client, namespacedName types.Namespace
 		sendMetricsInterval: defaultSendMetricsInterval,
 		metricsPrefix:       defaultMetricsNamespace,
 		stopChan:            make(chan struct{}),
+		errorChan:           make(chan error, 100),
```
ahmed-mez: should we reduce this capacity?

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
```diff
@@ -229,7 +231,15 @@ type ReconcileDatadogAgentDeployment struct {
 // Note:
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+//
+// Reconcile wraps internelReconcile to send metrics based on reconcile errors
 func (r *ReconcileDatadogAgentDeployment) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	resp, err := r.internelReconcile(request)
+	r.forwarders.ProcessError(request.NamespacedName, err)
```
xornivore: Do you need to check for `err != nil` before calling `ProcessError`?

pkg/controller/utils/datadog/event.go
```diff
@@ -0,0 +1,45 @@
...
+type Event struct {
+	Title string
+	Type  EventType
+}
+
+// EventType enumerates the possible event types to be sent
+type EventType string
+
+// CreationEvent should be used for resource creation events
+const CreationEvent EventType = "Creation"
```
xornivore: You could wrap these four in a single `const` block for better readability.

pkg/controller/utils/datadog/metrics_forwarder.go
xornivore: what is the rationale for this value?

pkg/controller/utils/datadog/metrics_forwarder_test.go
```diff
@@ -624,3 +635,184 @@ func TestMetricsForwarder_setTags(t *testing.T) {
...
+					return errors.New("Wrong number of calls")
+				}
+				return nil
+			},
+		},
+		{
+			name: "last error init value, new auth error => send unsucess metric",
+			loadFunc: func() (*metricsForwarder, *fakeMetricsForwarder) {
+				f := &fakeMetricsForwarder{}
+				f.On("delegatedSendReconcileMetric", 0.0, []string{"cr_namespace:foo", "cr_name:bar", "reconcile_err:Unauthorized"})
```
xornivore: You could add `.Once()` here then below instead of `AssertCalled` and `AssertNumberOfCalls` you could do `AssertExpectations(t)`. It might make sense to do this for cases when you expect no calls as well.

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
ahmed-mez: we're interested in processing all errors whatever their values, a `nil` error indicates a reconcile success

pkg/controller/utils/datadog/metrics_forwarder_test.go
ahmed-mez: thanks 💯

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
xornivore: I thought that success maps to `ProcessEvent`?

pkg/controller/utils/datadog/metrics_forwarder.go
ahmed-mez: I couldn't find a rational quantification of this really, it depends on the errors and if the reconcile returns a `RequeueAfter` - I thought buffering 100 errors is enough in the worst case (when the reconcile loop keeps returning non nil errors continuously)

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
ahmed-mez: `ProcessEvent` only converts k8s events into datadog events -> sends events
`ProcessError` sends a metric

pkg/controller/datadogagentdeployment/agent.go
```diff
@@ -111,7 +112,7 @@ func (r *ReconcileDatadogAgentDeployment) deleteDaemonSet(logger logr.Logger, ag
 		return err
 	}
 	logger.Info("Delete DaemonSet", "daemonSet.Namespace", ds.Namespace, "daemonSet.Name", ds.Name)
-	r.recorder.Event(agentdeployment, corev1.EventTypeNormal, "Delete DaemonSet", fmt.Sprintf("%s/%s", ds.Namespace, ds.Name))
+	r.recordEvent(agentdeployment, corev1.EventTypeNormal, "Delete DaemonSet", fmt.Sprintf("%s/%s", ds.Namespace, ds.Name), datadog.DeletionEvent)
```
clamoriniere: maybe we can move the `fmt.Sprintf()` inside the function and just request an `ObjectMeta` parameter instead

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
```diff
@@ -75,6 +75,8 @@ func newReconciler(mgr manager.Manager) (reconcile.Reconciler, metricForwardersM
 type metricForwardersManager interface {
 	Register(types.NamespacedName)
 	Unregister(types.NamespacedName)
+	ProcessError(types.NamespacedName, error)
+	ProcessEvent(types.NamespacedName, datadog.Event)
```
clamoriniere: maybe you can change the functions signature with the interface that I proposed in previous comment. Like this you dont need anymore the function  `func getNamespacedName(dad *datadoghqv1alpha1.DatadogAgentDeployment) types.NamespacedName ` since from any Object (composition of ObjectMeta) they are all compatible with this interface

```
type metricForwardersManager interface {
	Register(NamespaceName)
	Unregister(NamespaceName)
	ProcessError(NamespaceName, error)
	ProcessEvent(NamespaceName, datadog.Event)
}
```

pkg/controller/datadogagentdeployment/event.go
```diff
@@ -0,0 +1,26 @@
...
+// the reason argument should contain the action and the object type e.g: Create DaemonSet
+// the message argument should contain object namespace and name e.g: default/datadog-agent
+func (r *ReconcileDatadogAgentDeployment) recordEvent(dad *datadoghqv1alpha1.DatadogAgentDeployment, eventtype, reason, message string, ddEventType datadog.EventType) {
+	r.recorder.Event(dad, eventtype, reason, message)
+	ddEvent := datadog.Event{
+		Title: fmt.Sprintf("%s %s", reason, message),
+		Type:  ddEventType,
+	}
+	r.forwarders.ProcessEvent(getNamespacedName(dad), ddEvent)
+}
```
clamoriniere: see other comment related to this function

pkg/controller/utils/datadog/forwarders_manager.go
```diff
@@ -62,6 +62,43 @@ func (f *ForwardersManager) Unregister(namespacedName types.NamespacedName) {
...
+	}
+	forwarder.errorChan <- err
+}
+
+// ProcessEvent dispatches recorded events to their corresponding metric forwarders
+func (f *ForwardersManager) ProcessEvent(namespacedName types.NamespacedName, event Event) {
+	f.Lock()
+	defer f.Unlock()
+	id := namespacedName.String()
+	forwarder, found := f.forwarders[id]
```
clamoriniere: I thinks this part can be mutualised in a sub function.

pkg/controller/utils/datadog/metrics_forwarder.go
```diff
@@ -112,22 +130,40 @@ func (mf *metricsForwarder) start(wg *sync.WaitGroup) {
 
 	mf.logger.Info("Datadog metrics forwarder initilized successfully")
 
+	// Send CR detection event
+	crEvent := crDetected(mf.id)
```
clamoriniere: it means each time the operator restart we have this event.
is it necessary?

ahmed-mez: I thought it would make sense to have a CR detection event since we have a CR deletion event

clamoriniere: but we have already a Create Event too?

clamoriniere: if we dont, maybe we can create on if we see: "dad.Status" is empty

ahmed-mez: we don't have a CR creation event, I named it `detection` event because at that point in the metric forwarder we only know that a CR exists (caused the creation of the metric forwarder itself)

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
ahmed-mez: > Like this you don't need anymore the function func getNamespacedName(dad *datadoghqv1alpha1.DatadogAgentDeployment) types.NamespacedName since from any Object (composition of ObjectMeta) they are all compatible with this interface

What if we're processing a `ClusterRole` creation event? we will pass an empty namespace to the forwarder manager and it won't match any metric forwarder and return an error, the event won't be sent

ahmed-mez: by getting the namespacedName from dad we make sure we match the corresponding metric forwarder ID

pkg/controller/datadogagentdeployment/agent.go
ahmed-mez: I agree that we need to reduce the number of arguments, but internal objects don't implement the `NameNamespace` interface (missing `GetKind()` method). We might need to use an intermediate struct to pass event data to `recordEvent`

ahmed-mez: also we always pass `corev1.EventTypeNormal` as `eventtype`, maybe we can remove the argument use the value directly in `recordEvent`

pkg/controller/datadogagentdeployment/datadogagentdeployment_controller.go
clamoriniere: my proposition was to change also the register and unregister function. Like this the forwardID generation is an implementation detail in the forwarder

pkg/controller/datadogagentdeployment/utils.go
```diff
@@ -1194,3 +1194,26 @@ func referSameObject(a, b metav1.OwnerReference) bool {
 
 	return aGV == bGV && a.Kind == b.Kind && a.Name == b.Name
 }
+
+// namespacedName implements the datadog.MonitoredObject interface
+// used to convert reconcile.Request into datadog.MonitoredObject
+type namespacedName struct {
+	namespace string
+	name      string
+}
```
clamoriniere: maybe we can do something like

```golang
type namespacedName struct {
	reconcile.Request
}

func (nsn namespacedName) GetNamespace() string {
	return nsn.Namespace
}

func (nsn namespacedName) GetName() string {
	return nsn.Name
}

// ...
```

pkg/controller/datadogagentdeployment/utils.go
```diff
@@ -1194,3 +1194,25 @@ func referSameObject(a, b metav1.OwnerReference) bool {
...
+	return nsn.Namespace
+}
+
+func (nsn namespacedName) GetName() string {
+	return nsn.Name
+}
+
+// getMonitoredObj returns a namespacedName from a reconcile.Request object
+func getMonitoredObj(req reconcile.Request) namespacedName {
+	var nsn namespacedName
```
clamoriniere:
```suggestion
	return namespacedName{req}
```

pkg/controller/utils/datadog/metrics_forwarder.go
```diff
@@ -64,28 +73,37 @@ type metricsForwarder struct {
 	globalTags          []string
 	tags                []string
 	stopChan            chan struct{}
+	errorChan           chan error
+	eventChan           chan Event
+	lastReconcileErr    error
 	namespacedName      types.NamespacedName
```
clamoriniere: can be replaced by `MonitoredObject`

ahmed-mez: it's needed by the k8s client to `Get` CR object

PR #23, opened 2020-01-09
Refactor and split Test_newExtendedDaemonSetFromInstance

xornivore:
### What does this PR do?

Split Test_newExtendedDaemonSetFromInstance to simplify dev experience running the test.

### Motivation

Should become a bit easier to contribute more tests and run these tests individually.

### DIFF
--- a/pkg/controller/datadogagentdeployment/agent_test.go
+++ b/pkg/controller/datadogagentdeployment/agent_test.go
@@ -24,16 +24,37 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
 
-func Test_newExtendedDaemonSetFromInstance(t *testing.T) {
-	logf.SetLogger(logf.ZapLogger(true))
-	authTokenValue := &corev1.EnvVarSource{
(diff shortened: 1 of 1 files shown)


pkg/controller/datadogagentdeployment/agent_test.go
```diff
@@ -24,16 +24,37 @@ import (
...
 )
 
-func Test_newExtendedDaemonSetFromInstance(t *testing.T) {
-	logf.SetLogger(logf.ZapLogger(true))
-	authTokenValue := &corev1.EnvVarSource{
-		SecretKeyRef: &corev1.SecretKeySelector{},
+var (
+	dadName = "foo"
+
+	authTokenValue = &corev1.EnvVarSource{
```
vboulineau: Much cleaner overall, just one maybe dumb question. Why some default are vars and others functions? For consistency sake, I'd put everything function. Minor detail though.

xornivore: Good point, thank you, let me do that.

PR #25, opened 2020-01-14
Support secret backend

ahmed-mez:
### What does this PR do?

Support decrypting secrets using a secret backend binary

### Motivation

Avoid declaring plaintext secrets in the CR manifest

### Additional Notes

### DIFF
--- a/pkg/controller/utils/datadog/forwarders_manager.go
+++ b/pkg/controller/utils/datadog/forwarders_manager.go
@@ -10,6 +10,8 @@ import (
 	"sync"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/DataDog/datadog-operator/pkg/secrets"
 )
 
 // ForwardersManager is a collection of metricsForwarder per DatadogAgent
@@ -17,6 +19,7 @@ import (
 type ForwardersManager struct {
 	forwarders map[string]*metricsForwarder
 	k8sClient  client.Client
+	decryptor  secrets.Decryptor
 	wg         sync.WaitGroup
 	sync.Mutex
 }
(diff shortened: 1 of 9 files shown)


pkg/controller/utils/datadog/forwarders_manager.go
```diff
@@ -33,6 +35,7 @@ func NewForwardersManager(k8sClient client.Client) *ForwardersManager {
 
 // Start must be handled by the controller-runtime manager
 func (f *ForwardersManager) Start(stop <-chan struct{}) error {
+	f.initSecretBackend()
```
clamoriniere: can we move it in the `NewForwardersManager()` function

pkg/secrets/secrets.go
```diff
@@ -0,0 +1,197 @@
...
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/spf13/pflag"
+)
+
+var (
+	secretBackendCommand = ""
+	secretBackend        *SecretBackend
```
clamoriniere: can we avoid to have global var?

pkg/secrets/secrets.go
```diff
@@ -0,0 +1,197 @@
...
+	ErrorMsg string `json:"error,omitempty"`
+}
+
+func init() {
+	pflag.StringVarP(&secretBackendCommand, "secretBackendCommand", "", "", "secret backend command")
+}
+
+// GetSecretBackend returns a shared SecretBackend instance
+// returns an error if the secret backend command is not set
+func GetSecretBackend() (*SecretBackend, error) {
```
clamoriniere: I would prefer to have a function: `NewSecretBackend()` that return new instance each time instead of using a global var

PR #36, opened 2020-02-14
Reconcile() updates now MetricsForwarder.Status in DatadogAgentStatus

clamoriniere:
### What does this PR do?

Avoid race condition update of the DatadogAgentStatus. 

### Motivation

before 2 differents goroutine may update the DatadogAgentStatus at the same time. that leads to concurrent update errors. 

### Additional Notes

### DIFF
--- a/pkg/controller/utils/datadog/forwarders_manager.go
+++ b/pkg/controller/utils/datadog/forwarders_manager.go
@@ -12,7 +12,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/DataDog/datadog-operator/pkg/secrets"
-
 	datadoghqv1alpha1 "github.com/DataDog/datadog-operator/pkg/apis/datadoghq/v1alpha1"
 )
 
@@ -100,7 +99,7 @@ func (f *ForwardersManager) ProcessEvent(obj MonitoredObject, event Event) {
 	forwarder.eventChan <- event
 }
 
-// MetricsForwarderStatusForObj used to retrieve the Metrics forwarder status for a give object
+// MetricsForwarderStatusForObj used to retrieve the Metrics forwarder status for a given object
 func (f *ForwardersManager) MetricsForwarderStatusForObj(obj MonitoredObject) *datadoghqv1alpha1.DatadogAgentCondition {
 	id := getObjID(obj)
 	forwarder, err := f.getForwarder(id)
--- a/pkg/controller/utils/datadog/forwarders_manager.go
+++ b/pkg/controller/utils/datadog/forwarders_manager.go
@@ -12,6 +12,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/DataDog/datadog-operator/pkg/secrets"
+
+	datadoghqv1alpha1 "github.com/DataDog/datadog-operator/pkg/apis/datadoghq/v1alpha1"
 )
 
 // ForwardersManager is a collection of metricsForwarder per DatadogAgent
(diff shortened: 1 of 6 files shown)


pkg/controller/utils/datadog/forwarders_manager.go
```diff
@@ -12,6 +12,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/DataDog/datadog-operator/pkg/secrets"
+
```
ahmed-mez: ✂️

pkg/controller/utils/datadog/forwarders_manager.go
```diff
@@ -98,6 +100,17 @@ func (f *ForwardersManager) ProcessEvent(obj MonitoredObject, event Event) {
 	forwarder.eventChan <- event
 }
 
+// MetricsForwarderStatusForObj used to retrieve the Metrics forwarder status for a give object
```
ahmed-mez:
```suggestion
// MetricsForwarderStatusForObj used to retrieve the Metrics forwarder status for a given object
```

pkg/controller/utils/datadog/metrics_forwarder.go
```diff
@@ -586,21 +576,23 @@ func (mf *metricsForwarder) getKeyFromSecret(dda *datadoghqv1alpha1.DatadogAgent
...
-		updatedDad.Status = *newStatus
-		return mf.k8sClient.Status().Update(context.TODO(), updatedDad)
+
+	mf.Lock()
+	defer mf.Unlock()
+	if mf.status == nil {
+		newCondition := condition.NewDatadogAgentStatusCondition(datadoghqv1alpha1.ConditionTypeActiveDatadogMetrics, conditionStatus, now, "", description)
+		mf.status = &newCondition
+	} else {
+		mf.status = condition.UpdateDatadogAgentStatusCondition(mf.status, now, datadoghqv1alpha1.ConditionTypeActiveDatadogMetrics, conditionStatus, description)
```
ahmed-mez: wdyt of using a thread safe `setStatus` method here instead?

pkg/controller/utils/datadog/metrics_forwarder.go
```diff
@@ -188,24 +194,17 @@ func (mf *metricsForwarder) connectToDatadogAPI() (bool, error) {
 	}
 	mf.logger.Info("Getting Datadog credentials")
 	apiKey, appKey, err := mf.getCredentials(dda)
+	defer mf.updateStatusIfNeeded(dda, err)
 	if err != nil {
 		mf.logger.Error(err, "cannot get Datadog credentials,  will retry later...")
-		if err = mf.updateStatusIfNeeded(dda, err); err != nil {
-			mf.logger.Error(err, "cannot update DatadogAgent status")
-		}
+		mf.updateStatusIfNeeded(dda, err)
```
ahmed-mez: This is not needed since we `defer mf.updateStatusIfNeeded(dda, err)`

ahmed-mez: ✂️

pkg/controller/utils/datadog/forwarders_manager.go
```diff
@@ -98,6 +100,17 @@ func (f *ForwardersManager) ProcessEvent(obj MonitoredObject, event Event) {
...
+func (f *ForwardersManager) MetricsForwarderStatusForObj(obj MonitoredObject) *datadoghqv1alpha1.DatadogAgentCondition {
+	id := getObjID(obj)
+	forwarder, err := f.getForwarder(id)
+	if err != nil {
+		// forwarder not present yet
+		return nil
```
ahmed-mez: shouldn't we at least log something here since we're ignoring the error ?

clamoriniere: I will add a log in the method caller

pkg/controller/utils/datadog/metrics_forwarder.go
clamoriniere: good idea 👍

pkg/controller/utils/datadog/metrics_forwarder.go
```diff
@@ -586,21 +581,24 @@ func (mf *metricsForwarder) getKeyFromSecret(dda *datadoghqv1alpha1.DatadogAgent
...
+		conditionStatus = corev1.ConditionFalse
+		description = "Datadog metrics forwarding error"
 	}
-	if !apiequality.Semantic.DeepEqual(&dda.Status, newStatus) {
-		updatedDad := dda.DeepCopy()
-		updatedDad.Status = *newStatus
-		return mf.k8sClient.Status().Update(context.TODO(), updatedDad)
+
+	mf.Lock()
+	defer mf.Unlock()
```
ahmed-mez: we no longer need to lock since getStatus and setStatus are thread safe

PR #42, opened 2020-02-20
[kubectl plugin] Add more commands

ahmed-mez:
### What does this PR do?

- Add get cluster agent leader command
- Add upgrade commands for the agent and the cluster agent
- Add commands to detect checks crashing

### Motivation

Better UX managing the datadog agent component in kuberentes

### Additional Notes


Makefile
```diff
@@ -1,6 +1,6 @@
 PROJECT_NAME=datadog-operator
 ARTIFACT=controller
-ARTIFACT_PLUGIN=kubectl-datadog
+ARTIFACT_PLUGIN=datadog
```
clamoriniere: I'm wondering why do we need this change?

pkg/plugin/cmd/agent/upgrade/upgrade.go
```diff
@@ -0,0 +1,165 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2020 Datadog, Inc.
+
+package upgrade
```
clamoriniere: Nice Idea to have an helper to only update the image version.

pkg/plugin/cmd/agent/upgrade/upgrade.go
```diff
@@ -0,0 +1,165 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	image          string
+	upgradeExample = `
+  # upgrade the version of the datadog agent to latest
+  %[1]s upgrade
```
clamoriniere: this means that the image will be: `datadog/agent:latest` ?
this command can work only once? 
WDYT about providing an option `--latest` to set to latest? 
so for the command will be: `kubectl datadog agent upgrade --latest`

pkg/plugin/cmd/agent/upgrade/upgrade.go
```diff
@@ -0,0 +1,165 @@
...
+
+  # upgrade the version of the datadog agent to latest defined in DatadogAgent foo
+  %[1]s upgrade foo
```
clamoriniere: sorry, but I dont get it. "foo" is the name of the DatadogAgent resource? 
what are the difference with the previous command? 

Last question what is "latest defined" version? is it the latest version supported by the operator? the plugin? how it works?

pkg/plugin/common/utils.go
```diff
@@ -60,3 +60,19 @@ func AskForInput(question string) (string, error) {
 	}
 	return strings.TrimSpace(text), nil
 }
+
+// HasImagePattern returns true if the image string respects the format <string>/<string>:<string>
+func HasImagePattern(image string) bool {
```
clamoriniere: can we use a regex here?

Makefile
ahmed-mez: that was just because in [all the existing krew manifests](https://krew.sigs.k8s.io/docs/developer-guide/plugin-manifest/) the binaries are named without the `kubectl-` prefix

ahmed-mez: but that's not a problem we can keep the current binary name, anyways the final name is generated from the `metadata.name` field based on https://krew.sigs.k8s.io/docs/developer-guide/plugin-manifest/#specifying-plugin-executable

cmd/kubectl-datadog/main.go
```diff
@@ -8,7 +8,7 @@ package main
 import (
 	"os"
 
-	"github.com/DataDog/datadog-operator/pkg/plugin/datadog"
+	"github.com/DataDog/datadog-operator/pkg/plugin/cmd/datadog"
```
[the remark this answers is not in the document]
ahmed-mez: yes I agree, I can move all the commands under `github.com/DataDog/datadog-operator/cmd/kubectl-datadog/cmd` and keep only the shared code `github.com/DataDog/datadog-operator/pkg/plugin/`

datadog.yaml
```diff
@@ -0,0 +1,24 @@
...
+  version: v0.0.1
+  shortDescription: Manage the Datadog Operator
+  description: |
+    The datadog kubectl plugin provides useful utilities to operate datadog-agent components
+    via the Datadog Operator and the DatadogAgent CRD.
+  homepage: https://github.com/DataDog/datadog-operator
+  platforms:
+  - uri: https://github.com/DataDog/datadog-operator/releases/download/v0.0.1/datadog-darwin-amd64.zip # TODO
+    sha256: # TODO
+    bin: datadog
```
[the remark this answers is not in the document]
ahmed-mez: it doesn't affect the final name but I'll change it anyway
> Krew will create a symbolic link named kubectl-foo (and kubectl-foo.exe on Windows) to your plugin executable after installation is complete. The name of the symbolic link comes from the plugin name (i.e. the metadata.name field).

pkg/plugin/cmd/agent/check/check.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+}
+
+// New provides a cobra command wrapping options
+func New(streams genericclioptions.IOStreams) *cobra.Command {
+	o := newOptions(streams)
+	cmd := &cobra.Command{
+		Use:          "check [flags]",
+		Short:        "Find check errors",
+		Example:      fmt.Sprintf(checkExample, "kubectl datadog agent"),
```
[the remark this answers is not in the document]
ahmed-mez: yes it would be super useful to have this especially the `--host` option (I prefer we name it `--node` though)

pkg/plugin/cmd/agent/check/check.go
```diff
@@ -0,0 +1,256 @@
...
+	restclient "k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/remotecommand"
+)
+
+const maxParallel = 10
+
+var (
+	checkExample = `
+  # check if the running Agents have detected check errors
+  %[1]s check
```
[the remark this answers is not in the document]
ahmed-mez: will add `node` and `pod-name` options 👍

pkg/plugin/cmd/agent/check/check.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	ns, err := cmd.Flags().GetString("namespace")
+	if err != nil {
+		return err
+	}
+
+	if ns != "" {
+		o.userNamespace = ns
+	}
```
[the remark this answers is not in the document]
ahmed-mez: makes sense, thanks for pointing out

pkg/plugin/cmd/agent/upgrade/upgrade.go
ahmed-mez: sounds good

pkg/plugin/cmd/agent/upgrade/upgrade.go
ahmed-mez: > what are the difference with the previous command?
when multiple DatadogAgent resources are applied, the first command will upgrade all of them, but if the DatadogAgent is mentioned the command upgrades only the image of the chosen DatadogAgent

> Last question what is "latest defined" version?
the wording is a bit confusing i'll update it, but by latest I mean the image tag `latest`, and the `version of the datadog agent` is defined in `DatadogAgent foo` 

although I guess we need to print a warning if we find the current deployed tag is already `latest` - and make it clear we're not going to trigger an upgrade

pkg/plugin/cmd/agent/check/check.go
```diff
@@ -0,0 +1,256 @@
...
+func (o *options) run() error {
+	agentPods, err := o.getPods("agent")
+	if err != nil {
+		return fmt.Errorf("unable to get Agent pods: %v", err)
+	}
+	clcPods, err := o.getPods("cluster-checks-runner")
+	if err != nil {
+		return fmt.Errorf("unable to get Agent pods: %v", err)
+	}
+	fmt.Println(fmt.Sprintf("Found %d node Agents and %d Cluster-Check Runners", len(agentPods), len(clcPods)))
```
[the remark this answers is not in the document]
ahmed-mez: actually it provides some functions that fallback to Stderr if not set if `outWriter` is not set - I'm not seeing the reason why we should call these functions since we want to print to Stdout, am I missing something here?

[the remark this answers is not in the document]
clamoriniere: One major interest is to be able to unit-test the command output.

cmd/kubectl-datadog/agent/upgrade/upgrade.go
```diff
@@ -0,0 +1,174 @@
...
+
+// options provides information required by agent upgrade command
+type options struct {
+	genericclioptions.IOStreams
+	configFlags          *genericclioptions.ConfigFlags
+	args                 []string
+	client               client.Client
+	userNamespace        string
+	userDatadogAgentName string
+}
```
clamoriniere: maybe you can create another struct `common.Options` 

```golang
package common

type Options struct {
        ConfigFlags          *genericclioptions.ConfigFlags
	Client               client.Client
	UserNamespace        string
	UserDatadogAgentName string
}

// SetNamespace configures the namespace
func (o *Options) SetNamespace(ns string) {
	o.userNamespace = ns
}

// SetClient configures the client if the field exists
func (o *Options) SetClient(client client.Client) {
	o.Client = client
}

// SetClientset configures the clientset if the field exists
func (o *Options) SetClientset(clientset *kubernetes.Clientset) {
}

// GetClientConfig returns the client config
func (o *Options) GetClientConfig() clientcmd.ClientConfig {
	return o.ConfigFlags.ToRawKubeConfigLoader()
}

/// another file
package upgrade

type options struct {
	genericclioptions.IOStreams
        common.Options
	args                 []string
}
```

cmd/kubectl-datadog/agent/check/check.go
```diff
@@ -0,0 +1,309 @@
...
+	o.restConfig, err = o.configFlags.ToRawKubeConfigLoader().ClientConfig()
+	if err != nil {
+		return fmt.Errorf("unable to instantiate restConfig: %v", err)
+	}
+	cmd.Println()
+	return common.SetupOptions(o, cmd)
+}
+
+// validate ensures that all required arguments and flag values are provided
+func (o *options) validate() error {
```
clamoriniere: I think it can be good to have:
```suggestion
func (o *options) validate(cmd *cobra.Command) error {
```
like this we can access all the useful methods in `Command`. I know it is not the case for now, but maybe in the futur.

cmd/kubectl-datadog/agent/check/check.go
```diff
@@ -0,0 +1,309 @@
...
+	if podName != "" && node != "" {
+		fmt.Println("pod-name and node flags are both set, ignoring the node flag")
+	}
+	return nil
+}
+
+// run runs the check command
+func (o *options) run() error {
```
clamoriniere: I think it can be good to have:
```suggestion
func (o *options) run(cmd *cobra.Command) error {
```
like this we can access all the useful methods in `Command`. I know it is not the case for now, but maybe in the futur.

cmd/kubectl-datadog/agent/upgrade/upgrade.go
ahmed-mez: I actually thought about this but `SetClient` and `SetClientset` are not the same for all the commands, some commands have only one client type, and some commands have both

cmd/kubectl-datadog/agent/check/check.go
ahmed-mez: will do that to access the writer methods

PR #48, opened 2020-03-06
Explicitly set DOCKER_HOST

L3n41c:
### What does this PR do?

Explicitly set DOCKER_HOST

### Motivation

Following f3e7d709, the docker socket is at a non-standard location inside the agent container.

### Additional Notes

This change implements the same behavior as the one introduced in helm/charts#21300

### DIFF
--- a/pkg/controller/datadogagent/utils.go
+++ b/pkg/controller/datadogagent/utils.go
@@ -11,6 +11,7 @@ import (
 	"math/rand"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"time"
 
 	datadoghqv1alpha1 "github.com/DataDog/datadog-operator/pkg/apis/datadoghq/v1alpha1"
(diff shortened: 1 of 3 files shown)


pkg/controller/datadogagent/utils.go
```diff
@@ -542,6 +543,13 @@ func getEnvVarsForAgent(dda *datadoghqv1alpha1.DatadogAgent) ([]corev1.EnvVar, e
 			Name:  datadoghqv1alpha1.DDCriSocketPath,
 			Value: filepath.Join(datadoghqv1alpha1.HostCriSocketPathPrefix, *dda.Spec.Agent.Config.CriSocket.CriSocketPath),
 		})
+
+		if strings.HasSuffix(*dda.Spec.Agent.Config.CriSocket.CriSocketPath, "docker.sock") {
+			envVars = append(envVars, corev1.EnvVar{
+				Name:  datadoghqv1alpha1.DockerHost,
+				Value: "unix://" + filepath.Join(datadoghqv1alpha1.HostCriSocketPathPrefix, *dda.Spec.Agent.Config.CriSocket.CriSocketPath),
```
clamoriniere: can we use `path.Join()` to incorporate the "unix://"

PR #49, opened 2020-03-10
Make the forwarder operators reply to /healthz

L3n41c:
### What does this PR do?

Make the forwarder operators reply to `/healthz` and make this endpoint used as liveness probe.

### Motivation

The way operators are working is that they elect a leader very early in their startup sequence and their startup is suspended until a they effectively become leader.

This also applies to the `/metrics` openmetrics endpoint.
This makes this endpoint unsuitable for liveness probe checks since it will make all the non-leader operators being uselessly restarted.

This change introduces a dedicated endpoint aiming at being used as liveness probe.
Because it must successfully respond even before an operator can become leader, it is currently checking nothing.


vendor/github.com/heptiolabs/healthcheck/README.md
```diff
@@ -0,0 +1,92 @@
+# healthcheck
+[![Build Status](https://travis-ci.org/heptiolabs/healthcheck.svg?branch=master)](https://travis-ci.org/heptiolabs/healthcheck)
+[![Go Report Card](https://goreportcard.com/badge/github.com/heptiolabs/healthcheck)](https://goreportcard.com/report/github.com/heptiolabs/healthcheck)
+[![GoDoc](https://godoc.org/github.com/heptiolabs/healthcheck?status.svg)](https://godoc.org/github.com/heptiolabs/healthcheck)
+
+Healthcheck is a library for implementing Kubernetes [liveness and readiness](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/) probe handlers in your Go application.
+
+## Features
+
+ - Integrates easily with Kubernetes. This library explicitly separates liveness vs. readiness checks instead of lumping everything into a single category of check.
```
apigirl:
```suggestion
 - This library integrates easily with Kubernetes because it explicitly separates liveness vs. readiness checks instead of lumping everything into a single category of check.
```

vendor/github.com/heptiolabs/healthcheck/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+ - If one of your liveness checks fails or your app becomes totally unresponsive, Kubernetes will restart your container.
+
+ ## HTTP Endpoints
+ When you run `go http.ListenAndServe("0.0.0.0:8086", health)`, two HTTP endpoints are exposed:
+
+  - **`/live`**: liveness endpoint (HTTP 200 if healthy, HTTP 503 if unhealthy)
+  - **`/ready`**: readiness endpoint (HTTP 200 if healthy, HTTP 503 if unhealthy)
+
+Pass the `?full=1` query parameter to see the full check results as JSON. These are omitted by default for performance.
```
apigirl:
```suggestion
Pass the `?full=1` query parameter to see the full check results as JSON. This parameter is excluded by default for performance.
```

LICENSE-3rdparty.csv
```diff
@@ -65,6 +65,7 @@ core,"github.com/gregjones/httpcache/diskcache",MIT
 core,"github.com/hako/durafmt",MIT
 core,"github.com/hashicorp/golang-lru",MPL-2.0
 core,"github.com/hashicorp/golang-lru/simplelru",MPL-2.0
+core,"github.com/heptiolabs/healthcheck",Apache-2.0
 core,"github.com/imdario/mergo",NewBSD
```
jeremy-lq: Please use SPDX identifiers for the license field.

jeremy-lq: 2 additional notes:
* The LICENSE file should contain the Apache 2 license template.
* We have a default NOTICE file that must be used.

clamoriniere: Hi @jeremy-lq
I created a dedicated PR for fixing the Licence issue: https://github.com/DataDog/datadog-operator/pull/53

vendor/github.com/heptiolabs/healthcheck/README.md
L3n41c: Hello @apigirl ,

This is “vendored” code.
I.e.: this is a 3rd party dependency that the `datadog-operator` depends on and that has been copied verbatim inside this repo in order to not depend on the external world.
Such vendored 3rd party component must remain unmodified compared to upstream.
In order to fix the wording of this documentation, a PR should be submitted upstream instead: https://github.com/heptiolabs/healthcheck

vendor/github.com/heptiolabs/healthcheck/README.md
L3n41c: Vendored code as well.

PR #53, opened 2020-03-24
Fix License file

clamoriniere:
### What does this PR do?

Fix LICENSE file and introduce NOTICE file

### Motivation

Solve the comment: https://github.com/DataDog/datadog-operator/pull/49#discussion_r396636891


NOTICE
```diff
@@ -0,0 +1,4 @@
+Datadog datadog-agent
```
jeremy-lq: s/datadog-agent/datadog-operator/

PR #59, opened 2020-03-29
change probes port, make it configurable

clamoriniere:
### What does this PR do?

Change the probes port from `8080` to `9090`, also make it parametrisable.

### Motivation

the Operatorhub.io validation process inject a container that use also the port `8080`. in order to make the validation test working.

### Additional Notes


chart/datadog-operator/templates/deployment.yaml
```diff
@@ -55,12 +55,12 @@ spec:
           readinessProbe:
             httpGet:
               path: /ready
-              port: 8080
+              port: 9090
```
ahmed-mez: shouldn't we make this configurable in values.yaml?

PR #64, opened 2020-04-03
Fix CLC config

ahmed-mez:
### What does this PR do?

- Get rid of `clusterChecksRunnerEnabled`
- Enable endpoint checks - be aligned with the helm chart

### Motivation

### Additional Notes

### DIFF
--- a/pkg/controller/datadogagent/datadogagent_controller_test.go
+++ b/pkg/controller/datadogagent/datadogagent_controller_test.go
@@ -1489,6 +1489,7 @@ func TestReconcileDatadogAgent_Reconcile(t *testing.T) {
 						Labels:                     map[string]string{"label-foo-key": "label-bar-value"},
 						Status:                     &datadoghqv1alpha1.DatadogAgentStatus{},
 						ClusterAgentEnabled:        true,
+						ClusterChecksEnabled:       true,
 						ClusterChecksRunnerEnabled: true,
 					}
 					dda := test.NewDefaultedDatadogAgent(resourcesNamespace, resourcesName, dadOptions)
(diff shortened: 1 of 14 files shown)


pkg/controller/datadogagent/utils.go
```diff
@@ -528,30 +544,22 @@ func getEnvVarsForAgent(dda *datadoghqv1alpha1.DatadogAgent) ([]corev1.EnvVar, e
 				ValueFrom: getClusterAgentAuthToken(dda),
 			},
 		}
-		if *spec.ClusterAgent.Config.ClusterChecksRunnerEnabled && spec.ClusterChecksRunner == nil {
-			clusterEnv = append(clusterEnv, corev1.EnvVar{
-				Name:  datadoghqv1alpha1.DDExtraConfigProviders,
-				Value: datadoghqv1alpha1.ClusterChecksConfigProvider,
-			})
+		if *spec.ClusterAgent.Config.ClusterChecksEnabled {
+			if spec.ClusterChecksRunner == nil {
```
vboulineau, line 548: Why do we have a distinction on provider when no CLC?

ahmed-mez: when clc runners disabled the node agent needs to be configured to pull cluster checks AND endpoint checks, when the clc runners enabled they take care of running cluster checks and the node agents must only endpoint checks

pkg/controller/datadogagent/datadogagent_controller_test.go
```diff
@@ -1662,6 +1665,7 @@ func TestReconcileDatadogAgent_Reconcile(t *testing.T) {
 						Labels:                     map[string]string{"label-foo-key": "label-bar-value"},
 						Status:                     &datadoghqv1alpha1.DatadogAgentStatus{},
 						ClusterAgentEnabled:        true,
+						ClusterChecksEnabled:       true,
 						ClusterChecksRunnerEnabled: true,
```
clamoriniere, line 1669: I think we need te remove this field.

PR #67, opened 2020-04-06
fix zip file version in getting_started.md

clamoriniere:
### What does this PR do?

Fix the .zip version to fit with the last datadog-operator docker image release.

### Motivation

sync the chart version and the docker image in the `getting_started.md` documentation

### Additional Notes


l0k0ms:
Let's not forget to update this at each release then :)

PR #71, opened 2020-04-07
Fix panic in case of missing `ClusterChecksEnabled` property in the CR

L3n41c:
### What does this PR do?

Fix panic in case of missing `ClusterChecksEnabled` property in the CR.

### Motivation

Panics are uncool.

### DIFF
--- a/pkg/controller/datadogagent/clusteragent.go
+++ b/pkg/controller/datadogagent/clusteragent.go
@@ -424,7 +424,7 @@ func getEnvVarsForClusterAgent(dda *datadoghqv1alpha1.DatadogAgent) []corev1.Env
 		},
 		{
 			Name:  datadoghqv1alpha1.DDClusterChecksEnabled,
-			Value: strconv.FormatBool(*spec.ClusterAgent.Config.ClusterChecksEnabled),
+			Value: datadoghqv1alpha1.BoolToString(spec.ClusterAgent.Config.ClusterChecksEnabled),
 		},
 		{
 			Name:  datadoghqv1alpha1.DDClusterAgentKubeServiceName,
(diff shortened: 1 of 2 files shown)


pkg/controller/datadogagent/clusteragent.go
```diff
@@ -484,7 +484,7 @@ func getEnvVarsForClusterAgent(dda *datadoghqv1alpha1.DatadogAgent) []corev1.Env
 	}
 
 	// Cluster Checks config
-	if *spec.ClusterAgent.Config.ClusterChecksEnabled {
+	if spec.ClusterAgent.Config.ClusterChecksEnabled != nil && *spec.ClusterAgent.Config.ClusterChecksEnabled {
```
ahmed-mez, line 487: can we use `datadoghqv1alpha1.BoolValue()` here

pkg/controller/datadogagent/utils.go
```diff
@@ -544,7 +544,7 @@ func getEnvVarsForAgent(dda *datadoghqv1alpha1.DatadogAgent) ([]corev1.EnvVar, e
 				ValueFrom: getClusterAgentAuthToken(dda),
 			},
 		}
-		if *spec.ClusterAgent.Config.ClusterChecksEnabled {
+		if spec.ClusterAgent.Config.ClusterChecksEnabled != nil && *spec.ClusterAgent.Config.ClusterChecksEnabled {
```
ahmed-mez, line 547: can we use `datadoghqv1alpha1.BoolValue()` here ?

pkg/controller/datadogagent/clusteragent.go
L3n41c: Yes we can !

pkg/controller/datadogagent/utils.go
L3n41c: Yes we can !

PR #73, opened 2020-04-09
Add SecCompCustomProfileConfigMap for system-probe

L3n41c:
### What does this PR do?

Add `SecCompCustomProfileConfigMap` for `system-probe`.

### Motivation

Implement the following behaviours regarding the `seccomp` profile for `system-probe`:

(“ConfigMap” is a shortcut for “the ConfigMap containing the seccomp profile for `system-probe`.
“init container” is a shortcut for “the init container that copies the seccomp profile from the ConfigMap to the host.)

* Let the operator use its own `seccomp` profile:
  * create a ConfigMap
  * use an init container to copy its content to the hosts.
  * ```
    systemProbe:
      enabled: true
* Make `system-probe` use the default `seccomp` profile:
  * do **not** create the ConfigMap
  * do **not** create the init container
  * ```
    systemProbe:
      enabled: true
      secCompProfileName: runtime/default
* Make `system-probe` use a `seccomp` profile that is already available on the hosts (in `/var/lib/kubelet/seccomp/my_custom_profile`)
  * do **not** create the ConfigMap
  * do **not** create the init container
  * ```
    systemProbe:
      enabled: true
      secCompProfileName: localhost/my_custom_profile
* Make `system-probe` use a `seccomp` profile provided in a user-supplied ConfigMap.
  * do **not** create the ConfigMap. The user has to create one beforehand.
  * use an init container to copy the content of the user provided ConfigMap to the hosts.
  * ```
    systemProbe:
      enabled: true
      secCompCustomProfileConfigMap: lenaic-system-probe-seccomp

### Additional Notes

### DIFF
--- a/pkg/apis/datadoghq/v1alpha1/datadogagent_types.go
+++ b/pkg/apis/datadoghq/v1alpha1/datadogagent_types.go
@@ -313,7 +313,7 @@ type SystemProbeSpec struct {
 
 	// SecCompCustomProfileConfigMap specify a pre-existing ConfigMap containing a custom SecComp profile
 	// +optional
-	SecCompCustomProfileConfigMap string `json:"secCompCustomProfileConfigMap"`
+	SecCompCustomProfileConfigMap string `json:"secCompCustomProfileConfigMap,omitempty"`
 
 	// SecCompProfileName specify a seccomp profile
 	// +optional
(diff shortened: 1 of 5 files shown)


pkg/apis/datadoghq/v1alpha1/datadogagent_types.go
```diff
@@ -311,6 +311,10 @@ type SystemProbeSpec struct {
 	// +optional
 	SecCompRootPath string `json:"secCompRootPath,omitempty"`
 
+	// SecCompCustomProfileConfigMap specify a pre-existing ConfigMap containing a custom SecComp profile
+	// +optional
+	SecCompCustomProfileConfigMap string `json:"secCompCustomProfileConfigMap"`
```
clamoriniere, line 316:
```suggestion
	SecCompCustomProfileConfigMap string `json:"secCompCustomProfileConfigMap,omitempty"`
```

PR #88, opened 2020-05-11
Add container logs scraping from file by default

vboulineau:
### What does this PR do?

Align with the Helm chart. We are mounting pod logs host path and container logs host path.

### Motivation

### Additional Notes


vboulineau:
> LGTM but let's see why e2e tests fail before merging

Seems to be random failure, I cannot reproduce it and different runs do not yield the same failure.
I'm guessing it's due to:

```
2020-05-11 16:20:23 UTC | SYS-PROBE | CRITICAL | (cmd/system-probe/main.go:122 in main) | failed to create system probe: could not enable kprobe(kprobe/tcp_get_info) used for offset guessing: cannot write "p:ptcp_get_info tcp_get_info\n" to kprobe_events: write /sys/kernel/debug/tracing/kprobe_events: device or resource busy
```

But I'm not sure. I believe we could improve readability of e2e tests logs, I'll make another PR later for that.

ahmed-mez:
Thanks for looking into it, :shipit:

PR #90, opened 2020-05-19
Honor DD_SITE by not adding DD_DD_URL env var by default

thehunt33r:
### What does this PR do?

While trying Datadog Operator for my use case, I started noticing that no matter if I had set `Site` to `EU`, my agents would try to send data to `https://app.datadoghq.com`. 

After inspecting the generated manifests for both the datadog agent and the datadog cluster agent, I noticed that even though `DD_SITE` was set to `EU`, a `DD_DD_URL` env var was populated and set to `https://app.datadoghq.com`, overriding `DD_SITE`.

In the documentation, it is said that `DD_DD_URL` is optional, and should only be set when one wants to override the endpoint that `DD_SITE` would make Datadog Agent use (for example if you use a proxy, or need a mock endpoint, etc...)

After looking through the codebase, I noticed that `DD_DD_URL` was indeed set by default. This PR is addressing this issue

### Motivation

Solving my issues with Datadog Operator

### Additional Notes

For Datadog people, a support case was opened (#342506)

### DIFF
--- a/cmd/kubectl-datadog/flare/flare.go
+++ b/cmd/kubectl-datadog/flare/flare.go
@@ -132,7 +132,7 @@ func (o *options) complete(cmd *cobra.Command, args []string) error {
 	}
 
 	if ddSite == "" {
-		ddSite, err = common.AskForInput("Please enter your Datatod site [us/eu] (default us): ")
+		ddSite, err = common.AskForInput("Please enter your Datatog site [us/eu] (default us): ")
 		if err != nil {
 			return err
 		}
--- a/pkg/apis/datadoghq/v1alpha1/datadogagent_default.go
+++ b/pkg/apis/datadoghq/v1alpha1/datadogagent_default.go
@@ -19,7 +19,6 @@ import (
 // default values
 const (
 	DefaultLogLevel                               string = "INFO"
-	defaultDatadogURL                             string = "https://app.datadoghq.com"
 	defaultAgentImage                             string = "datadog/agent:latest"
 	defaultCollectEvents                          bool   = false
 	defaultLeaderElection                         bool   = false
@@ -152,10 +151,6 @@ func IsDefaultedImageConfig(imageConfig *ImageConfig) bool {
 // returns true if yes, else false
 func IsDefaultedDatadogAgentSpecAgentConfig(config *NodeAgentConfig) bool {
 	if config == nil {
-		return false
-	}
-
-	if config.DDUrl == nil {
 		return false
 	}
 
@@ -400,10 +395,6 @@ func DefaultDatadogAgentSpecAgentImage(image *ImageConfig) *ImageConfig {
 func DefaultDatadogAgentSpecAgentConfig(config *NodeAgentConfig) *NodeAgentConfig {
 	if config == nil {
 		config = &NodeAgentConfig{}
-	}
-
-	if config.DDUrl == nil {
-		config.DDUrl = NewStringPointer(defaultDatadogURL)
 	}
 
 	if config.LogLevel == nil {
--- a/pkg/controller/datadogagent/agent_rbac.go
+++ b/pkg/controller/datadogagent/agent_rbac.go
@@ -52,7 +52,7 @@ func (r *ReconcileDatadogAgent) manageAgentRBACs(logger logr.Logger, dda *datado
 		return reconcile.Result{}, err
 	}
 
-	// Create ClusterRoleBindig
+	// Create ClusterRoleBinding
 	clusterRoleBinding := &rbacv1.ClusterRoleBinding{}
 	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: rbacResourcesName}, clusterRoleBinding); err != nil {
 		if errors.IsNotFound(err) {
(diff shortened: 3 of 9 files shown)


cmd/kubectl-datadog/flare/flare.go
```diff
@@ -132,7 +132,7 @@ func (o *options) complete(cmd *cobra.Command, args []string) error {
 	}
 
 	if ddSite == "" {
-		ddSite, err = common.AskForInput("Please enter your Datatod site [us/eu] (default us): ")
+		ddSite, err = common.AskForInput("Please enter your Datatog site [us/eu] (default us): ")
```
thehunt33r, line 135: Fixing a typo in the flare sending process. Happy to make this change in a dedicated PR.

pkg/apis/datadoghq/v1alpha1/datadogagent_default.go
```diff
@@ -19,7 +19,6 @@ import (
 // default values
 const (
 	DefaultLogLevel                               string = "INFO"
-	defaultDatadogURL                             string = "https://app.datadoghq.com"
```
thehunt33r, line 22: Removing the default URL for `DD_DD_URL`

pkg/apis/datadoghq/v1alpha1/datadogagent_default.go
```diff
@@ -402,10 +397,6 @@ func DefaultDatadogAgentSpecAgentConfig(config *NodeAgentConfig) *NodeAgentConfi
 		config = &NodeAgentConfig{}
 	}
 
-	if config.DDUrl == nil {
-		config.DDUrl = NewStringPointer(defaultDatadogURL)
```
thehunt33r, line 406: Do not try to assign a value (in this case, default value) for `config.DDUrl` if it is not set

pkg/controller/datadogagent/agent_rbac.go
```diff
@@ -52,7 +52,7 @@ func (r *ReconcileDatadogAgent) manageAgentRBACs(logger logr.Logger, dda *datado
 		return reconcile.Result{}, err
 	}
 
-	// Create ClusterRoleBindig
+	// Create ClusterRoleBinding
```
thehunt33r, line 55: Another typo. Again, happy to make this change in a dedicated PR.

pkg/controller/datadogagent/clusteragent.go
```diff
@@ -436,7 +436,7 @@ func getEnvVarsForClusterAgent(dda *datadoghqv1alpha1.DatadogAgent) []corev1.Env
 		},
 	}
 
-	if spec.Agent != nil {
+	if spec.Agent != nil && spec.Agent.Config.DDUrl != nil {
```
thehunt33r, line 439: Only set `DDUrl` if it has been set.

pkg/controller/datadogagent/clusterchecksrunner.go
```diff
@@ -374,6 +370,13 @@ func getEnvVarsForClusterChecksRunner(dda *datadoghqv1alpha1.DatadogAgent) []cor
 		})
 	}
 
+	if spec.Agent.Config.DDUrl != nil {
```
thehunt33r, line 373: This conditional checks whether `DDUrl` has been set. As it was set by default previously, the methods would assume it was set. Now we need to check it to make sure we don't cause a panic

pkg/controller/datadogagent/clusterchecksrunner.go
```diff
@@ -374,6 +370,13 @@ func getEnvVarsForClusterChecksRunner(dda *datadoghqv1alpha1.DatadogAgent) []cor
...
+		envVars = append(envVars, corev1.EnvVar{
+			Name:  datadoghqv1alpha1.DDddURL,
+			Value: *spec.Agent.Config.DDUrl,
```
ahmed-mez, line 376: nit
```suggestion
			Value: datadoghqv1alpha1.BoolValue(spec.Agent.Config.DDUrl),
```
using `datadoghqv1alpha1.BoolValue` would be safer here, we can also use it to get `spec.ClusterChecksRunner.Config.LogLevel` in the block above

thehunt33r:
@vboulineau Is there anything I need to do in order for the `dd-gitlab` checks to run?

thehunt33r:
So I've successfully deployed those changes to my cluster. With the latest commit, all the agents start correctly.

However, despite the agents functioning as expected, the agent is still seen as `Progressing`
```bash
$ kubectl get dd datadog
NAME      ACTIVE   AGENT                 CLUSTER-AGENT     CLUSTER-CHECKS-RUNNER   AGE
datadog   True     Progressing (0/0/0)   Running (2/2/2)                           22m
$ kubectl get deployment datadog-cluster-agent
NAME                    READY   UP-TO-DATE   AVAILABLE   AGE
datadog-cluster-agent   2/2     2            2           22m
$ kubectl get daemonset datadog-agent
NAME            DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR   AGE
datadog-agent   1         1         1       1            1           <none>          22m
```

Is this an expected behavior or am I maybe missing something else?

vboulineau:
@thehunt33r I think we can merge this PR fixing DD_SITE. You're right about the client inside the operator not using `site`, we'll fix it.

About the `Progressing` state that's not expected, if you're willing to spend more time on time, feel free, but of course we'll look into it on our side and fix it otherwise!

PR #91, opened 2020-05-20
Add an all-in-one chart that deploys both the operator and an agent

L3n41c:
### What does this PR do?

Add an all-in-one chart that deploys both the operator and an agent

### Motivation

Simplify on-boarding

### Additional Notes


docs/getting_started.md
```diff
@@ -24,140 +28,36 @@ To use the Datadog Operator, deploy it in your Kubernetes cluster. Then create a
 4. Create the namespace:
 
    ```shell
-   kubectl create ns $DD_NAMESPACE
+   kubectl create namespace $DD_NAMESPACE
```
cohenyair, line 31: @L3n41c could we remove step 3 and 4 and use the default namespace and a default deployment name? it would simplify the basic deployment even further

L3n41c: Hi @cohenyair !
Thanks you for the feedback.
I’ve simplified the instructions by letting the agent being installed in the current namespace.
Can you PTAL ?

cohenyair: That's great! LGTM

PR #92, opened 2020-05-22
Use DdSite or DdUrl for MetricsForwarder API client

thehunt33r:
### What does this PR do?
 
This PR:
*  Adds a field to the `metricsForwarder` struct: `baseURL`
* Adds a `getBaseURL` method that returns the appropriate base URL based on the `ddUrl` or `site` fields in the definition of the agent
* Sets the base URL after the client has been initialized
* Adds the base URL used when error-ing out on invalid credentials.

### Motivation

After submitting #90 and testing it out, I realized that the operator was also trying to push metrics, grabbing the credentials from the agent. Unfortunately the metrics forwarder wasn't using the DD_SITE / DD_DD_URL setting and would use the default of the client SDK, which is `https://api.datadoghq.com`

### Additional Notes

I've tested this successfully in a 1.16 EKS cluster, using `site: datadoghq.eu`

PS: The validation build failure doesn't seem to be related to my changes.

### DIFF
--- a/pkg/apis/datadoghq/v1alpha1/test/new.go
+++ b/pkg/apis/datadoghq/v1alpha1/test/new.go
@@ -57,6 +57,7 @@ type NewDatadogAgentOptions struct {
 	ClusterChecksRunnerVolumeMounts []corev1.VolumeMount
 	ClusterChecksRunnerEnvVars      []corev1.EnvVar
 	APIKeyExistingSecret            string
+	Site                            string
 }
 
 // NewDefaultedDatadogAgent returns an initialized and defaulted DatadogAgent for testing purpose
@@ -104,6 +105,7 @@ func NewDefaultedDatadogAgent(ns, name string, options *NewDatadogAgentOptions) 
 		}
 
 		ad.Spec.Agent.DaemonsetName = options.AgentDaemonsetName
+		ad.Spec.Site = options.Site
 
 		if options.Status != nil {
 			ad.Status = *options.Status
@@ -123,6 +125,7 @@ func NewDefaultedDatadogAgent(ns, name string, options *NewDatadogAgentOptions) 
 				},
 				DeploymentName: options.ClusterAgentDeploymentName,
 			}
+
 			if options.MetricsServerEnabled {
 				ad.Spec.ClusterAgent.Config.MetricsProviderEnabled = datadoghqv1alpha1.NewBoolPointer(true)
 
--- a/pkg/controller/utils/datadog/metrics_forwarder.go
+++ b/pkg/controller/utils/datadog/metrics_forwarder.go
@@ -49,6 +49,7 @@ const (
 	reconcileMetricFormat       = "%s.reconcile.success"
 	reconcileErrTagFormat       = "reconcile_err:%s"
 	datadogOperatorSourceType   = "datadog_operator"
+	defaultbaseURL              = "https://api.datadoghq.com"
 )
 
 var (
@@ -81,6 +82,7 @@ type metricsForwarder struct {
 	delegator           delegatedAPI
 	decryptor           secrets.Decryptor
 	creds               sync.Map
+	baseURL             string
 	sync.Mutex
 	status *datadoghqv1alpha1.DatadogAgentCondition
 }
@@ -108,6 +110,7 @@ func newMetricsForwarder(k8sClient client.Client, decryptor secrets.Decryptor, o
 		lastReconcileErr:    errInitValue,
 		decryptor:           decryptor,
 		creds:               sync.Map{},
+		baseURL:             defaultbaseURL,
 		logger:              log.WithValues("CustomResource.Namespace", obj.GetNamespace(), "CustomResource.Name", obj.GetName()),
 	}
 }
@@ -133,7 +136,7 @@ func (mf *metricsForwarder) start(wg *sync.WaitGroup) {
 		return
 	}
 
-	mf.logger.Info("Datadog metrics forwarder initilized successfully")
+	mf.logger.Info("Datadog metrics forwarder initialized successfully")
 
 	// Send CR detection event
 	crEvent := crDetected(mf.id)
@@ -200,6 +203,8 @@ func (mf *metricsForwarder) connectToDatadogAPI() (bool, error) {
 	}
 	mf.logger.Info("Getting Datadog credentials")
 	apiKey, appKey, err := mf.getCredentials(dda)
+	mf.baseURL = getbaseURL(dda)
+	mf.logger.Info("Got Datadog Site", "site", mf.baseURL)
 	defer mf.updateStatusIfNeeded(err)
 	if err != nil {
 		mf.logger.Error(err, "cannot get Datadog credentials,  will retry later...")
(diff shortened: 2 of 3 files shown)


pkg/controller/utils/datadog/metrics_forwarder.go
```diff
@@ -653,3 +659,12 @@ func (mf *metricsForwarder) isErrChanFull() bool {
 func (mf *metricsForwarder) isEventChanFull() bool {
 	return len(mf.eventChan) == cap(mf.eventChan)
 }
+
+func getbaseURL(dda *datadoghqv1alpha1.DatadogAgent) string {
+	if dda.Spec.Agent.Config.DDUrl != nil {
```
ahmed-mez, line 664: this `dda.Spec.Agent` can be `nil`

pkg/controller/utils/datadog/metrics_forwarder.go
```diff
@@ -200,6 +203,8 @@ func (mf *metricsForwarder) connectToDatadogAPI() (bool, error) {
 	}
 	mf.logger.Info("Getting Datadog credentials")
 	apiKey, appKey, err := mf.getCredentials(dda)
+	mf.logger.Info("Getting Datadog Site")
+	mf.baseURL = getbaseURL(dda)
```
ahmed-mez, line 207: nit
```suggestion
	mf.baseURL = getbaseURL(dda)
	mf.logger.Info("Got Datadog Site", "site", mf.baseURL)
```

thehunt33r:
@ahmed-mez Will get to it now :)

pkg/controller/utils/datadog/metrics_forwarder_test.go
```diff
@@ -1029,3 +1029,49 @@ func Test_metricsForwarder_getSecretsFromCache(t *testing.T) {
 		})
 	}
 }
+
+func Test_getbaseURL(t *testing.T) {
```
thehunt33r, line 1033: @ahmed-mez Added the unit test. 

However I thought of something while doing this. Typically, `ddUrl` would be a custom URL for the Datadog endpoint (`app.datadoghq.com`). However, the API client uses by default `api.datadoghq.com`. 

Therefore it might be best to only use Site and totally disregard `ddUrl` to set the API endpoint for the Metrics Forwarder. What do you think?

ahmed-mez: Thanks, what you said is true most of the time the difference is either `.com` or `.eu` but in some advanced cases we might want to use a completely different url. So if `dda.Spec.Agent.Config.DDUrl` is set we need to use it to send operator metrics to the same backend as the agent. (the current logic in your PR)

pkg/controller/utils/datadog/metrics_forwarder_test.go
```diff
@@ -1029,3 +1029,49 @@ func Test_metricsForwarder_getSecretsFromCache(t *testing.T) {
...
+			name: "Compute baseURL from ddUrl when Site is not defined",
+			args: args{
+				dda: test.NewDefaultedDatadogAgent("foo", "bar", &test.NewDatadogAgentOptions{
+					NodeAgentConfig: &datadoghqv1alpha1.NodeAgentConfig{
+						DDUrl: datadoghqv1alpha1.NewStringPointer("https://test.url.com"),
+					}}),
+			},
+			want: "https://test.url.com",
+		},
+		// TODO: Add test cases.
```
ahmed-mez, line 1068: nit: can we add a test case where both `Site` and `DDUrl` are explicitly set to test the order of precedence?

pkg/controller/datadogagent/datadogagent_controller.go
```diff
@@ -103,7 +103,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler, registerFunc func(datadog.
 	// Watch for changes to primary resource DatadogAgent
 	err = c.Watch(&source.Kind{Type: &datadoghqv1alpha1.DatadogAgent{}}, &handler.EnqueueRequestForObject{}, onCreate)
 	if err != nil {
-		return err
+	return err
```
ahmed-mez, line 106: why?

pkg/controller/utils/datadog/metrics_forwarder.go
```diff
@@ -653,3 +659,12 @@ func (mf *metricsForwarder) isErrChanFull() bool {
...
 	return len(mf.eventChan) == cap(mf.eventChan)
 }
+
+func getbaseURL(dda *datadoghqv1alpha1.DatadogAgent) string {
+	if dda.Spec.Agent != nil && dda.Spec.Agent.Config.DDUrl != nil {
+		return *dda.Spec.Agent.Config.DDUrl
+	} else if dda.Spec.Site != "" {
+		return fmt.Sprintf("https://api.%s", dda.Spec.Site)
+	}
+	return "https://api.datadoghq.com"
```
ahmed-mez, line 669:
```suggestion
	return defaultbaseURL
```

pkg/controller/utils/datadog/metrics_forwarder_test.go
thehunt33r: Done

pkg/controller/datadogagent/datadogagent_controller.go
thehunt33r: I'm not sure how that happened. Anyhow, I undid the change. Good spot!

pkg/controller/utils/datadog/metrics_forwarder.go
thehunt33r: Done :)

ahmed-mez:
thanks @thehunt33r 
happy to approve and merge this once the CI is all green, the current problem seems caused by a missing `}` in the metrics forwarder file

thehunt33r:
🤦 whoops. Addressed.

pkg/apis/datadoghq/v1alpha1/test/new.go
```diff
@@ -86,6 +87,7 @@ func NewDefaultedDatadogAgent(ns, name string, options *NewDatadogAgentOptions)
 			Log:                datadoghqv1alpha1.LogSpec{},
 			Process:            datadoghqv1alpha1.ProcessSpec{},
 		},
+		Site: options.Site,
```
ahmed-mez, line 90: this causes a panic when `options` is `nil`
it should be moved to the block below when we check `if options != nil`

pkg/apis/datadoghq/v1alpha1/test/new.go
```diff
@@ -104,6 +105,7 @@ func NewDefaultedDatadogAgent(ns, name string, options *NewDatadogAgentOptions)
 		}
 
 		ad.Spec.Agent.DaemonsetName = options.AgentDaemonsetName
+		ad.Spec.Site = options.Site
```
thehunt33r, line 108: @ahmed-mez Addressed here

pkg/apis/datadoghq/v1alpha1/test/new.go
thehunt33r: 🤦 Should have caught it earlier. Thank you!

PR #100, opened 2020-05-26
Add hostPort for Agent

thehunt33r:
### What does this PR do?

This PR adds the `hostPort` parameter to the Agent config, which allows exposing the Dogstatsd port to the cluster, and therefore receiving custom metrics from our applications.

### Motivation

I couldn't expose port 8125 to my applications without modifying manually the DaemonSet definition.

### Additional Notes

1. I saw that the `newDaemonSetfromInstance` method is not tested in `agent_test.go`. Therefore I added the test case to the `newExtendedDaemonSetfromInstance` method tests. I' m not sure whether this is an oversight or a design decision in an effort to reduce the amount of duplicated code, but I'm happy to write tests for `newDaemonSetFromInstance` if needed.
2. I'm also not sure if the lack of `hostPort` for Agent was by design. I saw that the APM Agent already had that setting. Please let me know if this isn't supposed to be how we should expose port 8125 to our applications.

Thanks!

Matthieu

### DIFF
--- a/pkg/apis/datadoghq/v1alpha1/const.go
+++ b/pkg/apis/datadoghq/v1alpha1/const.go
@@ -52,7 +52,7 @@ const (
 	DDLogsConfigContainerCollectAll  = "DD_LOGS_CONFIG_CONTAINER_COLLECT_ALL"
 	DDLogsContainerCollectUsingFiles = "DD_LOGS_CONFIG_K8S_CONTAINER_USE_FILE"
 	DDDogstatsdOriginDetection       = "DD_DOGSTATSD_ORIGIN_DETECTION"
-	DDDogstatsdPort                  = "DD_DOGSTATD_PORT"
+	DDDogstatsdPort                  = "DD_DOGSTATSD_PORT"
 	DDClusterAgentEnabled            = "DD_CLUSTER_AGENT_ENABLED"
 	DDClusterAgentKubeServiceName    = "DD_CLUSTER_AGENT_KUBERNETES_SERVICE_NAME"
 	DDClusterAgentAuthToken          = "DD_CLUSTER_AGENT_AUTH_TOKEN"
--- a/pkg/apis/datadoghq/v1alpha1/const.go
+++ b/pkg/apis/datadoghq/v1alpha1/const.go
@@ -52,6 +52,7 @@ const (
 	DDLogsConfigContainerCollectAll  = "DD_LOGS_CONFIG_CONTAINER_COLLECT_ALL"
 	DDLogsContainerCollectUsingFiles = "DD_LOGS_CONFIG_K8S_CONTAINER_USE_FILE"
 	DDDogstatsdOriginDetection       = "DD_DOGSTATSD_ORIGIN_DETECTION"
+	DDDogstatsdPort                  = "DD_DOGSTATD_PORT"
 	DDClusterAgentEnabled            = "DD_CLUSTER_AGENT_ENABLED"
 	DDClusterAgentKubeServiceName    = "DD_CLUSTER_AGENT_KUBERNETES_SERVICE_NAME"
 	DDClusterAgentAuthToken          = "DD_CLUSTER_AGENT_AUTH_TOKEN"
--- a/pkg/apis/datadoghq/v1alpha1/const.go
+++ b/pkg/apis/datadoghq/v1alpha1/const.go
@@ -31,6 +31,8 @@ const (
 	DefaultClusterAgentServicePort = 5005
 	// DefaultMetricsServerServicePort default metrics-server port
 	DefaultMetricsServerServicePort = 443
+	// DefaultDogstatsdPort default dogstatsd port
+	DefaultDogstatsdPort = 8125
 )
 
 // Datadog env var names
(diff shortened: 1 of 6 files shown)


pkg/controller/datadogagent/utils.go
```diff
@@ -159,6 +159,17 @@ func getAgentContainer(dda *datadoghqv1alpha1.DatadogAgent) (*corev1.Container,
 	if err != nil {
 		return nil, err
 	}
+
+	udpPort := corev1.ContainerPort{
+		ContainerPort: datadoghqv1alpha1.DefaultDogstatsdPort,
```
clamoriniere, line 164: It will also required to add the `DD_DOGSTATSD_PORT` env-var to the "agent" container to override the default value if never the value provided it not "8125"

thehunt33r: Do we necessarily want to have `DD_DOGSTATSD_PORT` match the `hostPort` ? 
I think we should definitely support this, but this PR was intended more towards setting `hostPort` (the port that is bound on the host). 

I can see cases where you'd want to have `hostPort=18125` for example but not need to change the `DD_DOGSTATSD_PORT`, since it's "abstracted-away" in the container.

clamoriniere: The problem comes when the user also set: `hostNetwork`

thehunt33r: @clamoriniere Implemented this.
I'm now thinking we could only add that variable when both `hostPort` and `hostNetwork` are set. What do you think?

pkg/apis/datadoghq/v1alpha1/const.go
```diff
@@ -50,6 +52,7 @@ const (
 	DDLogsConfigContainerCollectAll  = "DD_LOGS_CONFIG_CONTAINER_COLLECT_ALL"
 	DDLogsContainerCollectUsingFiles = "DD_LOGS_CONFIG_K8S_CONTAINER_USE_FILE"
 	DDDogstatsdOriginDetection       = "DD_DOGSTATSD_ORIGIN_DETECTION"
+	DDDogstatsdPort                  = "DD_DOGSTATD_PORT"
```
ahmed-mez, line 55: typo
```suggestion
	DDDogstatsdPort                  = "DD_DOGSTATSD_PORT"
```

thehunt33r: @ahmed-mez Thanks!

pkg/controller/datadogagent/utils.go
clamoriniere: Yes,  this is is a good idea.


```golang
	udpPort := corev1.ContainerPort{
		ContainerPort: datadoghqv1alpha1.DefaultDogstatsdPort,
                 Name:          "dogstatsdport",
		Protocol:      corev1.ProtocolUDP,
	}
	if agentSpec.Config.HostPort != nil && datadoghqv1alpha1.BoolValue(agentSpec.Config.HostNetwork) {
	    udpPort.ContainerPort = *agentSpec.Config.HostPort
	    // TODO add also the envvar
        }

Comment (thehunt33r) [reply pkg/controller/datadogagent/utils.go:164]: @clamoriniere Hey! Could you review my changes? I've implemented something similar to your suggestion.

Comment (clamoriniere) [reply pkg/controller/datadogagent/utils.go:164]: Hi @thehunt33r
sorry for the delay. 
@ahmed-mez could you make a final review?

File: pkg/controller/datadogagent/utils.go [hunk 3]
```diff
@@ -159,6 +159,24 @@ func getAgentContainer(dda *datadoghqv1alpha1.DatadogAgent) (*corev1.Container,
...
+	if agentSpec.Config.HostPort != nil {
+		udpPort.HostPort = *agentSpec.Config.HostPort
+		if agentSpec.HostNetwork {
+			udpPort.ContainerPort = *agentSpec.Config.HostPort
+			envVars = append(envVars, corev1.EnvVar{
+				Name:  datadoghqv1alpha1.DDDogstatsdPort,
+				Value: strconv.Itoa(int(*agentSpec.Config.HostPort)),
+			})
+		}
+	}
```
ahmed-mez, line 178: would be nice to have a comment explaining why we're doing this

thehunt33r: Agreed, will add it now! :)

pkg/controller/datadogagent/utils.go
```diff
@@ -159,6 +159,28 @@ func getAgentContainer(dda *datadoghqv1alpha1.DatadogAgent) (*corev1.Container,
...
+		Name:          "dogstatsdport",
+		Protocol:      corev1.ProtocolUDP,
+	}
+
+	if agentSpec.Config.HostPort != nil {
+		// Create the host port configuration
+		udpPort.HostPort = *agentSpec.Config.HostPort
+		// If HostNetwork is enabled, set the container port
+		// and the DD_DOGSTATSD_PORT environment variable to match the host port
+		// so that Dogstatsd can be reached on the port configured in HostPort
```
thehunt33r, line 174: @ahmed-mez Is this descriptive enough? Or maybe a bit too much? Let me know if you want any changes there :)

ahmed-mez: it looks good, thanks 💯

PR #102, opened 2020-05-26
Fix DatadogMetric generation with operator-sdk 0.17

clamoriniere:
### What does this PR do?

The CRD generator doesn't support `float64` as a type of a CRD field.
The suggested workaround is to used instead `resource.Quantity` but we choose to use `string`.
Related issue: https://github.com/kubernetes-sigs/controller-tools/issues/245

### Motivation

Fix the Master branch, that is broken due to a merge issue.

### Additional Notes


clamoriniere:
@vboulineau let me know if this change didn't impact to much the functionality inside the cluster-agent.

PR #109, opened 2020-06-05
Use 8443 as default port for the external metrics provider

ahmed-mez:
### What does this PR do?

- Use 8443 as default port for the external metrics provider

### Motivation

443 doesn't work OOTB on openshift


ahmed-mez:
Custom metrics documentation is available here https://github.com/DataDog/datadog-operator/blob/master/docs/cluster_agent_setup.md

PR #122, opened 2020-07-16
Propagate custom check definitions to CLC Runners

juliogreff:
It's already possible to define`.Spec.Agent.Config.Checksd`, but this option is not propagated to
the Clusterchecks Runner. As a runner is basically an agent, we're just
reusing the same configuration.

This applies the same changes as this proposed change to
stable/datadog's Helm chart: https://github.com/helm/charts/pull/23139

### DIFF
--- a/pkg/controller/datadogagent/clusterchecksrunner_test.go
+++ b/pkg/controller/datadogagent/clusterchecksrunner_test.go
@@ -23,6 +23,32 @@ func clusterChecksRunnerDefaultPodSpec() corev1.PodSpec {
 	return corev1.PodSpec{
 		Affinity:           getPodAffinity(nil, "foo-cluster-checks-runner"),
 		ServiceAccountName: "foo-cluster-checks-runner",
+		InitContainers: []corev1.Container{
+			{
+				Name:            "init-volume",
+				Image:           "datadog/agent:latest",
+				ImagePullPolicy: corev1.PullIfNotPresent,
+				Resources:       corev1.ResourceRequirements{},
+				Command:         []string{"bash", "-c"},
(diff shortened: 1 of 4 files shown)


pkg/controller/datadogagent/clusterchecksrunner_test.go
```diff
@@ -40,6 +66,20 @@ func clusterChecksRunnerDefaultPodSpec() corev1.PodSpec {
 
 func clusterChecksRunnerDefaultVolumeMounts() []corev1.VolumeMount {
 	return []corev1.VolumeMount{
+		{
+			Name:      datadoghqv1alpha1.ConfdVolumeName,
```
clamoriniere, line 70: I think we don't need to mound the Confd volume, because the `ClusterCheckRunner` only get the config from the `ClusterAgent`.

pkg/controller/datadogagent/utils.go
```diff
@@ -771,6 +741,52 @@ func getVolumesForAgent(dda *datadoghqv1alpha1.DatadogAgent) []corev1.Volume {
...
+	}
+	if dda.Spec.Agent.Config.Confd != nil {
+		confdVolumeSource = corev1.VolumeSource{
+			ConfigMap: &corev1.ConfigMapVolumeSource{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: dda.Spec.Agent.Config.Confd.ConfigMapName,
+				},
+			},
+		}
+	}
```
clamoriniere, line 757: You are using `getVolumeForConfig()` for the `Agent` and the `ClusterCheckRunner`, but for the  "runner" we don't need to mount the "Confd" `confimap` since all the check configs will come from the `cluster-agent`

PR #124, opened 2020-07-27
Update to operator-sdk 0.19.0

vboulineau:
### What does this PR do?

Update to operator-sdk 0.19.0 - finally fixing `--stacktrace-level`

### Additional Notes


chart/datadog-operator/templates/deployment.yaml
```diff
@@ -48,7 +48,7 @@ spec:
           args:
             - --zap-level={{ .Values.logLevel }}
             - --zap-encoder=console
-            # - --zap-stacktrace-level=error // TODO: uncomment after releasing v0.3.0
+            # - --zap-stacktrace-level=panic // TODO: uncomment after releasing v0.3.0
```
CharlyF, line 51: When will we remove the comment?

vboulineau: When we release a new official Docker image (otherwise this parameter will break people using the chart with current official image)

PR #137, opened 2020-08-20
[external metrics] Add mgmt of missing resources for metrics provider

xornivore:
### What does this PR do?

- Adds missing `ClusterRole` and `ClusterRoleBinding` to allow reading external metrics from HPA.
- Adds missing `APIService` registration for `v1beta1.external.metrics.k8s.io`.
- Operator vendoring and `ClusterRole` changes to allow management of `APIService` resources.

### Motivation

Configuration for enabling external metrics is currently not working out of the box for `DatagogAgent` resources.

### Additional Notes

To test - make sure the `DatadogAgent` CRD is configured to enable this feature:

```
kind: DatadogAgent
metadata:
  name: datadog-agent
spec:
...
  agent:
    image:
      name: 'datadog/agent:latest'
...
  clusterAgent:
    image:
      name: 'datadog/cluster-agent:latest'
    config:
      externalMetrics:
        enabled: true
        useDatadogMetrics: true
```
Ensure that the metrics can be consumed by pod autoscalers in the cluster and proper RBAC and APIService resource is created.

On cleanup of the `DatadogAgent` resource make sure that these resources get cleaned up as well.


pkg/controller/datadogagent/clusteragent_metricsserver.go
```diff
@@ -43,34 +41,95 @@ func buildMetricsServerClusterRoleBinding(dda *datadoghqv1alpha1.DatadogAgent, n
...
-		r.recordEvent(dda, event)
-		if err := r.client.Delete(context.TODO(), clusterRoleBinding); err != nil {
-			if errors.IsNotFound(err) {
-				return reconcile.Result{}, nil
-			}
-			return reconcile.Result{}, err
+func (r *ReconcileDatadogAgent) createHPAClusterRoleBinding(logger logr.Logger, dda *datadoghqv1alpha1.DatadogAgent, name, agentVersion string) (reconcile.Result, error) {
+	clusterRoleBinding := buildMetricsServerClusterRoleBinding(dda, name, agentVersion)
+	if clusterRoleBinding == nil {
+		return reconcile.Result{}, nil
```
CharlyF, line 47: Could we log something here ?

deploy/clusterrole.yaml
```diff
@@ -51,3 +51,13 @@ rules:
   - 'cronjobs'
   verbs:
   - 'get'
+- apiGroups:
+  - apiregistration.k8s.io
+  resources:
+  - apiservices
+  verbs:
+  - get
```
CharlyF, line 59: Do we not `delete` the apiservice ? If so, we'll need to add the verb here.

pkg/controller/datadogagent/service.go
```diff
@@ -149,6 +177,11 @@ func (r *ReconcileDatadogAgent) cleanupMetricsServerService(dda *datadoghqv1alph
 	return cleanupService(r.client, serviceName, dda.Namespace)
 }
 
+func (r *ReconcileDatadogAgent) cleanupMetricsServerAPIService(dda *datadoghqv1alpha1.DatadogAgent) (reconcile.Result, error) {
+	apiServiceName := getMetricsServerServiceName(dda)
```
CharlyF, line 181: same as the createAPIService, could we add a log:
`logger.Info("Delete APIService", "name", apiServiceName)` ? (maybe below in case it's not found)

deploy/clusterrole.yaml
xornivore: Yes, I missed this one, it's odd that it's able to delete it.

pkg/controller/datadogagent/clusteragent_metricsserver.go
xornivore: Definitely though what concern are we would be addressing here?

pkg/controller/datadogagent/service.go
xornivore: Yes, will add these.

xornivore:
Re: 
> Have we considered if the usage of certain libs v1 would work with all expected versions of kube ?
Good question, I'll check we need to use a different version in some case. What's the minimum version we support for the operator?

pkg/controller/datadogagent/clusteragent_metricsserver.go
CharlyF: It's just a nit, but if we fail to build the CRB, there is no error surfaced https://github.com/DataDog/datadog-operator/pull/137/files#diff-bbe5868bca18c548b30bed831ffd53acR20-R42 so a log could help imo

deploy/clusterrole.yaml
CharlyF: That's what we do in this code path correct https://github.com/DataDog/datadog-operator/pull/137/files#diff-7554a7176dee880f102333a54563b582R255 ?

pkg/controller/datadogagent/clusteragent_metricsserver.go
xornivore: So the way it can be nil is if `!isMetricsProviderEnabled`, so we are just guarding trying to create it by that. This is not a new func btw. The diff is a bit off on this file: https://github.com/DataDog/datadog-operator/pull/137/files#diff-bbe5868bca18c548b30bed831ffd53acL67

juliogreff:
@xornivore re kubernetes version support, it seems we require >= 1.14, but things "should work on versions >= 1.11": https://github.com/DataDog/datadog-operator/blob/master/docs/getting_started.md#prerequisites

pkg/controller/datadogagent/service.go
```diff
@@ -112,6 +113,28 @@ func (r *ReconcileDatadogAgent) manageMetricsServerService(logger logr.Logger, d
 	return r.updateIfNeededMetricsServerService(logger, dda, service)
 }
 
+func (r *ReconcileDatadogAgent) manageMetricsServerAPIService(logger logr.Logger, dda *datadoghqv1alpha1.DatadogAgent) (reconcile.Result, error) {
+	if dda.Spec.ClusterAgent == nil {
+		return r.cleanupMetricsServerAPIService(logger)
+	}
+
+	if !isMetricsProviderEnabled(dda.Spec.ClusterAgent) {
+		return reconcile.Result{}, nil
```
juliogreff, line 122: Should't we also cleanup the APIService if the metrics provider is not enabled?

Also, `isMetricsProviderEnabled` already checks for a nil spec, so maybe those two conditionals can be just one that checks for `isMetricsProviderEnabled`.

PR #3, opened 2019-11-22
[ADD] support of APM and Process Agents

clamoriniere:
### What does this PR do?

Add support of APM and Process Agents:

* Update API for APM and Process configuration
* Update `PodTemplate` generation
* Add unit-tests
* Add end2end tests
* Add Github worflows action
* Add Codecov support

### Motivation

What inspired you to submit this pull request?

### Additional Notes

Anything else we should know when reviewing?

PR #139, opened 2020-08-27
Remove DD_HOSTNAME from agent and move it back to check runner

carloscastrojumo:
### What does this PR do?

Remove `DD_HOSTNAME` as environment variable from DataDog agent and put it back to check runner.

### Motivation

`DD_HOSTNAME` was removed from check runner and added by mistake in the agent in this PR [#3](https://github.com/DataDog/datadog-operator/pull/3)

### Additional Notes

Add dependency for `bin/operator-sdk` in `make container` and `make e2e` commands.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ build-plugin: ${ARTIFACT_PLUGIN}
 ${ARTIFACT_PLUGIN}: ${SOURCES}
 	CGO_ENABLED=0 go build -i -installsuffix cgo ${LDFLAGS} -o ${ARTIFACT_PLUGIN} ./cmd/${ARTIFACT_PLUGIN}/main.go
 
-container:
+container: bin/operator-sdk
 	./bin/operator-sdk build $(PREFIX):$(TAG)
     ifeq ($(KINDPUSH), true)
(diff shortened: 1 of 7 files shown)


ahmed-mez:
Hi @carloscastrojumo thanks for the contribution.

Actually `DD_HOSTNAME` should be set explicitly only for the `datadog-cluster-checks-runner`. (see the [helm chart](https://github.com/DataDog/helm-charts/tree/master/charts/datadog) for reference)

We shouldn't set the hostname for the `datadog-agent` and it's the case currently because of an oversight I believe, introduced by mistake in the refactoring done by this PR https://github.com/DataDog/datadog-operator/pull/3

The agent has automatic mechanisms to discover the node name, the option should be only used in particular cases - that's why the option exists.

Would you be interested in fixing it by removing `DD_HOSTNAME` from the agent, and move it to the check runner only?

Thanks!

carloscastrojumo:
Hi @ahmed-mez thank you for the review.

As suggested, updated my PR to remove the `DD_HOSTNAME` from agent and cluster agent and just introduce it back to check runner.

ahmed-mez:
@carloscastrojumo that's great! could you please update the PR's title and description accordingly? thanks!

PR #142, opened 2020-08-28
Support wpa configuration

ahmed-mez:
### What does this PR do?

Add a dedicated field in the CRD and the required rbacs for WPA


pkg/apis/datadoghq/v1alpha1/datadogagent_types.go
```diff
@@ -664,6 +664,11 @@ type ExternalMetricsConfig struct {
 	// +optional
 	Enabled bool `json:"enabled,omitempty"`
 
+	// Enable informer and controller of the watermark pod autoscaler
+	// NOTE: You need to install the `WatermarkPodAutoscaler` CRD before
```
clamoriniere, line 668: maybe something more generic: the WatermarkPodAutoscaler-Controller (https://github.com/DataDog/watermarkpodautoscaler) need to be install

ahmed-mez: good point, done 👍

clamoriniere (approved):

pkg/controller/datadogagent/datadogagent_controller_test.go
```diff
@@ -2103,6 +2165,34 @@ func hasAllClusterLevelRbacResources(policyRules []rbacv1.PolicyRule) bool {
...
+				groupFound = true
+			}
+		}
+		if reflect.DeepEqual(policyRule.Verbs, requiredVerbs) {
+			verbsFound = true
+		}
+	}
+
+	return resourceFound && groupFound && verbsFound
+}
```
L3n41c, line 2194: It looks like this function will return `true` even for something like:
```yaml
- apiGroups:
  - "datadoghq.com"
  resources:
  - "datadogagent"
  verbs:
  - "create"
- apiGroups:
  - "something.unrelated.to.datadog"
  resources:
  - "watermarkpodautoscalers"
  verbs:
  - "update"
- apiGroups:
  - ""
  resources:
  - "pods"
  verbs:
  - "list"
  - "get"
  - "watch"
```
whereas we expect it to return `true` only when all the required elements are in the same list item:
```yaml
- apiGroups:
  - "datadoghq.com"
  resources:
  - "watermarkpodautoscalers"
  verbs:
  - "list"
  - "get"
  - "watch"
```

```suggestion
func hasWpaRbacs(policyRules []rbacv1.PolicyRule) bool {
	requiredVerbs := []string{
		datadoghqv1alpha1.ListVerb,
		datadoghqv1alpha1.WatchVerb,
		datadoghqv1alpha1.GetVerb,
	}

	for _, policyRule := range policyRules {
		resourceFound := false
		groupFound := false
		verbsFound := false

		for _, resource := range policyRule.Resources {
			if resource == "watermarkpodautoscalers" {
				resourceFound = true
				break
			}
		}
		for _, group := range policyRule.APIGroups {
			if group == "datadoghq.com" {
				groupFound = true
				break
			}
		}
		if reflect.DeepEqual(policyRule.Verbs, requiredVerbs) { // Test to be updated
			verbsFound = true
			break
		}
		
		if resourceFound && groupFound && verbsFound {
			return true
		}
	}

	return false
}
```

pkg/controller/datadogagent/datadogagent_controller_test.go
```diff
@@ -2103,6 +2165,34 @@ func hasAllClusterLevelRbacResources(policyRules []rbacv1.PolicyRule) bool {
...
+			}
+		}
+		for _, group := range policyRule.APIGroups {
+			if group == "datadoghq.com" {
+				groupFound = true
+			}
+		}
+		if reflect.DeepEqual(policyRule.Verbs, requiredVerbs) {
+			verbsFound = true
+		}
```
L3n41c, line 2190: This will make the function return false in the cases where we have more privileges than expected (i.e.: not only `get`, `list` and `watch` but also something else or even `*`.)

L3n41c (changes requested):

pkg/controller/datadogagent/datadogagent_controller_test.go
ahmed-mez: yes that's true, it's a util function for a test case though I wanted to leave it as dumb as possible - I think we can keep it like this and extended if needed in other test cases in the future

L3n41c (approved):

PR #144, opened 2020-09-01
Fix pod anti affinity

ahmed-mez:
### What does this PR do?

- Make sure we set the correct labels for CLC runner pod anti affinity
- Remove the default pod anti affinity from the DCA, it's not required - two DCAs can live on the same node

### Motivation

Before this fix two CLC runners can be scheduled on the same node, which messes up the clc dispatching

### Additional Notes

Can be easily tested on a 1 node cluster


pkg/controller/datadogagent/clusteragent_test.go
```diff
@@ -31,7 +31,7 @@ var testClusterAgentReplicas int32 = 1
 
 func clusterAgentDefaultPodSpec() corev1.PodSpec {
 	return corev1.PodSpec{
-		Affinity:           getPodAffinity(nil, "foo-cluster-agent"),
+		Affinity:           nil,
```
clamoriniere, line 34: is it necessary?

clamoriniere:
```suggestion
```

clamoriniere (approved):

pkg/controller/datadogagent/clusteragent_test.go
ahmed-mez: no

PR #149, opened 2020-09-04
[system-probe] Add missing system probe settings for parity with helm chart

xornivore:
### What does this PR do?

Adds missing toggles for `OOMKill` and `TCPQueueLength` modules as well as `collectDNSStats` setting.

### Motivation

Parity with the helm chart in terms of full set of system probe features.

### Additional Notes

Adding unit tests for these and similar features (hello security agent!) becomes rather tedious and results in rather repetitive code, there might be some quick wins in terms of refactoring to use more helper functions and injecting additional volumes, mounts. It would be ideal to keep that refactoring separate from these changes here.

### Describe your test plan

Configure `DatadogAgent` CR with desired features and make sure they are properly enabled:

```yaml
kind: DatadogAgent
metadata:
  name: datadog-agent
spec:
# ...
  agent:
    # ...
    systemProbe:
      enabled: true
      enableOOMKill: true
      enableTCPQueueLength: true
   # ...
```


clamoriniere (approved):

pkg/apis/datadoghq/v1alpha1/const.go
```diff
@@ -145,6 +145,10 @@ const (
 	SystemProbeAgentSecurityVolumePath = "/etc/config"
 	SystemProbeSecCompRootVolumeName   = "seccomp-root"
 	SystemProbeSecCompRootVolumePath   = "/host/var/lib/kubelet/seccomp"
+	SystemProbeLibModulesVolumeName    = "modules"
+	SystemProbeLibModulesVolumePath    = "/lib/modules"
```
clamoriniere, line 149: I think we are missing the "kernelHeader folder"
@L3n41c could you confirm?

L3n41c: Well, the kernel headers are looked for in `/lib/modules/$(uname -r)/build` which is generally a symlink to `/usr/src/kernel-$(uname-r)`.
So, mounting the `modules` and `src` volumes defined here should be fine.
👍🏻

L3n41c: Well, the kernel headers are looked for in `/lib/modules/$(uname -r)/build` which is generally a symlink to `/usr/src/kernel-$(uname-r)`.
So, mounting the `modules` and `src` volumes defined here should be fine.
👍🏻

L3n41c (approved):

PR #151, opened 2020-09-15
Fix RBAC with compliance and admission controller

clamoriniere:
### What does this PR do?

Fix missing Operator RBAC permission when the DCA is activated:
* Create DCA RBAC even if the ExternalMetrics service is not enabled
* Add missing DCA RBAC for Admission-webhook 
* Add missing DCA RBAC compliance
* Add missing DCA RBAC for orchestrator app: cluster-id detection.

### Motivation

* Solve several permission error in the Operator for the creation of the DCA RBAC
* Fix DCA feature due to missing RBAC.

### Additional Notes

### Describe your test plan

Use the `examples/datadog-agent-all.yaml` and verify that the Operator and the DCA pod logs don't contain RBAC error.

To test the DCA deployment without ExternalMetric, you can change the ClusterAgent configuration by:
```yaml
  clusterAgent:
    image:
      name: "datadog/cluster-agent:latest"
    config:
      externalMetrics:
        enabled: false
```


chart/datadog-operator/templates/clusterrole.yaml
```diff
@@ -57,4 +72,166 @@ rules:
...
+  - admissionregistration.k8s.io
+  resources:
+  - mutatingwebhookconfigurations
+  - secrets
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
```
ahmed-mez, line 163: why do we define these twice? we have them defined above

clamoriniere: Also the "secrets" here look wrong.

clamoriniere: wrong cut/past

ahmed-mez: same thing in deploy/clusterrole

examples/datadog-agent-all.yaml
```diff
@@ -18,3 +18,16 @@ spec:
     systemProbe:
       enabled: true
       bpfDebugEnabled: true
+    security:
```
xornivore, line 21: 👍

xornivore (approved):

chart/datadog-operator/templates/clusterrole.yaml
```diff
@@ -57,4 +61,165 @@ rules:
...
   - "get"
   - "watch"
+- apiGroups:
+  - autoscaling
+  resources:
+  - horizontalpodautoscalers
+  verbs:
+  - list
+  - watch
+## To fix: should be role not a clusterrole
```
clamoriniere, line 71: All those configmap are only accessed by the DCA and should be in the DCA namespace. So I think we can move it in the role.

examples/datadog-agent-all.yaml
```diff
@@ -18,3 +18,16 @@ spec:
...
+      runtime:
+        enabled: false
+  clusterAgent:
+    image:
+      name: "datadog/cluster-agent:latest"
+    config:
+      externalMetrics:
+        enabled: true
+      admissionController:
+        enabled: true
```
xornivore, line 33: nit: new line missing

pkg/controller/datadogagent/clusteragent.go
```diff
@@ -622,18 +622,25 @@ func (r *ReconcileDatadogAgent) manageClusterAgentRBACs(logger logr.Logger, dda
...
 	// Create or update ClusterRole
 	clusterRole := &rbacv1.ClusterRole{}
 	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: rbacResourcesName}, clusterRole); err != nil {
 		if errors.IsNotFound(err) {
 			return r.createClusterAgentClusterRole(logger, dda, rbacResourcesName, clusterAgentVersion)
 		}
 		return reconcile.Result{}, err
-	}
-	if result, err := r.updateIfNeededClusterAgentClusterRole(logger, dda, rbacResourcesName, clusterAgentVersion, clusterRole); err != nil {
+	} else if result, err := r.updateIfNeededClusterAgentClusterRole(logger, dda, rbacResourcesName, clusterAgentVersion, clusterRole); err != nil {
```
xornivore, line 641: Is this change functional or cosmetic here (joining the two checks into `else if`)? Personally I prefer to avoid else if when the conditions are complex like this and there are error checks involved - it makes it rather hard to follow and may make the code susceptible to misinterpretation and refactoring leading to bugs.

pkg/controller/datadogagent/clusteragent.go
```diff
@@ -646,7 +653,29 @@ func (r *ReconcileDatadogAgent) manageClusterAgentRBACs(logger logr.Logger, dda
...
 	}
-
+	// Create or update Role
+	role := &rbacv1.Role{}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: rbacResourcesName, Namespace: dda.Namespace}, role); err != nil {
+		if errors.IsNotFound(err) {
+			return r.createClusterAgentRole(logger, dda, rbacResourcesName, clusterAgentVersion)
+		}
+		return reconcile.Result{}, err
+	} else if result, err := r.updateIfNeededClusterAgentRole(logger, dda, rbacResourcesName, clusterAgentVersion, role); err != nil {
```
xornivore, line 663: Similar to the above comment about `else if`.

pkg/controller/datadogagent/clusteragent.go
```diff
@@ -1070,25 +1075,25 @@ func buildClusterAgentClusterRole(dda *datadoghqv1alpha1.DatadogAgent, name, age
 				datadoghqv1alpha1.UpdateVerb},
 		})
 
-		// Replicasets
+		// ExtendedDaemonsetReplicaSets
```
ahmed-mez, line 1078: do we need this?

clamoriniere: Yes for pod, managed by an extendeddaemonset

clamoriniere: it is to be futur proof :)

pkg/controller/datadogagent/clusteragent.go
clamoriniere: agree, I have reverted this change

vboulineau (approved):

PR #155, opened 2020-09-23
Always mount `config` volume when security agent is enabled

juliogreff:
If only the runtime security agent is enabled, the process won't be able
to start because it can't find the `datadog.yaml` that's available in
the `config` volume, as it was only mounted when the compliance agent
was enabled. Now, it's mounted if either of them is enabled.

### Describe your test plan

This can be tested by enabling the security agent in three different permutations (only runtime, only compliance, and both), and checking that the process starts successfully.


ahmed-mez (approved):
could we add a unit test? lgtm otherwise

xornivore (approved):
👍

PR #159, opened 2020-10-13
Update `verify-pr-label-action` GH action to allow contributions from fork

vboulineau:
### What does this PR do?

See https://github.com/jesusvasquez333/verify-pr-label-action/issues/9#issuecomment-706639627

### Motivation

### Additional Notes

### Describe your test plan

### DIFF
--- a/.github/workflows/pr-linter.yml
+++ b/.github/workflows/pr-linter.yml
@@ -2,6 +2,8 @@ name: pull request linter
 on:
   pull_request:
    types: [opened, labeled, unlabeled, synchronize]
+  pull_request_target:
+   types: [opened, labeled, unlabeled, synchronize]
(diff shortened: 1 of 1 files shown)


ahmed-mez (approved):

.github/workflows/pr-linter.yml
```diff
@@ -1,6 +1,6 @@
 name: pull request linter
 on:
-  pull_request:
+  pull_request_target:
```
clamoriniere, line 3: I think we need both `pull_request` and `pull_request_target`. because for example this pull request wasn't check.

vboulineau: It was not supposed to be the case, checking.

PR #165, opened 2020-11-02
Re-add handling of WATCH_NAMESPACE following migration to Operator SDK 1.0

vboulineau:
### What does this PR do?

Re-add handling of WATCH_NAMESPACE following migration to Operator SDK 1.0.

### Motivation

### Additional Notes

### Describe your test plan

Setting WATCH_NAMESPACE should limit the watches set by Datadog Operator. Deploy Datadog Operator with `WATCH_NAMESPACE=<ns1>` and create two DatadogAgent objects in `ns1` and `ns2`. Datadog Operator should ignore `ns2`.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,0 +1,56 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2019 Datadog, Inc.
+
+package config
+
+import (
+	"os"
+	"strings"
+
+	"github.com/go-logr/logr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+)
+
+const (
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -0,0 +1,56 @@
...
+	return []string{ns}
+}
+
+// ManagerOptionsWithNamespaces returns an updated Options with namespaces information
+func ManagerOptionsWithNamespaces(logger logr.Logger, opt ctrl.Options) ctrl.Options {
+	namespaces := GetWatchNamespaces()
+	switch {
+	case len(namespaces) == 0:
+		logger.Info("Manager will watch and manage resources in all namespaces")
+	case len(namespaces) == 1:
```
clamoriniere, line 45: can we remove this `case` and use all the time `cache.MultiNamespacedCacheBuilder()`. do we have performance drawback?

vboulineau: Not sure about performance impact (not documented), but the `MultiNamespacedCacheBuilder()` is indeed a bit more complex, creating a map of cache / namespace, so I'd say we keep the recommended way.

clamoriniere (approved):

PR #166, opened 2020-11-04
[makefile] add helper functions 

nammn:
### What does this PR do?

Adding helper output to the makefile.

Please Note: 
- that make now defaults to help and one can choose a command from there. 
- I just added some comments and sections because I don't know all of them. Feel free to add some additionally later. 
```
⛵ gke_datadog-sandbox_europe-west1-b_nam-test (datadog) in datadog-operator on  nnguyen/add-make-help via 🐹 v1.15.3 took 4s 
❯ make
Usage: make <OPTIONS> ... <TARGETS>

Available targets are:


Usage:
  make <target>

Development
  all                        Build test
  build                      Builds manager + kubectl plugin
  fmt                        Run go fmt against code
  vet                        Run go vet against code

Deploy
  manager                    Build manager binary
  run                        Run against the configured Kubernetes cluster in ~/.kube/config
  install                    Install CRDs into a cluster
  uninstall                  Uninstall CRDs from a cluster
  deploy                     Deploy controller in the configured Kubernetes cluster in ~/.kube/config
  manifests                  Generate manifestcd s e.g. CRD, RBAC etc.
  generate                   Generate code
  docker-build               Build the docker image
  docker-push                Push the docker image

Test
  test                       Run tests
  controller-gen             Find or download controller-gen, download controller-gen if necessary
  bundle                     Generate bundle manifests and metadata, then validate generated files.
  bundle-build               Build the bundle image.

Datadog Custom part
  patch-crds                 Patch-crds
  lint                       Lint
  verify-license             Verify licenses
  tidy                       Run go tidy
  vendor                     Run go vendor
  help                       Show this help screen.


```
### Motivation

Outdated make docs

### Additional Notes

Anything else we should know when reviewing?

### Describe your test plan

Write there any instructions and details you may have to test your PR.

PR #167, opened 2020-11-04
[datadogfeature] adding orchestrator explorer as a feature

nammn:
### What does this PR do?

This PR does 2 changes.
- Adding a new `type` the `datadogfeature`, which contains features for the `agent` and `dca`
- Adding orchestrator explorer as a setting for the datadog operator
- Adding the option `processCollection` which defaults to `false` if this is set and `processAgent` is `enabled` live process collection happens as well. 

### Motivation

- Adding orchestrator_explorer to the operator

### Additional Notes

Do we want to add the `DDClusterName` to the process-agent as well? 

### Describe your test plan

- added integration test
- added unit test

manual tests:
- apply supplied `config/samples/datadog-agent-with-clusteragent-with-orchestrator.yaml` with correct API and APP Key settings
either:
- wait for information appear on orchestrator page (e.g. https://dddev.datadoghq.com/orchestration/overview?cols=name%2Cstatus%2Cready%2Crestarts%2Cage%2Clabels&end=1604667730878&paused=false&start=1604666830878&view=pod&tags=kube_cluster_name%3Anam-test) 
- look at DCA and process-agent logs:
process-agent
```
2020-11-06 12:42:32 UTC | PROCESS | INFO | (pkg/forwarder/transaction.go:293 in internalProcess) | Successfully posted payload to "https://orchestrator.datadoghq.com/api/v1/orchestrator", the agent will only log transaction success every 500 transactions
2020-11-06 12:42:32 UTC | PROCESS | INFO | (collector.go:375 in updateStatus) | real time interval updated to 2s
2020-11-06 12:42:42 UTC | PROCESS | INFO | (collector.go:157 in runCheck) | Finished container check #2 in 58.648578ms
2020-11-06 12:42:42 UTC | PROCESS | INFO | (collector.go:157 in runCheck) | Finished pod check #2 in 4.851825ms
```

DCA using a new image containing the updated status information
```
❯ k exec -it orchestrator-cluster-agent-cbd877866-vspfk -- agent status

  Cache Stats
  ===========
    Elements in the cache: 49
    Pods:
      Last Run: (Hits: 1 Miss: 0) | Total: (Hits: 3 Miss: 1)
    Deployments:
      Last Run: (Hits: 10 Miss: 0) | Total: (Hits: 30 Miss: 10)
    ReplicaSets:
      Last Run: (Hits: 23 Miss: 0) | Total: (Hits: 69 Miss: 23)
    Services:
      Last Run: (Hits: 7 Miss: 0) | Total: (Hits: 21 Miss: 7)
    Nodes:
      Last Run: (Hits: 3 Miss: 0) | Total: (Hits: 8 Miss: 4)
```


Makefile
```diff
@@ -30,69 +30,69 @@ else
 GOBIN=$(shell go env GOBIN)
```
nammn, line 30: note this will be removed once https://github.com/DataDog/datadog-operator/pull/166 gets merged

config/manager/kustomization.yaml
```diff
@@ -2,7 +2,7 @@ resources:
 - manager.yaml
 images:
 - name: controller
-  newName: datadog/operator
+  newName: nammn/operator
```
nammn, line 5: fix me

sarina-dd (approved):
👍 LGTM for docs

api/v1alpha1/datadogagent_types.go
```diff
@@ -358,6 +374,33 @@ type ProcessSpec struct {
...
+type OrchestratorExplorerConfig struct {
+	// Enable this to activate live kubernetes monitoring.
+	// ref: https://docs.datadoghq.com/infrastructure/livecontainers/#kubernetes-resources
+	//
+	// +optional
+	Enabled *bool `json:"enabled,omitempty"`
+
+	// +optional
+	// Deactivate this to stop the scrubbing of sensitive container data (passwords, tokens etc. ).
+	ContainerScrubbingEnabled *bool `json:"containerScrubbingEnabled,omitempty"`
```
clamoriniere, line 389: Do you think in the future you will introduce `Scrubbling` option? 
if yes maybe you can create a `ContainerScrubbling` struct

```golang
type ContainerScrubbling struct {
}
```

nammn: We were thinking about the possibility of adding multiple features to Container something like this

```go
type OrchestratorContainerOptions struct {
    ScrubbingEnabled *bool `json:"enabled,omitempty"`
    Feature2
    Feature3
}
```

But this is something we may add in the further future. Do you prefer to handle that case in this PR?

clamoriniere: I agree with other features enabling.

But for this specific feature, do you think we can have other sub-settings, or maybe `scrubbing` other resources like `configmap` `secret`?

in this case maybe a struct like
```yaml
    scrubbing:
       container: true
       configMap: true
```

could make more sens. WDYT?

api/v1alpha1/datadogagent_default.go
```diff
@@ -371,6 +403,12 @@ func DefaultDatadogAgent(ad *DatadogAgent) *DatadogAgent {
 		}
 	}
 
+	if defaultedAD.Spec.DatadogFeatures != nil {
+		if defaultedAD.Spec.DatadogFeatures.OrchestratorExplorer != nil {
+			defaultedAD.Spec.DatadogFeatures.OrchestratorExplorer = DefaultDatadogFeatureOrchestratorExplorer(defaultedAD.Spec.DatadogFeatures.OrchestratorExplorer)
+		}
+	}
```
clamoriniere, line 410:
```suggestion
	if defaultedAD.Spec.DatadogFeatures != nil && defaultedAD.Spec.DatadogFeatures.OrchestratorExplorer != nil {
		defaultedAD.Spec.DatadogFeatures.OrchestratorExplorer = DefaultDatadogFeatureOrchestratorExplorer(defaultedAD.Spec.DatadogFeatures.OrchestratorExplorer)
	}
```

api/v1alpha1/datadogagent_types.go
nammn: ah this makes sense. I don't think we had that granularity in mind while designing this feature. It certainly makes sense to invest more time into that in the future. I think we could leave it currently as it is and revisit if needed? The refactoring should not be too much work. 🤔 
But I don't have a big opinion either. 

Maybe @hkaj has an opinion on it?

PR #168, opened 2020-11-04
Fix scheduling of system-probe related checks in the core agent and missing permissions for network policies

vboulineau:
### What does this PR do?

To auto-activate OOM and TCP checks, the core agent needs to have access to `system-probe` socket and configuration file.
RBAC declaration for NetworkPolicies was missing (Kustomize)
Fixed DCA RBAC to collect events following introduction of `Spec.ClusterAgent.Config.CollectEvents`

### Motivation

### Additional Notes

### Describe your test plan

Activate system-probe OOM or TCP checks and verify they run in core agent.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -211,6 +211,12 @@ rules:
   - list
   - watch
 - apiGroups:
+  - networking.k8s.io
+  resources:
+  - networkpolicies
+  verbs:
+  - '*'
+- apiGroups:
   - policy
   resources:
   - poddisruptionbudgets
--- a/controllers/datadogagent/agent_test.go
+++ b/controllers/datadogagent/agent_test.go
@@ -825,6 +825,20 @@ func defaultPodSpec() corev1.PodSpec {
 }
 
 func defaultSystemProbePodSpec() corev1.PodSpec {
+	agentWithSystemProbeVolumeMounts := []corev1.VolumeMount{}
+	agentWithSystemProbeVolumeMounts = append(agentWithSystemProbeVolumeMounts, defaultMountVolume()...)
+	agentWithSystemProbeVolumeMounts = append(agentWithSystemProbeVolumeMounts, []corev1.VolumeMount{
+		{
+			Name:      "sysprobe-socket-dir",
+			ReadOnly:  true,
+			MountPath: "/opt/datadog-agent/run",
+		},
+		{
+			Name:      "system-probe-config",
+			MountPath: "/etc/datadog-agent/system-probe.yaml",
+			SubPath:   "system-probe.yaml",
+		},
+	}...)
 	return corev1.PodSpec{
 		ServiceAccountName: "foo-agent",
 		InitContainers: []corev1.Container{
@@ -850,7 +864,7 @@ func defaultSystemProbePodSpec() corev1.PodSpec {
 				Command:         []string{"bash", "-c"},
 				Args:            []string{"for script in $(find /etc/cont-init.d/ -type f -name '*.sh' | sort) ; do bash $script ; done"},
 				Env:             defaultEnvVars(),
-				VolumeMounts:    defaultMountVolume(),
+				VolumeMounts:    agentWithSystemProbeVolumeMounts,
 			},
 			{
 				Name:            "seccomp-setup",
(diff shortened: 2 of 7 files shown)


config/rbac/role.yaml
```diff
@@ -210,6 +210,12 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
```
clamoriniere, line 213: could you reference also adding networkpolicies RBAC in the PR description

vboulineau: Done

controllers/datadogagent/agent_test.go
```diff
@@ -825,6 +825,20 @@ func defaultPodSpec() corev1.PodSpec {
...
+		{
+			Name:      "sysprobe-socket-dir",
+			ReadOnly:  true,
+			MountPath: "/opt/datadog-agent/run",
+		},
+		{
+			Name:      "system-probe-config",
+			MountPath: "/etc/datadog-agent/system-probe.yaml",
+			SubPath:   "system-probe.yaml",
+		},
```
clamoriniere, line 840: does it mean that before we didn't mount the config map that contains the system-probe.yaml configuration?

vboulineau: Exactly

L3n41c (approved):
LGTM

controllers/datadogagent/agent_test.go
clamoriniere: https://github.com/DataDog/datadog-operator/blob/15e9acebe3d7340f1954c745dbb3047f3ba70a49/controllers/datadogagent/utils.go#L838

L3n41c: The thing is that, before the `oom_kill` and `tcp_queue_length` checks, `system-probe` talked only to the `process-agent`.
So, it was not needed to access `system-probe.yaml` from the core agent.

When `oom_kill` and `tcp_queue_length` checks were introduced, it initially worked without `system-probe.yaml` because the core agent was looking for the `system-probe` socket at its default path. (And as we are inside a container, there was no value to make this path configurable.)
But following some code change, I think that the core agent now needs to have access to `system-probe.yaml` to have the socket path.

clamoriniere: https://github.com/DataDog/datadog-operator/blob/15e9acebe3d7340f1954c745dbb3047f3ba70a49/controllers/datadogagent/utils.go#L838

vboulineau: BTW running `agent check tcp_queue_length ` still does not work, wonder if related?

clamoriniere: should we create an `emptyDir` to share the socket between the agents  that need it?

L3n41c, line 840: Can the `system-probe-config` volume be mounted as read-only ?

```suggestion
		{
			Name:      "system-probe-config",
			MountPath: "/etc/datadog-agent/system-probe.yaml",
			SubPath:   "system-probe.yaml",
			ReadOnly:  true,
		},
```

controllers/datadogagent/utils.go
```diff
@@ -1153,6 +1153,23 @@ func getVolumeMountsForAgent(spec *datadoghqv1alpha1.DatadogAgentSpec) []corev1.
...
+			{
+				Name:      datadoghqv1alpha1.SystemProbeSocketVolumeName,
+				MountPath: datadoghqv1alpha1.SystemProbeSocketVolumePath,
+				ReadOnly:  true,
+			},
+			{
+				Name:      datadoghqv1alpha1.SystemProbeConfigVolumeName,
+				MountPath: datadoghqv1alpha1.SystemProbeConfigVolumePath,
+				SubPath:   datadoghqv1alpha1.SystemProbeConfigVolumeSubPath,
+			},
```
L3n41c, line 1169: Can the config file be mounted read-only ?

```suggestion
			{
				Name:      datadoghqv1alpha1.SystemProbeConfigVolumeName,
				MountPath: datadoghqv1alpha1.SystemProbeConfigVolumePath,
				SubPath:   datadoghqv1alpha1.SystemProbeConfigVolumeSubPath,
				ReadOnly:  true,
			},
```

L3n41c (commented):
I’m just wondering if the `system-probe.yaml` file could be mounted read-only in the core agent container ?

controllers/datadogagent/agent_test.go
L3n41c: But the metrics are showing up in the app ?

vboulineau: Yes status is fine too

clamoriniere (approved):

controllers/datadogagent/agent_test.go
celenechang: i think you have to do something like `agent check tcp_queue_length -c /etc/datadog-agent/system-probe.yaml`

PR #169, opened 2020-11-05
Fix compliance CheckInterval set in DCA/SecAgent

vboulineau:
### What does this PR do?

Since move from string to `metav1.Duration`, CheckInterval is not working anymore. It's because `String()` of a `metav1.Duration` is not what the Agent/ClusterAgent expects.

### Motivation

### Additional Notes

### Describe your test plan

Activate compliance, set CheckInterval, checks the outputted check interval in the logs, should be the same value.


xornivore (approved):
👍

PR #181, opened 2020-11-24
Caroline/add command to plugin to list datadogmetrics

khewonc:
### What does this PR do?

Add `metrics get` command to list DatadogMetrics: `kubectl datadog metrics get`

### Motivation

Easily get a list of DatadogMetrics on older Kubernetes clusters that don't support `kubectl get <crd>`

### DIFF
--- a/cmd/kubectl-datadog/metrics/get/get.go
+++ b/cmd/kubectl-datadog/metrics/get/get.go
@@ -118,36 +118,19 @@ func (o *options) run() error {
 	table := newTable(o.Out)
 	for _, item := range ddList.Items {
 		metric := &metricData{
-			Namespace: item.Namespace,
-			Name:      item.Name,
+			Namespace:  item.Namespace,
+			Name:       item.Name,
+			Value:      item.Status.Value,
+			References: item.Status.AutoscalerReferences,
 		}
-		if item.Status.Conditions != nil {
-			for i := range item.Status.Conditions {
-				if item.Status.Conditions[i].Type == v1alpha1.DatadogMetricConditionTypeActive {
-					if string(item.Status.Conditions[i].Status) != "" {
-						metric.Active = string(item.Status.Conditions[i].Status)
-					}
-				}
-				if item.Status.Conditions[i].Type == v1alpha1.DatadogMetricConditionTypeValid {
-					if string(item.Status.Conditions[i].Status) != "" {
-						metric.Valid = string(item.Status.Conditions[i].Status)
-					}
-				}
-			}
-		}
-		if item.Status.Value != "" {
-			metric.Value = item.Status.Value
-		}
-		if item.Status.AutoscalerReferences != "" {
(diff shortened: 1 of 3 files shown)


cmd/kubectl-datadog/metrics/get/get.go
```diff
@@ -0,0 +1,176 @@
...
+		}
+		if item.Status.Value != "" {
+			data = append(data, item.Status.Value)
+		} else {
+			data = append(data, "")
+		}
+		if item.Status.AutoscalerReferences != "" {
+			data = append(data, item.Status.AutoscalerReferences)
+		} else {
+			data = append(data, "")
```
clamoriniere, line 139: for visibility, it could be better to create a struct like `metricsData` that contains the data fields like

```golang
type metricsData {
  Namespace string
  Name string
  Active string
  Value string
  // ...
}
```

and then create a function `func metricDataToData(metric *metricData) []data`

clamoriniere (commented):
Awesome PR 👍

I added some comment to improve the code readability and testing

cmd/kubectl-datadog/metrics/get/get.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	table := newTable(o.Out)
+	for _, item := range ddList.Items {
+		metric := &metricData{
+			Namespace: item.Namespace,
+			Name:      item.Name,
+		}
+		if item.Status.Conditions != nil {
+			for i := range item.Status.Conditions {
```
clamoriniere, line 125: nits: `range` support the fact that `item.Status.Conditions`, so no need to have `if item.Status.Conditions != nil {` before

```suggestion
		for i := range item.Status.Conditions {
```

cmd/kubectl-datadog/metrics/get/get.go
```diff
@@ -0,0 +1,187 @@
...
+				if item.Status.Conditions[i].Type == v1alpha1.DatadogMetricConditionTypeActive {
+					if string(item.Status.Conditions[i].Status) != "" {
+						metric.Active = string(item.Status.Conditions[i].Status)
+					}
+				}
```
clamoriniere, line 130: I think you can define a sub function that handle this logic, because I see almost the same code in 3 different places.

clamoriniere (commented):
two other small comments, before merging it.

khewonc:
I omitted some `if` statements checking for non-empty values as it seemed redundant because for
```
type metricData struct {
	Value  string
}
```
if `item.Status.Value=""`, checking and adding only non-empty values using
```
if item.Status.Value != "" {
    metricData.Value = item.Status.Value
}
```
would be equivalent to
```
metricData.Value = item.Status.Value
```
since `metricData.Value` would be set to zero values. Instead of a separate function for the `item.Status.Conditions`, I used a switch because it looked cleaner, but please let me know if these are not recommended.

clamoriniere (approved):

PR #182, opened 2020-11-30
[ddm] Add maxAge to DatadogMetric CRD

xornivore:
### What does this PR do?

Adds `MaxAge` field to the DatadogMetric CRD.

### Motivation

Ability to override the default in the agent on per resource basis.


clamoriniere (approved):
💯

PR #178, opened 2020-11-18
[datadogfeature] adding orchestrator explorer as a feature

nammn:
### What does this PR do?

Reopening https://github.com/DataDog/datadog-operator/pull/167 as it got merged by accident. 

----------
This PR does 2 changes.
- Adding a new `type` the `datadogfeature`, which contains features for the `agent` and `dca`
- Adding orchestrator explorer as a setting for the datadog operator
- Adding the option `processCollection` which defaults to `false` if this is set and `processAgent` is `enabled` live process collection happens as well. 

### Motivation

- Adding orchestrator_explorer to the operator

### Additional Notes

Do we want to add the `DDClusterName` to the process-agent as well? 

### Describe your test plan

- added integration test
- added unit test

manual tests:
- apply supplied `config/samples/datadog-agent-with-clusteragent-with-orchestrator.yaml` with correct API and APP Key settings
either:
- wait for information appear on orchestrator page (e.g. https://dddev.datadoghq.com/orchestration/overview?cols=name%2Cstatus%2Cready%2Crestarts%2Cage%2Clabels&end=1604667730878&paused=false&start=1604666830878&view=pod&tags=kube_cluster_name%3Anam-test) 
- look at DCA and process-agent logs:
process-agent
```
2020-11-06 12:42:32 UTC | PROCESS | INFO | (pkg/forwarder/transaction.go:293 in internalProcess) | Successfully posted payload to "https://orchestrator.datadoghq.com/api/v1/orchestrator", the agent will only log transaction success every 500 transactions
2020-11-06 12:42:32 UTC | PROCESS | INFO | (collector.go:375 in updateStatus) | real time interval updated to 2s
2020-11-06 12:42:42 UTC | PROCESS | INFO | (collector.go:157 in runCheck) | Finished container check #2 in 58.648578ms
2020-11-06 12:42:42 UTC | PROCESS | INFO | (collector.go:157 in runCheck) | Finished pod check #2 in 4.851825ms
```

DCA using a new image containing the updated status information
```
❯ k exec -it orchestrator-cluster-agent-cbd877866-vspfk -- agent status

  Cache Stats
  ===========
    Elements in the cache: 49
    Pods:
      Last Run: (Hits: 1 Miss: 0) | Total: (Hits: 3 Miss: 1)
    Deployments:
      Last Run: (Hits: 10 Miss: 0) | Total: (Hits: 30 Miss: 10)
    ReplicaSets:
[...]

PR #183, opened 2020-11-30
Introducing support for kube_state_metrics core.

CharlyF:
### What does this PR do?

Feature relies on:
- `DD_CLUSTER_AGENT_ENABLED` and `DD_CLUSTER_CHECKS_ENABLED` for the new check is instantiated as a Cluster Level Check (CLC), scheduled by the cluster agent (DCA).

Will force the configuration of:
- `DD_IGNORE_AUTOCONF` to remove the AutoDiscovery of the template of KSM V1.

Workflow:
- Enable the core check
- if no override of the config -> creates a ConfigMap with the check config as a CLC (2 CLCs by default)
- create a cluster role inspired from KSM upstream, bind it to the CLC service account.
- mount the config in the DCA.
- DCA distribute the check to the CLC(s)

### Additional Notes

The kube-state-metrics-core cluster role is not cleaned up, similarly to the external metrics provider associated cluster level roles. Because the controller is namespaced the owner can't be configured. I don't like this but we can tackle this in a different contribution.

### Describe your test plan

The 
```  
clusterAgent:
    image:
      name: "datadog/cluster-agent:1.10.0-rc6"
    config:
[...]
      kubeStateMetricsCoreEnabled: true
      clusterChecksEnabled: true
  clusterChecksRunner:
    image:
      name: "datadog/agent:7.24.0-rc.9"
```

- [ ] Create a `DatadogAgent` custom resource with the above.
- [ ] Check the creation of the cluster role:
```
{"level":"INFO","ts":"2020-11-30T21:30:27Z","logger":"controllers.DatadogAgent","msg":"createKubeStateMetricsClusterRole","datadogagent":"datadog-operator-system/datadog","clusterRole.name":"kube-state-metrics-core"}
```
- [ ] Check the creation of the configmap (default at first), should contain [the following](https://github.com/DataDog/datadog-operator/blob/5a9a863710a716db52a52dadddd55f76bf4bb51a/controllers/datadogagent/kubestatemetrics.go#L18-L43)
```
datadog-kube-state-metrics-core-config   1      80s
``` 
- [ ] Check it is mounted and interpreted by the DCA:
```
➜  ~ ke datadog-cluster-agent-77b54f6f6c-g9tjn agent clusterchecks
[...]
=== kubernetes_state_core check ===
Configuration provider: file
Configuration source: file:/etc/datadog-agent/conf.d/kubernetes_state_core.yaml
Instance ID: kubernetes_state_core:d4eb2e03da98bb72
collectors:
- pods
- replicationcontrollers
[...]
Instance ID: kubernetes_state_core:83ea7f53d9f648ac
collectors:
- deployments
- configmaps
[...]
```

- [ ] Validate the CLC worker is starting the informers and the checks:
```
I1130 21:30:54.548408     399 builder.go:157] Active resources: cronjobs,jobs,nodes,pods,replicasets,replicationcontrollers,statefulsets
I1130 21:30:54.551203     399 builder.go:157] Active resources: configmaps,daemonsets,deployments,endpoints,horizontalpodautoscalers,limitranges,namespaces,persistentvolumeclaims,persistentvolumes,resourcequotas,secrets,services
2020-11-30 21:30:54 UTC | CORE | INFO | (pkg/collector/scheduler/scheduler.go:85 in Enter) | Scheduling check kubernetes_state_core with an interval of 15s
2020-11-30 21:30:54 UTC | CORE | INFO | (pkg/collector/scheduler/scheduler.go:85 in Enter) | Scheduling check kubernetes_state_core with an interval of 15s
```
(The above is ~ out of scope)

- [ ] Use a custom CM:

```
{"level":"INFO","ts":"2020-11-30T23:14:00Z","logger":"controllers.DatadogAgent","msg":"Using custom configuration for Kubernetes State Metrics Core","datadogagent":"datadog-operator-system/datadog","ConfigMap name":"test-ksm"}
```

- [ ] Use the configData field.

- [ ] Ensure that the documentation is clear
- [ ] Delete the `DatadogAgent` and ensure that all is removed (except the custom CM if used).

TODO 
- [x] more test coverage
- [x] doc


controllers/datadogagent/kubestatemetrics.go
```diff
@@ -0,0 +1,194 @@
...
+func buildKubeStateMetricsCoreRBAC(dda *datadoghqv1alpha1.DatadogAgent, name, version string) *rbacv1.ClusterRole {
+	clusterRole := &rbacv1.ClusterRole{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels:      getDefaultLabels(dda, name, version),
+			Annotations: getDefaultAnnotations(dda),
+			Name:        name,
+		},
+	}
+
+	rbacRules := []rbacv1.PolicyRule{
```
xornivore, line 94: Is it worth adding something to the `DatadogAgent` CRD to customize this set of rules?

controllers/datadogagent/utils.go
```diff
@@ -650,6 +650,13 @@ func getEnvVarsForAgent(dda *datadoghqv1alpha1.DatadogAgent) ([]corev1.EnvVar, e
 					Value: datadoghqv1alpha1.ClusterAndEndpointsConfigPoviders,
 				})
 			} else {
+				// Remove ksm v1 conf if the cluster checks are enabled and the ksm core is enabled
+				if isKSMCoreEnabled(spec.ClusterAgent) {
+					envVars = append(envVars, corev1.EnvVar{
```
xornivore, line 655: If the user provides their own ignore list - are we going to replace it with just one containing `kubernetes_state`?

controllers/datadogagent/kubestatemetrics.go
CharlyF: That's what we discussed in standup last week, for now we keep it simple, but if someone asks that we make it customizable we can add the logic.

controllers/datadogagent/utils.go
CharlyF: hum, my comment did not make it, it seems... I thought appending to the envvar list was enough but you are right, I need to append to the value of the var

xornivore (approved):
👍

api/v1alpha1/datadogagent_types.go
```diff
@@ -760,7 +760,16 @@ type ClusterAgentConfig struct {
...
+	// Enable this to start the Kubernetes State Metrics Core check.
+	// TODO add reference to the documentation.
+	// +optional
+	KubeStateMetricsCoreEnabled *bool `json:"kubeStateMetricsCoreEnabled,omitempty"`
+
+	// To override the configuration for the default Kubernetes State Metrics Core check.
+	// Must point to a ConfigMap containing a valid cluster check configuration.
+	KubeStateMetricsCoreConf *string `json:"kubeStateMetricsCoreConf,omitempty"`
+
+	// Enable this to start event collection from the kubernetes API
```
clamoriniere, line 772: Can we create a struct to agreggate the `KubeStateMetricsCore` Integration

```golang
type KubeStateMetricsCore struct {
   Enabled *bool
   Conf *conf
}
```
It will be more future proof. 

Since https://github.com/DataDog/datadog-operator/pull/178 was merge we now have the `DatadogAgent.Spec.Features` section, I think it cool be good to move this configuration in the `features` section

api/v1alpha1/datadogagent_types.go
```diff
@@ -760,7 +760,16 @@ type ClusterAgentConfig struct {
 	// Autodiscovery via Kube Service annotations is automatically enabled
 	ClusterChecksEnabled *bool `json:"clusterChecksEnabled,omitempty"`
 
-	// Enables this to start event collection from the Kubernetes API
+	// Enable this to start the Kubernetes State Metrics Core check.
+	// TODO add reference to the documentation.
```
clamoriniere, line 764: could you address this TODO?

api/v1alpha1/datadogagent_types.go
clamoriniere: Also about the name, I'm wondering if we can name it: `KubeStateMetrics`

controllers/datadogagent/kubestatemetrics.go
```diff
@@ -0,0 +1,194 @@
...
+      - endpoints
+      - daemonsets
+      - horizontalpodautoscalers
+      - limitranges
+      - resourcequotas
+      - secrets
+      - namespaces
+      - persistentvolumeclaims
+      - persistentvolumes
+    telemetry: true
```
clamoriniere, line 44: should we have once instance by collector, as we discussed?
Or maybe for now, only one instance?

clamoriniere (commented):
Awesome work here 👍 
Few comment about the integration in the `DatadogAgent.Spec`

api/v1alpha1/datadogagent_types.go
CharlyF: We have been introducing the new check as a core check, I think it is important to distinguish that it is not the V1.
I tried changing out of curiosity, and it gets confusing with what is KSM, the conf of the check and the check runtime itself.

CharlyF: I split up in 2 commits, the last one refactors the code to use the feature. Let me know which one you prefer (can revert the last commit if necessary, I actually like using the `features`)

controllers/datadogagent/kubestatemetrics.go
CharlyF: moved to 1 instance

api/v1alpha1/datadogagent_types.go
```diff
@@ -375,6 +377,20 @@ type ProcessSpec struct {
...
+// +k8s:openapi-gen=true
+type KubeStateMetricsCore struct {
+	// Enable this to start the Kubernetes State Metrics Core check.
+	// Refer to https://github.com/DataDog/datadog-operator/blob/master/docs/kubernetes_state_metrics.md
+	// +optional
+	Enabled *bool `json:"enabled,omitempty"`
+
+	// To override the configuration for the default Kubernetes State Metrics Core check.
+	// Must point to a ConfigMap containing a valid cluster check configuration.
+	Conf *string `json:"conf,omitempty"`
```
clamoriniere, line 391: to configure the configmap, I think we can reuse some go struct that we already use in this project like in [CustomConfigSpec](https://github.com/DataDog/datadog-operator/blob/cba7872e332dfbf2db8d2e2397bd20d4e25815f9/api/v1alpha1/datadogagent_types.go#L561) we have the [ConfigFileConfigMapSpec](https://github.com/DataDog/datadog-operator/blob/cba7872e332dfbf2db8d2e2397bd20d4e25815f9/api/v1alpha1/datadogagent_types.go#L550) use to provide the ConfigMap information.

Maybe we need to rename the `ConfigFileConfigMapSpec` in a more generic way.

controllers/datadogagent/common_configmap.go
```diff
@@ -68,6 +68,7 @@ func (r *Reconciler) updateIfNeededConfigMap(dda *datadoghqv1alpha1.DatadogAgent
 	// Copy possibly changed fields
 	updateCM := oldConfigMap.DeepCopy()
 	updateCM.Data = newConfigMap.Data
+	// TODO if new CM does not have labels it panics.
```
clamoriniere, line 71: what is this todo?

docs/kubernetes_state_metrics.md
```diff
@@ -0,0 +1,132 @@
+# Monitoring your cluster with Kubernetes State Metrics
+
+## Introduction
+You can easily monitor your cluster using [Kubernetes State Metrics][1].
```
clamoriniere, line 4:
```suggestion

You can easily monitor your cluster using [Kubernetes State Metrics][1].
```

docs/kubernetes_state_metrics.md
```diff
@@ -0,0 +1,132 @@
...
+
+The Datadog Operator has a feature that allows users to configure the new Kubernetes State Metrics check (v2.0.0) as a Cluster Level Check.
+See the Further Reading section for more details on the v2.0.0 of the check.
+You will need to have the [Datadog Cluster Agent][2] as well as the Cluster Level Check features enabled.
+
+## Configuration 
+To enable this feature, you will need to use the option `kubeStateMetricsCore.enabled: true`, the DatadogAgent spec should look like this:
```
clamoriniere, line 11:
```suggestion

To enable this feature, you will need to use the option `kubeStateMetricsCore.enabled: true`, the DatadogAgent spec should look like this:
```

docs/kubernetes_state_metrics.md
```diff
@@ -0,0 +1,132 @@
...
+      enabled: true
+      conf: custom-kubernetes-state-core-check
+```
+
+## Further Reading
+
+The v2 of the Kubernetes State Metrics check is embedded as a "core check" in the Datadog Agent.
+When an agent (Cluster Check Runner) receives the instruction to schedule this check, the configured Kube State Metrics collectors will be started as separate routines in the container.
+
+This means that the agent does not need to monitor independent instances of Kube State Metrics anymore.
```
clamoriniere, line 126: maybe you can add again a link to the kube-state-metrics here

```suggestion
This means that the agent does not need to monitor independent instances of [Kube State Metrics][1] anymore. 
```

api/v1alpha1/datadogagent_types.go
clamoriniere: The doc is great 👍 , just few nits.

For the name, `KubeStateMetrics` or `KubeStateMetricsCore` is not a big deal. I just wanted to have a same that talk to the user in term of a feature. 

For the configuration in the CRD, I added another comment, let me know what do you think.

controllers/datadogagent/common_configmap.go
CharlyF: I know :( I will add a card to investigate.

api/v1alpha1/datadogagent_default.go
```diff
@@ -99,7 +102,10 @@ func IsDefaultedDatadogAgent(ad *DatadogAgent) bool {
 	}
 
 	if ad.Spec.Features != nil {
-		if !IsDefaultedOrchestratorExplorer(ad.Spec.Features.OrchestratorExplorer) {
+		if ad.Spec.Features.OrchestratorExplorer != nil && !IsDefaultedOrchestratorExplorer(ad.Spec.Features.OrchestratorExplorer) {
+			return false
+		}
+		if ad.Spec.Features.KubeStateMetricsCore != nil && !IsDefaultedKubeStateMetricsCore(ad.Spec.Features.KubeStateMetricsCore) {
```
clamoriniere, line 108: I think checking if the value is `nil` can be part of the function  `IsDefaultedKubeStateMetricsCore()`
```suggestion
		if ad.Spec.Features.KubeStateMetricsCore != nil && !IsDefaultedKubeStateMetricsCore(ad.Spec.Features.KubeStateMetricsCore) {
```

clamoriniere: in fact the function is already doing it :)

api/v1alpha1/datadogagent_default.go
```diff
@@ -407,8 +424,13 @@ func DefaultDatadogAgent(ad *DatadogAgent) *DatadogAgent {
 		}
 	}
 
-	if defaultedAD.Spec.Features != nil && defaultedAD.Spec.Features.OrchestratorExplorer != nil {
-		defaultedAD.Spec.Features.OrchestratorExplorer = DefaultDatadogFeatureOrchestratorExplorer(defaultedAD.Spec.Features.OrchestratorExplorer)
+	if defaultedAD.Spec.Features != nil {
+		if defaultedAD.Spec.Features.OrchestratorExplorer != nil {
+			defaultedAD.Spec.Features.OrchestratorExplorer = DefaultDatadogFeatureOrchestratorExplorer(defaultedAD.Spec.Features.OrchestratorExplorer)
+		}
```
clamoriniere, line 430: same here the if is maybe not needed because you already handle the case in the function

api/v1alpha1/datadogagent_default.go
```diff
@@ -407,8 +424,13 @@ func DefaultDatadogAgent(ad *DatadogAgent) *DatadogAgent {
...
+		if defaultedAD.Spec.Features.KubeStateMetricsCore != nil {
+			defaultedAD.Spec.Features.KubeStateMetricsCore = DefaultDatadogFeatureKubeStateMetricsCore(defaultedAD.Spec.Features.KubeStateMetricsCore)
+		}
```
clamoriniere, line 433: Same comment

api/v1alpha1/datadogagent_validation.go
```diff
@@ -39,6 +39,13 @@ func IsValidDatadogAgent(spec *DatadogAgentSpec) error {
 		}
 	}
 
+	if spec.Features != nil && spec.Features.KubeStateMetricsCore != nil {
+		if spec.Features.KubeStateMetricsCore.Conf != nil {
+			if err = IsValidCustomConfigSpec(spec.Features.KubeStateMetricsCore.Conf); err != nil {
+				errs = append(errs, fmt.Errorf("invalid spec.features.kubeStateMetricsCore.conf, err: %v", err))
+			}
+		}
+	}
```
clamoriniere, line 48:
```suggestion
	if spec.Features != nil && spec.Features.KubeStateMetricsCore != nil && spec.Features.KubeStateMetricsCore.Conf != nil {
		if err = IsValidCustomConfigSpec(spec.Features.KubeStateMetricsCore.Conf); err != nil {
			errs = append(errs, fmt.Errorf("invalid spec.features.kubeStateMetricsCore.conf, err: %v", err))
		}
	}
```

bundle/manifests/datadoghq.com_datadogagents.yaml
```diff
@@ -2489,6 +2489,10 @@ spec:
                           description: 'Enable informer and controller of the watermark pod autoscaler NOTE: The WatermarkPodAutoscaler controller needs to be installed see https://github.com/DataDog/watermarkpodautoscaler for more details.'
                           type: boolean
                       type: object
+                    kubeStateMetricsCoreEnabled:
+                      description: Enable this to start the Kubernetes State Metrics
+                        Core check. TODO add reference to the documentation.
+                      type: boolean
```
clamoriniere, line 2495: I think this part should be present anymore,  since `enable` is now a field in the `kubeStateMetricsCore` section

api/v1alpha1/datadogagent_default.go
```diff
@@ -682,6 +704,20 @@ func DefaultDatadogFeatureOrchestratorExplorer(explorerConfig *OrchestratorExplo
...
+	if ksmCore == nil {
+		ksmCore = &KubeStateMetricsCore{}
+	}
+
+	if ksmCore.Enabled == nil {
+		ksmCore.Enabled = NewBoolPointer(defaultKubeStateMetricsCoreEnabled)
+	}
+
+	return ksmCore
+}
```
clamoriniere, line 719:
```suggestion
func DefaultDatadogFeatureKubeStateMetricsCore(ksmCore *KubeStateMetricsCore) *KubeStateMetricsCore {
	if ksmCore == nil {
		return ksmCore
	}

	if ksmCore.Enabled == nil {
		ksmCore.Enabled = NewBoolPointer(defaultKubeStateMetricsCoreEnabled)
	}

	return ksmCore
}
```

api/v1alpha1/datadogagent_default.go
CharlyF: So for this one, I actually want to keep it that way. 
If the `spec.Features` is != nil, and the orchestrator explorer is not activated, I don't think we should mutate the field.

api/v1alpha1/datadogagent_default.go
CharlyF: true, but in this case I only want to check if it's defaulted if the main field is not nil.
with the above:
```
features:
    kubeStateMetricsCore:
         enabled: true
```
will not return `false`. Whereas if I don't split it, I would end up needing to mutate the orchestrator explorer field. Which I don't think we should do.

api/v1alpha1/datadogagent_validation.go
CharlyF: so `spec.Features.KubeStateMetricsCore.Conf` can be nil and actually we rely on the fact that if it is not set it is nil to create the default config.

api/v1alpha1/datadogagent_default.go
CharlyF: this is not necessary anymore

bundle/manifests/datadoghq.com_datadogagents.yaml
CharlyF: ah, odd that it still there

api/v1alpha1/datadogagent_validation.go
CharlyF: edit: after discussing, we will instantiate the features field if nil and Default all features.

bundle/manifests/datadoghq.com_datadogagents.yaml
CharlyF: updated the bundle... Seems like it was missing a few things from previous changes too

clamoriniere (commented):
Nice work 💯

clamoriniere (approved):

PR #186, opened 2020-12-07
Fix process-agent and trace-agent custom-config configuration

clamoriniere:
### What does this PR do?

Add missing `custom-config` volumeMount in `process-agent` and `trace-agent`. 
 
### Motivation

Fix `process-agent` and `trace-agent` configuration when the user provide a `customConfig`.

### Additional Notes

N/A

### Describe your test plan

Create a new DatadogAgent deployment. In the agent.spec add a `customConfig` configMap. Check
If this configMap is properly mounted in the `process-agent` and the `trace-agent`.

### DIFF
--- a/controllers/datadogagent/agent_test.go
+++ b/controllers/datadogagent/agent_test.go
@@ -25,6 +25,7 @@ import (
 )
 
 const testDdaName = "foo"
+const agentConfigFile = "/etc/datadog-agent/datadog.yaml"
 
 func apiKeyValue() *corev1.EnvVarSource {
 	return &corev1.EnvVarSource{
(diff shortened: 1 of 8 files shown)


controllers/datadogagent/agent_test.go
```diff
@@ -824,6 +872,29 @@ func defaultPodSpec() corev1.PodSpec {
...
+		Command:         []string{"trace-agent", "--config=/etc/datadog-agent/datadog.yaml"},
+		Resources:       corev1.ResourceRequirements{},
+		Ports:           []corev1.ContainerPort{{Name: "traceport", ContainerPort: 8126, Protocol: "TCP"}},
+		Env:             defaultAPMContainerEnvVars(),
+		VolumeMounts: []corev1.VolumeMount{
+			{Name: "config", MountPath: "/etc/datadog-agent"},
+			{
+				Name:      "custom-datadog-yaml",
+				ReadOnly:  true,
+				MountPath: "/etc/datadog-agent/datadog.yaml",
```
xornivore, line 889: I feel like we should have these as consts, it's repeated in another form above.

xornivore (commented):
Looks good (small comment about const), thoughts about adding more tests? 😄

controllers/datadogagent/agent_test.go
clamoriniere: the const value exist in `api/v1alpha1`. but the idea here was to no use it in the test

clamoriniere: I can add another const value in the test file.

xornivore (approved):

PR #190, opened 2020-12-08
Fix issues with Secret management

### LINKED ISSUE #96: If appKeyExistingSecret is specified, then the token secret fails to create
What is the info page??

**Describe what happened:**

I have setup my app key in a secret that I have created previously, called `secretapp`. I set up my cluster token in plain text under the `token` field. This is a snippet of my agent spec:

```
spec:
  credentials:
    apiKeyExistingSecret: secretapi
    appKeyExistingSecret: secretapp
    token: "<ThirtyX2XcharactersXlongXtoken>"
```

The cluster-agent fails to run, because it expects that the `secretapp` key to have a token key. 

```
  Warning  Failed     2s (x3 over 16s)  kubelet, kind-worker2  Error: couldn't find key token in Secret datadog/secretapp
```

If the `token` key is not there, edit the secret and add the `token` key using the string passed in the `token` spec.

Create an agent and cluster agent with the appkey passed as secret and try to pass a token:

```
spec:
  credentials:
    apiKeyExistingSecret: secretapi
    appKeyExistingSecret: secretapp
    token: "<ThirtyX2XcharactersXlongXtoken>"
```

### LINKED ISSUE #191: Mention required token for cluster agent
I just installed the datadog-operator into a cluster and for the most part the process was pretty straightforward, nice work on this!

However I did run into an issue getting the cluster-agent working: I didn't set a 'secret token' anywhere for the cluster-agent to use, which caused a `CreateContainerConfigError` error to throw from the cluster-agent complaining that `secret "datadog" not found`. I didn't realize it at first after reading through the installation/getting started docs, but a token is required and can be set in a secret called 'datadog' ([as mentioned here](https://docs.datadoghq.com/agent/cluster_agent/setup/?tab=secret#step-2---secure-cluster-agent-to-agent-communication)), or it can be set in the in `spec.credentials.token` field of the `DatadogAgent` config. Currently, I see that it is shown in [this example with the cluster agent](https://github.com/DataDog/datadog-operator/blob/master/examples/datadog-agent-with-clusteragent.yaml#L9), but not in [the 'all' example](https://github.com/DataDog/datadog-operator/blob/master/examples/datadog-agent-all.yaml). I feel like having to set this token should be called out more explicitly in the docs.

Separately, I did think it was kind of odd that the `api_key` and the `app_key` can be configured through the `DatadogAgent` to get the value from a secret (instead of setting the plain text value in the config), but the `token` [can only be set plainly in the config](https://github.com/DataDog/datadog-operator/blob/v0.4.0/config/crd/bases/v1/datadoghq.com_datadogagents.yaml#L7765) and can't be fetched from a secret. Was that intentional? That functionality might be nice too, so I could potentially just have one secret with my api/app keys and token for the cluster-agent.

L3n41c:
### What does this PR do?

Fix the following issues with Secret management:
* When a secret was provided for the API key but the APP key was provided in clear, the operator wrongly tried to create a secret for the APP key with the same name as the user-provided one for the API key.
* When a secret was provided for the API key and for the APP key, but the DCA token was not provided, the DCA pod manifest were referencing a secret for the token that was not created by the operator. The DCA pod were staying in `CreateContainerConfigError` state forever with a event saying `Error: secret "datadog" not found`.

### Motivation

Make things work!

### Additional Notes

Fixes #96
Fixes #191

### Describe your test plan

Try to create a `datadogagent` CR with the following `credentials` part:
```
spec:
  credentials:
    appKey: PUTYOUROWNHERE
    apiSecret:
      keyName: api_key
      secretName: my-datadog-secrets
```
The operator must spawn the agents properly instead of doing nothing and logging the following error:
```
{"level":"ERROR","ts":"2020-12-08T14:09:47Z","logger":"controller","msg":"Reconciler error","reconcilerGroup":"datadoghq.com","reconcilerKind":"DatadogAgent","controller":"datadogagent","name":"datadog","namespace":"datadog-operator-system","error":"secrets \"my-datadog-secrets\" already exists"}
```

Try to create a `datadogagent` CR with the following `credentials` part:
```
spec:
  credentials:
    appSecret:
      keyName: app_key
      secretName: my-datadog-secrets
    apiSecret:
      keyName: api_key
      secretName: my-datadog-secrets
```
The operator must spawn the agents properly instead of having DCA pods stuck in `CreateContainerConfigError` state with the following error: 
```
Events:
  Type     Reason          Age                  From               Message
  ----     ------          ----                 ----               -------
  Warning  Failed          109s (x9 over 2m8s)  kubelet            Error: secret "datadog" not found
```

### DIFF
--- a/controllers/datadogagent/secret.go
+++ b/controllers/datadogagent/secret.go
@@ -21,7 +21,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	datadoghqv1alpha1 "github.com/DataDog/datadog-operator/api/v1alpha1"
-	"github.com/DataDog/datadog-operator/pkg/controller/utils"
 	"github.com/DataDog/datadog-operator/pkg/controller/utils/condition"
 	"github.com/DataDog/datadog-operator/pkg/controller/utils/datadog"
 )
@@ -33,17 +32,17 @@ func (r *Reconciler) manageAgentSecret(logger logr.Logger, dda *datadoghqv1alpha
 	}
 	now := metav1.NewTime(time.Now())
 	// checks token secret
-	secretName, _ := utils.GetAppKeySecret(dda)
 	secret := &corev1.Secret{}
-	err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: dda.Namespace, Name: secretName}, secret)
+	err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: dda.Namespace, Name: dda.Name}, secret)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			if (dda.Spec.Credentials.AppKeyExistingSecret == "" && dda.Spec.Credentials.APPSecret == nil) ||
-				(dda.Spec.Credentials.APIKeyExistingSecret == "" && dda.Spec.Credentials.APISecret == nil) {
+				(dda.Spec.Credentials.APIKeyExistingSecret == "" && dda.Spec.Credentials.APISecret == nil) ||
+				dda.Spec.ClusterAgent != nil {
 				return r.createAgentSecret(logger, dda)
 			}
 			// return error since the secret didn't exist and we are not responsible to create it.
-			err = fmt.Errorf("secret %s didn't exist", secretName)
+			err = fmt.Errorf("secret %s didn't exist", dda.Name)
 			condition.UpdateDatadogAgentStatusConditions(newStatus, now, datadoghqv1alpha1.ConditionTypeSecretError, corev1.ConditionTrue, fmt.Sprintf("%v", err), false)
 		}
 		return reconcile.Result{}, err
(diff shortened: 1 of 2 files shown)


controllers/datadogagent/secret.go
```diff
@@ -33,17 +32,17 @@ func (r *Reconciler) manageAgentSecret(logger logr.Logger, dda *datadoghqv1alpha
 	}
 	now := metav1.NewTime(time.Now())
 	// checks token secret
-	secretName, _ := utils.GetAppKeySecret(dda)
```
clamoriniere, line 36: the secret name is always equal to the `DatadogAgent.Name` ?

clamoriniere: should we remove this function if it isn't used anymore

clamoriniere (commented):
Could you try to add a unit-test that validate this fix?

controllers/datadogagent/secret.go
L3n41c: In fact, `utils.GetAppKeySecret(dda)` returns the secret that holds the APP key.
So, it can be:
* either the user-supplied secret if the user has setup `credentials.appSecret` in its `datadogagent` CR,
* or `DatadogAgent.Name` if the APP key is in the operator-managed secret if the user has setup `credentials.appKey` in its `datadogagent` CR.

So, `func GetAppKeySecret(…)` is used to build the agent manifest to setup the `DD_APP_KEY` environment variable.
But in this context, here, we need the name of the operator-managed secret and we don’t care whether the APP key as been provided in clear or in a user-supplied secret.

clamoriniere (approved):

PR #197, opened 2020-12-12
allow dd credentials to be set in operator

celenechang:
### What does this PR do?

Currently, Datadog API/app keys need to be set in the DatadogAgent custom resource. This is so that if one were to have multiple agent deployments, unique keys could be used as needed. However, if one desires to use a single API/app key across operator resources, this setup can become cumbersome. This PR allows for the credentials to be set in the operator as environmental variables which can be used as a "fallback" if one chooses to omit credentials from DatadogAgent.

### Motivation

Ease of configuration in forthcoming DatadogMonitor resource.

### Describe your test plan

- Set up kind cluster
- Build and deploy operator and DatadogAgent CRDs as described in the docs
- Deploy a [basic DatadogAgent](https://github.com/DataDog/datadog-operator/blob/master/examples/datadog-agent.yaml) resource with the credentials defined; confirm that there are no errors
- Add `DD_API_KEY` and `DD_APP_KEY` env vars to the DatadogOperator deployment
- Deploy a second DatadogAgent resource without credentials defined, and verify there are no errors. Can check the output of `agent status` to match the API key against the one set in the operator.


vboulineau (approved):

clamoriniere (approved):
/lgtm

PR #200, opened 2020-12-30
Add DatadogMonitor controller

celenechang:
### What does this PR do?

Adds DatadogMonitor controller, with support for the Metric Alert type creation in Datadog.

### Motivation

### Additional Notes

Note: tests still need to be added to this PR.

### Describe your test plan

- Start a `kind` cluster
- Modify `config/manager/manager.yaml` to use `ImagePullPolicy: Never` and add `DD_API_KEY` and `DD_APP_KEY` env vars
- `make build`
- `make IMG=test/operator:test docker-build`
- `kind load docker-image test/operator:test`
- `make IMG=test/operator:test deploy`
- `k apply -f examples/datadog-monitor.yaml`
Check that the monitor in the corresponding Datadog account has been created, and that after a few minutes the monitor state is reflected in `k get datadogmonitor` and `k describe datadogmonitor` (if it is triggering).


controllers/datadogmonitor/datadogmonitor_controller.go
```diff
@@ -0,0 +1,385 @@
...
+
+	return nil
+}
+
+func (r *Reconciler) updateStatusIfNeeded(logger logr.Logger, datadogMonitor *datadoghqv1alpha1.DatadogMonitor, now metav1.Time, status *datadoghqv1alpha1.DatadogMonitorStatus, currentErr error, result ctrl.Result) (ctrl.Result, error) {
+	// Update Error and Active conditions
+	setErrorActiveConditions(status, now, currentErr)
+
+	if !apiequality.Semantic.DeepEqual(&datadogMonitor.Status, status) {
+		datadogMonitorCopy := datadogMonitor.DeepCopy()
```
vboulineau, line 271: I think a datadogMonitor with `metadata` and `status` is enough as you update a sub-resource)

controllers/datadogmonitor/controller.go
```diff
@@ -0,0 +1,325 @@
...
+		return r.updateStatusIfNeeded(logger, instance, now, newStatus, err, result)
+
+	}
+	statusSpecHash := instance.Status.CurrentHash
+
+	// Create or update monitor, or check monitor state. Fall through this block (without returning)
+	// if the result should be requeued with the default period
+	if instance.Status.ID == 0 {
+		// If the monitor ID is 0, then it doesn't exist yet in Datadog. Create the monitor (only metric alerts)
+		if string(instance.Spec.Type) != string(datadogapiclientv1.MONITORTYPE_METRIC_ALERT) {
```
clamoriniere, line 116: we could use a switch case here already

```golang
switch instance.Spec.Type {
case datadogapiclientv1.MONITORTYPE_METRIC_ALERT:
  // ....
case default:
  // unknown type
}
```

controllers/datadogmonitor/controller.go
```diff
@@ -0,0 +1,325 @@
...
+		}
+		// Make sure required tags are present
+		if result, err = r.checkRequiredTags(logger, instance); err != nil || result.Requeue {
+			return r.updateStatusIfNeeded(logger, instance, now, newStatus, err, result)
+		}
+
+		err = r.create(logger, instance, newStatus, now)
+		if err != nil {
+			logger.Error(err, "Error creating monitor")
+		}
```
clamoriniere, line 129: nit 
```suggestion
		if err = r.create(logger, instance, newStatus, now); err != nil {
			logger.Error(err, "Error creating monitor")
		}
```

controllers/datadogmonitor/controller.go
```diff
@@ -0,0 +1,325 @@
...
+		if instanceSpecHash != statusSpecHash {
+			// Make sure required tags are present
+			if result, err = r.checkRequiredTags(logger, instance); err != nil || result.Requeue {
+				return r.updateStatusIfNeeded(logger, instance, now, newStatus, err, result)
+			}
+			// Update action
+			err = r.update(logger, instance, newStatus, now)
+			if err != nil {
+				logger.Error(err, "Error updating monitor")
+			}
```
clamoriniere, line 142:
```suggestion
			if err = r.update(logger, instance, newStatus, now); err != nil {
				logger.Error(err, "Error updating monitor")
			}
```

controllers/datadogmonitor/controller.go
```diff
@@ -0,0 +1,325 @@
...
+			newStatus.CurrentHash = instanceSpecHash
+		} else {
+			// Spec has not changed, just check if monitor state has changed (alert, warn, OK, etc.)
+			err = r.get(logger, instance, newStatus)
+			if err != nil {
+				logger.Error(err, "Error getting monitor")
+			}
```
clamoriniere, line 149:
```suggestion
			if err = r.get(logger, instance, newStatus); err != nil {
				logger.Error(err, "Error getting monitor")
			}
```

controllers/datadogmonitor/controller.go
```diff
@@ -0,0 +1,325 @@
...
+	}
+
+	// Update the status via the Kubernetes API
+	return r.updateStatusIfNeeded(logger, instance, now, newStatus, err, result)
+}
+
+func (r *Reconciler) create(logger logr.Logger, datadogMonitor *datadoghqv1alpha1.DatadogMonitor, status *datadoghqv1alpha1.DatadogMonitorStatus, now metav1.Time) error {
+	logger.Info(fmt.Sprint(datadogMonitor.Spec.Query))
+
+	logger.Info(fmt.Sprint(datadogMonitor.Status.ID))
```
clamoriniere, line 165: maybe you can merge it in on log ligne

```suggestion
	logger.Info("monitor creation", "spec.query", datadogMonitor.Spec.Query, "status.id",  datadogMonitor.Status.ID)
```

controllers/datadogmonitor/controller.go
```diff
@@ -0,0 +1,325 @@
...
+	// Validate monitor in Datadog
+	err := validateMonitor(r.datadogAuth, r.datadogClient, datadogMonitor)
+	if err != nil {
+		return err
+	}
```
clamoriniere, line 170: to scope the error to the if condition
```suggestion
	if err := validateMonitor(r.datadogAuth, r.datadogClient, datadogMonitor); err != nil {
		return err
	}
```

controllers/datadogmonitor/controller.go
```diff
@@ -0,0 +1,325 @@
...
+
+	return nil
+}
+
+func (r *Reconciler) update(logger logr.Logger, datadogMonitor *datadoghqv1alpha1.DatadogMonitor, status *datadoghqv1alpha1.DatadogMonitorStatus, now metav1.Time) error {
+	// Validate monitor in Datadog
+	err := validateMonitor(r.datadogAuth, r.datadogClient, datadogMonitor)
+	if err != nil {
+		return err
+	}
```
clamoriniere, line 200:
```suggestion
	if err := validateMonitor(r.datadogAuth, r.datadogClient, datadogMonitor); err != nil {
		return err
	}
```

controllers/datadogmonitor/controller.go
```diff
@@ -0,0 +1,325 @@
...
+
+	// Update monitor in Datadog
+	_, err = updateMonitor(r.datadogAuth, r.datadogClient, datadogMonitor)
+	if err != nil {
+		return err
+	}
```
clamoriniere, line 206:
```suggestion
	
	if _, err := updateMonitor(r.datadogAuth, r.datadogClient, datadogMonitor); err != nil {
		return err
	}
```

controllers/datadogmonitor/controller.go
```diff
@@ -0,0 +1,325 @@
...
+	logger.Info("Created a new DatadogMonitor", "datadogMonitor.Namespace", datadogMonitor.Namespace, "datadogMonitor.Name", datadogMonitor.Name, "m.ID", m.GetId())
+
+	return nil
+}
+
+func (r *Reconciler) update(logger logr.Logger, datadogMonitor *datadoghqv1alpha1.DatadogMonitor, status *datadoghqv1alpha1.DatadogMonitorStatus, now metav1.Time) error {
+	// Validate monitor in Datadog
+	err := validateMonitor(r.datadogAuth, r.datadogClient, datadogMonitor)
+	if err != nil {
+		return err
```
clamoriniere, line 199: it can be good to enrich the error maybe with more context

controllers/datadogmonitor/monitor.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// TODO
+	o := datadogapiclientv1.MonitorOptions{}
+
+	m := datadogapiclientv1.NewMonitor()
+	m.SetType(monitorType)
+	m.SetQuery(query)
+	m.SetName(name)
+	m.SetMessage(msg)
+	m.SetOptions(o)
```
clamoriniere, line 32: maybe it is overkill but I like to scope things

```suggestion
	m := datadogapiclientv1.NewMonitor()
	{
	    m.SetType(monitorType)
	    m.SetQuery(query)
	    m.SetName(name)
	    m.SetMessage(msg)
	    m.SetOptions(o)
	}
```

I looked if it was able to chain `Set()` function it is not possible :(

controllers/setup.go
```diff
@@ -35,11 +44,27 @@ func SetupControllers(mgr manager.Manager, supportExtendedDaemonset bool) error
...
+			SupportExtendedDaemonset: options.SupportExtendedDaemonset,
 		},
 	}).SetupWithManager(mgr); err != nil {
 		return fmt.Errorf("unable to create controller DatadogAgent: %w", err)
 	}
 
+	ddClient, err := datadogclient.InitDatadogClient(options.APIKey, options.AppKey)
+	if err != nil {
+		return fmt.Errorf("unable to create Datadog API Client: %w", err)
+	}
```
clamoriniere, line 56: does it means that api and app key are mandatory to start the operator?

clamoriniere: we can work on it in another PR.

clamoriniere (commented):
Looks awsome.
Thanks for putting some much effort on unit-testing this feature.

Most of my comments are only nits.

controllers/datadogmonitor/controller.go
celenechang: thanks; i added a monitor ID to the error log where `r.update` is called, but let me know if you think additional info should be added (or moved here)

controllers/setup.go
celenechang: yes, at the moment. open to other ideas since it will be a breaking change

controllers/datadogmonitor/monitor.go
```diff
@@ -0,0 +1,117 @@
...
+	return nil
+}
+
+func translateClientError(err error, msg string) error {
+	if msg == "" {
+		msg = "an error occurred"
+	}
+
+	if apiErr, ok := err.(datadogapiclientv1.GenericOpenAPIError); ok {
+		return fmt.Errorf(msg+": %v: %s", err, apiErr.Body())
```
clamoriniere, line 109: nit: when you wrap and error into another error with `fmt.Errorf()` it is good to use `%w` instead of `%v`. like this we can `Unwrap` the error later. [doc](https://golang.org/pkg/errors/)

```suggestion
		return fmt.Errorf(msg+": %w: %s", err, apiErr.Body())
```

controllers/datadogmonitor/monitor.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	if errURL, ok := err.(*url.Error); ok {
+		return fmt.Errorf(msg+" (url.Error): %s", errURL)
+	}
+
+	return fmt.Errorf(msg+": %s", err.Error())
```
clamoriniere, line 116:
```suggestion
	return fmt.Errorf(msg+": %w", err)
```

controllers/setup.go
clamoriniere: until we don't release an new operator with this change it is only a work in progress :)

clamoriniere (commented):
juste a few additional nits. otherwise 👍

controllers/datadogmonitor/utils.go
```diff
@@ -0,0 +1,13 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2020 Datadog, Inc.
+
+package datadogmonitor
+
+func getMax(val1, val2 int64) int64 {
```
vboulineau, line 8: nit: should probably be in `pkg/utils`

controllers/datadogmonitor_controller.go
```diff
@@ -1,42 +1,62 @@
...
+
+	internal, err := datadogmonitor.NewReconciler(r.Client, r.DDClient, r.VersionInfo, r.Scheme, r.Log, r.Recorder)
+	if err != nil {
+		return err
+	}
+	r.internal = internal
+
+	builder := ctrl.NewControllerManagedBy(mgr).
+		For(&datadoghqv1alpha1.DatadogMonitor{})
+		// Owns(&corev1.Secret{}).
```
vboulineau, line 54: nit: Should be removed if not sued

controllers/setup.go
vboulineau: I think it should lazily instantiated and/or be a singleton

clamoriniere: the problem will that we can have concurrent client initialisation in this case.

api/v1alpha1/datadogmonitor_types.go
```diff
@@ -140,9 +150,9 @@ const (
 // The DatadogMonitor is triggering if one of its groups is in Alert, Warn, or No Data
 type DatadogMonitorTriggeredState struct {
 	// MonitorGroup is the name of the triggering group
-	MonitorGroup       string              `json:"monitorGroup,omitempty"`
-	State              DatadogMonitorState `json:"state,omitempty"`
-	LastTransitionTime metav1.Time         `json:"lastTransitionTime,omitempty"`
+	MonitorGroup     string              `json:"monitorGroup,omitempty"`
+	State            DatadogMonitorState `json:"state,omitempty"`
+	LastTransitionTs int64               `json:"lastTransitionTime,omitempty"`
```
xornivore, line 155: Maybe missed a prior comment, but curious why this change?

api/v1alpha1/datadogmonitor_validation.go
```diff
@@ -0,0 +1,36 @@
...
+	utilserrors "k8s.io/apimachinery/pkg/util/errors"
+)
+
+// IsValidDatadogMonitor use to check if a DatadogMonitorSpec is valid by checking
+// that the required fields are defined
+func IsValidDatadogMonitor(spec *DatadogMonitorSpec) error {
+	var errs []error
+	var err error
+	if spec.Query == "" {
+		errs = append(errs, fmt.Errorf("invalid spec.Query, err: %v", err))
```
xornivore, line 20: `err` looks to be `nil` here.

api/v1alpha1/datadogmonitor_validation_test.go
```diff
@@ -0,0 +1,84 @@
...
+		{
+			name:  "monitor missing message",
+			spec:  missingMessage,
+			isErr: true,
+		},
+	}
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+			result := IsValidDatadogMonitor(test.spec)
+			if test.isErr {
```
xornivore, line 77: Any reason not to `assert` on a specific error message in this test?

controllers/datadogmonitor/controller.go
```diff
@@ -0,0 +1,318 @@
...
+		logger.Error(err, "Error generating hash")
+		return r.updateStatusIfNeeded(logger, instance, now, newStatus, err, result)
+	}
+
+	statusSpecHash := instance.Status.CurrentHash
+
+	// Create or update monitor, or check monitor state. Fall through this block (without returning)
+	// if the result should be requeued with the default period
+	if instance.Status.ID == 0 {
+		// If the monitor ID is 0, then it doesn't exist yet in Datadog. Create the monitor (only metric alerts)
```
xornivore, line 114: This would be a great spot to log that we are about to create monitor?

controllers/datadogmonitor/finalizer.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// Proceed in reconcile loop
+	return ctrl.Result{}, nil
+}
+
+func (r *Reconciler) finalizeDatadogMonitor(logger logr.Logger, dm *datadoghqv1alpha1.DatadogMonitor) {
+	if dm.Status.Primary {
+		err := deleteMonitor(r.datadogAuth, r.datadogClient, dm.Status.ID)
+		if err != nil {
+			logger.Error(err, "failed to finalize monitor")
```
xornivore, line 55: Should probably add more detail about which monitor failed here.

main.go
```diff
@@ -126,20 +126,18 @@ func main() {
 	customSetupHealthChecks(mgr)
 	customSetupEndpoints(pprofActive, mgr)
 
+	options := controllers.SetupOptions{
+		SupportExtendedDaemonset: supportExtendedDaemonset,
+		APIKey:                   os.Getenv(config.DDAPIKeyEnvVar),
```
xornivore, line 131: Would there be a case for supporting multiple sets of creds in the future? Example that comes to mind - monitors created by different teams operating on the same cluster. We currently support one set of creds per DatadogAgent resource, this difference with monitors might be confusing. At the same time DatadogAgent is usually a singleton, we don't have a good multi-tenancy story for it.

api/v1alpha1/datadogmonitor_types.go
celenechang: i think i changed it to be consistent with how timestamps are reported by the API ([link](https://github.com/DataDog/datadog-api-client-go/blob/80c5e2c698b188780cb9e71debc9cd95fc2f0df3/api/v1/datadog/model_monitor_state_group.go#L18))

api/v1alpha1/datadogmonitor_validation.go
celenechang: good catch, thanks

main.go
celenechang: i **think** i discussed this with someone (vincent?) before, and it was decided to have a separate controller per namespace that would be watched independently and could have different credentials. i think this would cover most cases of teams operating on the same cluster; what do you think?

xornivore: That makes a lot of sense, thank you.

api/v1alpha1/datadogmonitor_types.go
xornivore: Interesting, I wonder if it impacts the readability of conditions in the status as seen in `kubectl`?

controllers/datadogmonitor/controller.go
celenechang: there's a log of when a monitor is created; think there should be one prior to creation as well?

api/v1alpha1/datadogmonitor_types.go
celenechang: you're right, it was ugly. i switched it back

controllers/datadogmonitor/controller.go
xornivore: Not a big deal, but might be helpful in troubleshooting, but iirc there isn't Debug level, only Info?

xornivore (approved):
👍

vboulineau (approved):

PR #204, opened 2021-01-12
Add the possibility to configure the dogstatsd/apm socket host path

clamoriniere:
### What does this PR do?

* Add the possibility to override the default host filepath used for the dockstatsd socket
* Add the possibility to override the default host filepath used for the APM socket
* Add missing HostPath `Volumes` for the dogstatsd and APM sockets.

### Motivation

Support properly the dogstatsd configuration using a unix-socket

### Additional Notes

N/A

### Describe your test plan

Create a new DatadogAgent resource with dogstatsd and apm unix-socket enabled and override the default socket path with the option `hostSocketPath`. the `examples/datadog-agent-unix-socket.yaml` can help here.
The socket host path should be created according to the configuration


controllers/datadogagent/utils.go
```diff
@@ -825,6 +825,23 @@ func getVolumesForAgent(dda *datadoghqv1alpha1.DatadogAgent) []corev1.Volume {
 		volumes = append(volumes, volume)
 	}
 
+	// Dogstatsd volume
+	if datadoghqv1alpha1.BoolValue(dda.Spec.Agent.Config.Dogstatsd.UseDogStatsDSocketVolume) {
+		volumeType := corev1.HostPathDirectoryOrCreate
+		hostPath := *dda.Spec.Agent.Config.Dogstatsd.HostSocketPath
+
+		dsdsockerVolume := corev1.Volume{
+			Name: datadoghqv1alpha1.DogstatsdSockerVolumeName,
```
L3n41c, line 834:
```suggestion
			Name: datadoghqv1alpha1.DogstatsdSocketVolumeName,
```

Typo to be fixed at the other places in the code.

L3n41c (commented):
The implementation looks good to me.
There’s just a typo which was present before this change.

Concerning the feature, I’m wondering if there is a need to be able to have different directories for each socket that the agents can bind (`dogstatsd` and `trace-agent`).
Because it adds yet another parameter to the config.

If there is a need to be so highly configurable, this new parameter allows only to configure the directory path, not the socket full path.

In order to be full configurable, the parameter should take the full path (`/var/run/datadog/dsd.socket`) and the operator should split it between the directory path (used to build the `hostVolume` path) and the basename to be passed as `DD_DOGSTATSD_SOCKET` once concatenated with the mountpoint of the `hostVolume` path in the container.

controllers/datadogagent/utils.go
clamoriniere: thanks for catching this typo

clamoriniere:
@L3n41c 
thank for the feedback. to be honest I did the same configuration option than in the helm chart. I need to think about your idea.
It is like this we manage the container runtime socket?

controllers/datadogagent/utils.go
```diff
@@ -825,6 +845,23 @@ func getVolumesForAgent(dda *datadoghqv1alpha1.DatadogAgent) []corev1.Volume {
 		volumes = append(volumes, volume)
 	}
 
+	// Dogstatsd volume
+	if datadoghqv1alpha1.BoolValue(dda.Spec.Agent.Config.Dogstatsd.UnixDomainSocket.Enabled) {
+		volumeType := corev1.HostPathDirectoryOrCreate
+		hostPath := getDirFromFilepath(*dda.Spec.Agent.Config.Dogstatsd.UnixDomainSocket.HostFilepath)
+
+		dsdsockerVolume := corev1.Volume{
```
L3n41c, line 853:
```suggestion
		dsdsocketVolume := corev1.Volume{
```

controllers/datadogagent/utils.go
```diff
@@ -825,6 +845,23 @@ func getVolumesForAgent(dda *datadoghqv1alpha1.DatadogAgent) []corev1.Volume {
...
+			Name: datadoghqv1alpha1.DogstatsdSocketVolumeName,
+			VolumeSource: corev1.VolumeSource{
+				HostPath: &corev1.HostPathVolumeSource{
+					Path: hostPath,
+					Type: &volumeType,
+				},
+			},
+		}
+		volumes = append(volumes, dsdsockerVolume)
```
L3n41c, line 862:
```suggestion
		volumes = append(volumes, dsdsocketVolume)
```

L3n41c (commented):
LGTM
I only spotted one typo.

L3n41c (approved):
LGTM 👍🏻

PR #205, opened 2021-01-14
Use DD_EXTERNAL_METRICS_PROVIDER_ENDPOINT

juliogreff:
### What does this PR do?

We used to use DATADOG_HOST for setting the external metrics endpoint,
since the Cluster Agent did not have specific support for setting the
external metrics endpoint directly. This has since been added:
https://github.com/DataDog/datadog-agent/pull/6952 -- so we use that and
remove all the decision making and fallbacks to the DCA itself.

### Describe your test plan

Edit a `DatadogAgent` to add the key below, and check the DCA deployment to see the `DD_EXTERNAL_METRICS_PROVIDER_ENDPOINT` applied for the container with the appropriate value.

````
kind: DatadogAgent
metadata:
  name: datadog
spec:
  clusterAgent:
    config:
      externalMetrics:
        enabled: true
        endpoint: "https://example.com"
```

### DIFF
--- a/api/v1alpha1/const.go
+++ b/api/v1alpha1/const.go
@@ -97,6 +97,7 @@ const (
 	DDRuntimeSecurityConfigPoliciesDir           = "DD_RUNTIME_SECURITY_CONFIG_POLICIES_DIR"
 	DDRuntimeSecurityConfigSocket                = "DD_RUNTIME_SECURITY_CONFIG_SOCKET"
 	DDRuntimeSecurityConfigSyscallMonitorEnabled = "DD_RUNTIME_SECURITY_CONFIG_SYSCALL_MONITOR_ENABLED"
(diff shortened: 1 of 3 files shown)


vboulineau:
I think we should still honor `DD_SITE`. If I set it, I expect everything to go there, whether I took care of setting other variables or not.
Of course if `DD_EXTERNAL_METRICS_PROVIDER_ENDPOINT` is set, it gets priority.

However, I'm thinking that this code should be in the Agent, not in the operator

juliogreff:
@vboulineau that's what I meant by "remove all the decision making and fallbacks to the DCA itself", that's what this PR does: https://github.com/DataDog/datadog-agent/pull/6952

vboulineau:
Yes, my bad, I did not check what `GetMainEndpoint` was doing, smarter that I thought

vboulineau (approved):

PR #210, opened 2021-01-25
Increase to 250Mi the operator default resources

clamoriniere:
### What does this PR do?

Increase to 250Mi the operator default resources

### Motivation

with only 100Mi the operator can be in crashloopbackoff when it is deployed in a large cluster.

### Additional Notes

### Describe your test plan

Generate operator bundle and check that the `ClusterServiceVersion` resources contains the new settings

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -43,10 +43,10 @@ spec:
         resources:
           limits:
             cpu: 100m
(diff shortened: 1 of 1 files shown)


vboulineau (approved):

vboulineau:
Should be in Helm chart too?

PR #195, opened 2020-12-11
remove mounting of clusterID CM from process-agent

hkaj:
### What does this PR do?

remove mounting of clusterID CM from process-agent

### Motivation

https://github.com/DataDog/datadog-agent/pull/6925 removes the need for this step, the Process Agent will get the cluster-id from a Cluster Agent API endpoint instead, which has a nicer failure mode.

### Additional Notes

This change requires Agent >= 7.25 and Cluster Agent >= 1.11. I'm not sure how to handle this nicely for older versions. Maybe we should keep the configmap for now? I'm open to suggestions.

### Describe your test plan

Write there any instructions and details you may have to test your PR.

PR #223, opened 2021-02-11
[Fix][OrchExplorer] Allow the process agent to connect to the DCA

ahmed-mez:
### What does this PR do?

Inject the required env vars to let the process agent communicate with the DCA

### Motivation
The process agent gets the cluster ID from the DCA

### Additional Notes

Fix https://github.com/DataDog/datadog-operator/pull/195

### Describe your test plan

Same as https://github.com/DataDog/datadog-operator/pull/195

### DIFF
--- a/controllers/datadogagent/utils.go
+++ b/controllers/datadogagent/utils.go
@@ -501,6 +501,9 @@ func getEnvVarsForProcessAgent(dda *datadoghqv1alpha1.DatadogAgent) ([]corev1.En
 			return nil, err
 		}
 		envVars = append(envVars, envs...)
+
+		// The process agent retrieves the cluster id from the Cluster Agent
+		envVars = append(envVars, envForClusterAgentConnection(dda)...)
 	}
 
 	commonEnvVars, err := getEnvVarsCommon(dda, true)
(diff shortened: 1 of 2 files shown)


controllers/datadogagent/utils.go
```diff
@@ -501,6 +501,9 @@ func getEnvVarsForProcessAgent(dda *datadoghqv1alpha1.DatadogAgent) ([]corev1.En
 			return nil, err
 		}
 		envVars = append(envVars, envs...)
+
+		// The process agent retrieves the cluster id from the Cluster Agent
+		envVars = append(envVars, envForClusterAgentConnection(dda)...)
```
L3n41c, line 506: Inside `getEnvVarsForAgent`, the appending of `envForClusterAgentConnection(dda)…` is gated by a `if spec.ClusterAgent != nil`.
Shouldn’t we also add the DCA env vars to the `process-agent` only if the DCA is enabled ?

ahmed-mez: that's why I've put a `if dda.Spec.ClusterAgent != nil` inside `envForClusterAgentConnection` as a safegard

ahmed-mez: i.e `envForClusterAgentConnection` returns an empty list if the DCA is disabled

controllers/datadogagent/utils.go
```diff
@@ -670,20 +673,7 @@ func getEnvVarsForAgent(dda *datadoghqv1alpha1.DatadogAgent) ([]corev1.EnvVar, e
 	}
 
 	if spec.ClusterAgent != nil {
```
L3n41c, line 675: This `if` could be removed since it’s already present inside `envForClusterAgentConnection`.

L3n41c (approved):

controllers/datadogagent/utils.go
L3n41c: Indeed ! 👍🏻

controllers/datadogagent/utils.go
ahmed-mez: not removed but we can move it under, because `if datadoghqv1alpha1.BoolValue(spec.ClusterAgent.Config.ClusterChecksEnabled)` and lines under it are relying on it

PR #234, opened 2021-02-18
Allow custom VolumeMounts for all agent containers

juliogreff:
### What does this PR do?

Previously, it was only possible for the main agent container to have
user-specified VolumeMounts. This commit adds the possibility for each
container (process, apm, security) to have their own, specified
separately.

### Motivation

Internally, we want to have the agent auth_token in a different path, and we want that token to be available to all agents.

### Describe your test plan

Specify a VolumeMount for each of the agents, and check that those volumes get mounted as specified.

### DIFF
--- a/api/v1alpha1/datadogagent_types.go
+++ b/api/v1alpha1/datadogagent_types.go
@@ -295,6 +295,13 @@ type APMSpec struct {
 	// +listMapKey=name
 	Env []corev1.EnvVar `json:"env,omitempty"`
 
+	// Specify additional volume mounts in the APM Agent container
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	// +listMapKey=mountPath
+	VolumeMounts []corev1.VolumeMount `json:"volumeMounts,omitempty"`
+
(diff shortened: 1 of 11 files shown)


api/v1alpha1/datadogagent_types.go
```diff
@@ -295,6 +295,13 @@ type APMSpec struct {
 	// +listMapKey=name
 	Env []corev1.EnvVar `json:"env,omitempty"`
 
+	// Specify additional volume mounts in the Process Agent container
```
clamoriniere, line 298:
```suggestion
	// Specify additional volume mounts in the APM/Trace Agent container
```

clamoriniere (commented):
looks great! just a small comments

api/v1alpha1/datadogagent_types.go
juliogreff: There's three separate ones, but indeed the comment is wrong for this one. Nice spot, fixed!

sarina-dd (approved):

clamoriniere (approved):

PR #235, opened 2021-02-18
Default ClusterChecksRunner if ClusterChecksEnabled == true

### LINKED ISSUE #232: Different behaviour for creation of objects and update of objects for cluster checks runners deployment
OK, that was kind of a cryptic title for a bug report, but I couldn't find a better one, so let me explain with examples what happened:

This is using operator `0.5.0-RC2`. 

If I deploy this configuration for the first time in a 1 node cluster:

```
kind: DatadogAgent
metadata:
  name: datadog
spec:
  credentials:
    apiSecret:
      secretName: datadog-secret
      keyName: api-key
    appSecret:
      secretName: datadog-secret
      keyName: app-key
  agent:
    image:
      name: "datadog/agent:latest"
  clusterAgent:
    image:
      name: "datadog/cluster-agent:latest"
    config:
      clusterChecksEnabled: true
```

I correctly get 1 pod for the node agent, one pod for the cluster agent and 1 pod for the cluster checks runner:

```
NAME      ACTIVE   AGENT             CLUSTER-AGENT     CLUSTER-CHECKS-RUNNER   AGE
datadog   True     Running (1/1/1)   Running (1/1/1)           Running (1/1/1)         112s
```

But, if I first apply the same configuration, but without enabling cluster checks, like this one:

```
kind: DatadogAgent
metadata:
  name: datadog
spec:
  credentials:
    apiSecret:
      secretName: datadog-secret
      keyName: api-key
    appSecret:
      secretName: datadog-secret
      keyName: app-key
  agent:
    image:
      name: "datadog/agent:latest"
  clusterAgent:
    image:
      name: "datadog/cluster-agent:latest"
```

so, as expected, only get 1 node agent and 1 cluster agent:

```
NAME      ACTIVE   AGENT             CLUSTER-AGENT     CLUSTER-CHECKS-RUNNER   AGE
datadog   True     Running (1/1/1)   Running (1/1/1)                           2m57s
```

But, if I now edit the object and add the cluster checks option back, so it looks like the original object:

```
kind: DatadogAgent
metadata:
  name: datadog
spec:
  credentials:
    apiSecret:
      secretName: datadog-secret
      keyName: api-key
    appSecret:
      secretName: datadog-secret
      keyName: app-key
  agent:
    image:
      name: "datadog/agent:latest"
  clusterAgent:
    image:
      name: "datadog/cluster-agent:latest"
    config:
      clusterChecksEnabled: true
```

The operator knows that there has been an update and start updating the existing objects, but it won't create a cluster checks runner replica:

```
NAME      ACTIVE   AGENT              CLUSTER-AGENT      CLUSTER-CHECKS-RUNNER   AGE
datadog   True     Updating (1/1/0)   Updating (2/1/1)                                                          9m30s
```

```
NAME                                      READY   STATUS    RESTARTS   AGE
datadog-agent-pzhh2                       1/1     Running   0          29s
datadog-cluster-agent-59958b4dff-whwqr    1/1     Running   0          34s
```

I would expect the reconciliation loop to work in a way that sees that that configuration requires 1 cluster check runner and will create it when updating the object

juliogreff:
### What does this PR do?

Creating a DatadogAgent with ClusterChecksEnabled did not have an
equivalent effect of creating a DatadogAgent without
ClusterChecksEnabled and then updating it to have ClusterChecksEnabled
set to true. This happened because, in the latter, the object was
already considered defaulted, but did not have a ClusterChecksRunner
spec.

Now, when we check the ClusterAgent spec, we also check that
ClusterChecksRunner is also set when applicable.

### Motivation

Fixes #232.

### Describe your test plan

See #232.

### DIFF
--- a/api/v1alpha1/datadogagent_default.go
+++ b/api/v1alpha1/datadogagent_default.go
@@ -135,6 +135,10 @@ func IsDefaultedDatadogAgent(ad *DatadogAgent) bool {
 		if ad.Spec.ClusterAgent.Replicas == nil {
 			return false
 		}
+
+		if BoolValue(ad.Spec.ClusterAgent.Config.ClusterChecksEnabled) && ad.Spec.ClusterChecksRunner == nil {
+			return false
+		}
 	}
 
 	if ad.Spec.ClusterChecksRunner != nil {


clamoriniere (approved):
👍  nice catch

ruthnaebeck (approved):
👍  for docs

PR #236, opened 2021-02-18
Support Prometheus Scrape configuration

ahmed-mez:
### What does this PR do?

Support configuring Prometheus Autodiscovery.

### Motivation

Ease configuring of the feature

### Additional notes

Took the opportunity to align a naming inconsistency (`dda` vs `agentdeployment`)

### Describe your test plan

Apply `examples/datadog-agent-with-prometheus-autodiscovery.yaml` and/or `examples/datadog-agent-with-prometheus-autodiscovery-advanced-config.yaml`, `agent config` and the agent behaviour must reflect the config in both the node agent and the DCA. (requires Agent 7.27 and DCA 1.12)


controllers/datadogagent/utils.go
```diff
@@ -1694,6 +1699,41 @@ func isKSMCoreEnabled(dda *datadoghqv1alpha1.DatadogAgent) bool {
...
+			envVars = append(envVars, corev1.EnvVar{
+				Name:  datadoghqv1alpha1.DDPrometheusScrapeServiceEndpoints,
+				Value: datadoghqv1alpha1.BoolToString(dda.Spec.Features.PrometheusScrape.ServiceEndpoints),
+			})
+		}
+
+		if dda.Spec.Features.PrometheusScrape.AdditionalConfigs != nil {
+			jsonValue, err := yaml.YAMLToJSON([]byte(*dda.Spec.Features.PrometheusScrape.AdditionalConfigs))
+			if err != nil {
+				logger.Error(err, "Invalid additional prometheus config, ignoring it")
```
xornivore, line 1724: What do you think about returning an error from here on invalid YAML and converting that into a reconciliation error visible on the `DatadogAgent` resource. This seems like a good case for letting the user know their config is invalid, otherwise they'd have to track down in logs, which may not be trivial for some users.

xornivore (commented):
Looks great, just a comment about user visible error handling.

controllers/datadogagent/utils.go
ahmed-mez: I thought about it actually, and I choose to log the error and unblock the reconciling for 2 reasons
- The config is not for a critical feature in the agent, I found it too agressive to allow it to block the reconciliation
- The operator doesn't provide an agent config validation (example: the agent env vars, `confd` files, `datadog.yaml`) all can be passed via the CRD and the agent will then validate them. That being said, returning the error here is not aligned with what the operator does elsewhere.

xornivore: Interesting, ok then, I do wish we did more validation though of those config settings as well. Even if we don't want to bock we could add a "misconfiguration detected" condition. Something to think about later though.

xornivore (approved):

PR #245, opened 2021-02-25
[DatadogMonitor] Refactor creds management and support SB

ahmed-mez:
### What does this PR do?

- Add an argument to disable the `DatadogMonitor` controller
- Centralise the credentials management in the operator, which will allow the `DatadogMonitor` controller to get creds via secret backend
- If one of the controllers couldn't start, we don't `os.Exit(1)` anymore (example: if you don't provide creds, `DatadogMonitor` won't start, but `DatadogAgent` will do. 

### Motivation

- `DatadogMonitor` supports SB now
- `DatadogMonitor` can be disabled

### Additional Notes

- A follow-up PR (or commit) to revamp the metrics forwarder will come up next

### Describe your test plan

- Do not provide creds, this should prevent `DatadogMonitor` from starting but shouldn't affect `DatadogAgent`
- Try to provide creds via secret backend, `DatadogMonitor` should be able to get them and start (you can use a secret mount + `readsecret.sh`)
- Configure the `datadogMonitorEnabled` operator argument, make sure `DatadogMonitor` doesn't start

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -32,10 +32,6 @@ spec:
         imagePullPolicy: Always
         name: manager
         env:
-        - name: DD_API_KEY
-          value: "dummy_api_key"
-        - name: DD_APP_KEY
-          value: "dummy_app_key"
         - name: POD_NAME
           valueFrom:
             fieldRef:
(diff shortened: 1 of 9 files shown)


xornivore (approved):
Great work!

Issue #237, opened 2021-02-19
The operator deletes an existing external metrics API service when `externalMetrics` is set to `false` or unset


bruecktech:
**Describe what happened:**
The operator deletes an existing external metrics APIService when `externalMetrics` is set to `false` or unset as `false` is the default.
This is most likely because it calls the `cleanupMetricsServerAPIService` function [here](https://github.com/DataDog/datadog-operator/blob/c1f08be846985d7e087e573fa036f49af9ac165f/controllers/datadogagent/service.go#L113-L114) when external metrics it's not enabled but it does not check if the existing APIService was actually created by Datadog

Currently this blocks us from using the operator and it broke our HPA based on CloudWatch metrics.

**Describe what you expected:**
I would expect the operator to leave the external metrics APIService in place when they are not owned by the operator

**Steps to reproduce the issue:**

configure an external metric APIService like this
```
apiVersion: apiregistration.k8s.io/v1beta1
kind: APIService
metadata:
  name: v1beta1.external.metrics.k8s.io
spec:
  service:
    name: my-adapter
    namespace: my-namespace-not-used-by-datadog
  group: external.metrics.k8s.io
  version: v1beta1
  insecureSkipTLSVerify: true
  groupPriorityMinimum: 100
  versionPriority: 100
```

Then deploy the datadog operator with `externalMetrics: false`  or unset
After a while it will remove the existing metrics APIService.

**Additional environment details (Operating System, Cloud provider, etc):**
* Datadog operator 0.4.0 deployed into separate `datadog` namespace
* Kubernetes 1.18
* Cloud Provider: AWS

PR #247, opened 2021-02-25
Fix unwanted Resources deletion

clamoriniere:
### What does this PR do?

This PR avoid resources deletion not owned by the DatadogAgent Resource instance.

### Motivation

Fix: #237

The Datadog-operator was deleting resources without checking if it was the Owner of these resources.

### Additional Notes

N/A

### Describe your test plan

You can follow the test plan proposed in #237


xornivore (approved):

clamoriniere:
@xornivore 
could you check again? I have update logic for other resources.

xornivore (approved):

PR #248, opened 2021-02-26
Rework the creds manager

ahmed-mez:
### What does this PR do?

- Disable/Enable the metrics forwarder via `operatorMetricsEnabled`
- The metrics forwarder uses `pkg/config` to get creds from the operator config as a fallback
- Add a creds caching in `GetCredentials`
- Add an exp backoff in `GetCredentials`
- Disable Datadog Monitor by default
- Restart the operator if Datadog Monitor is enabled + no creds provided 

### Motivation

Follow-up on https://github.com/DataDog/datadog-operator/pull/245

### Describe your test plan

- Set the `operatorMetricsEnabled` operator argument to `false`, make sure the metrics forwarder doesn't start
- Enable Datadog Monitor by default, make sure the operator starts only when creds are provided

### DIFF
--- a/pkg/config/creds.go
+++ b/pkg/config/creds.go
@@ -69,28 +69,23 @@ func (cm *CredentialManager) GetCredentials() (Creds, error) {
 		encrypted = append(encrypted, appKey)
 	}
 
-	if len(encrypted) == 0 {
-		// Nothing to decrypt
-		creds := Creds{APIKey: apiKey, AppKey: appKey}
-		cm.cacheCreds(creds)
-		return creds, nil
-	}
+	if len(encrypted) > 0 {
+		decrypted := map[string]string{}
+		var decErr error
+		if err := retry.OnError(cm.decryptorBackoff, secrets.Retriable, func() error {
+			decrypted, decErr = cm.secretBackend.Decrypt(encrypted)
+			return decErr
+		}); err != nil {
+			return Creds{}, err
+		}
 
-	decrypted := map[string]string{}
-	var decErr error
-	if err := retry.OnError(cm.decryptorBackoff, secrets.Retriable, func() error {
-		decrypted, decErr = cm.secretBackend.Decrypt(encrypted)
-		return decErr
-	}); err != nil {
-		return Creds{}, err
-	}
+		if val, found := decrypted[apiKey]; found {
+			apiKey = val
+		}
 
-	if val, found := decrypted[apiKey]; found {
-		apiKey = val
-	}
-
-	if val, found := decrypted[appKey]; found {
-		appKey = val
+		if val, found := decrypted[appKey]; found {
+			appKey = val
+		}
 	}
 
 	creds := Creds{APIKey: apiKey, AppKey: appKey}
--- a/pkg/config/creds.go
+++ b/pkg/config/creds.go
@@ -8,6 +8,7 @@ package config
 import (
 	"errors"
 	"os"
+	"sync"
 
 	"github.com/DataDog/datadog-operator/pkg/secrets"
 )
@@ -18,15 +19,29 @@ type Creds struct {
 	AppKey string
 }
 
+// CredentialManager provides the credentials from the operator configuration
+type CredentialManager struct {
+	secretBackend secrets.Decryptor
+	creds         Creds
+	credsMutex    sync.Mutex
+}
+
+// NewCredentialManager returns a CredentialManager
+func NewCredentialManager() *CredentialManager {
+	return &CredentialManager{
+		secretBackend: secrets.NewSecretBackend(),
+		creds:         Creds{},
+	}
+}
+
 // GetCredentials returns the API and APP keys respectively from the operator configurations
 // This function tries to decrypt the secrets using the secret backend if needed
 // It returns an error if the creds aren't configured or if the secret backend fails to decrypt
-func GetCredentials() (Creds, error) {
-	return getCredentials(secrets.NewSecretBackend())
-}
+func (cm *CredentialManager) GetCredentials() (Creds, error) {
+	if creds, found := cm.getCredsFromCache(); found {
+		return creds, nil
+	}
 
-// getCredentials used to ease testing
-func getCredentials(decryptor secrets.Decryptor) (Creds, error) {
 	apiKey := os.Getenv(DDAPIKeyEnvVar)
 	appKey := os.Getenv(DDAppKeyEnvVar)
 
@@ -45,10 +60,12 @@ func getCredentials(decryptor secrets.Decryptor) (Creds, error) {
 
 	if len(encrypted) == 0 {
 		// Nothing to decrypt
-		return Creds{APIKey: apiKey, AppKey: appKey}, nil
+		creds := Creds{APIKey: apiKey, AppKey: appKey}
+		cm.cacheCreds(creds)
+		return creds, nil
 	}
 
-	decrypted, err := decryptor.Decrypt(encrypted)
+	decrypted, err := cm.secretBackend.Decrypt(encrypted)
 	if err != nil {
 		return Creds{}, err
 	}
@@ -61,5 +78,22 @@ func getCredentials(decryptor secrets.Decryptor) (Creds, error) {
 		appKey = val
 	}
 
-	return Creds{APIKey: apiKey, AppKey: appKey}, nil
+	creds := Creds{APIKey: apiKey, AppKey: appKey}
+	cm.cacheCreds(creds)
+	return creds, nil
+}
+
+func (cm *CredentialManager) cacheCreds(creds Creds) {
+	cm.credsMutex.Lock()
+	defer cm.credsMutex.Unlock()
+	cm.creds = creds
+}
+
+func (cm *CredentialManager) getCredsFromCache() (Creds, bool) {
+	cm.credsMutex.Lock()
+	defer cm.credsMutex.Unlock()
+	if cm.creds.APIKey != "" && cm.creds.AppKey != "" {
+		return cm.creds, true
+	}
+	return Creds{}, false
 }
--- a/pkg/config/creds.go
+++ b/pkg/config/creds.go
@@ -9,8 +9,12 @@ import (
 	"errors"
 	"os"
 	"sync"
+	"time"
 
 	"github.com/DataDog/datadog-operator/pkg/secrets"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
 )
 
 // Creds holds the api and app keys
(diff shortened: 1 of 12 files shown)


pkg/config/creds.go
```diff
@@ -18,6 +19,10 @@ type Creds struct {
 	AppKey string
 }
 
+const credsKey = "creds"
```
clamoriniere, line 22: could you add a description for this const value.

pkg/config/creds.go
```diff
@@ -18,6 +19,10 @@ type Creds struct {
...
+
+var credsCache sync.Map
```
clamoriniere, line 24: can we manage the credsCache in a struct. instead of a global env var? 

```suggestion
type CredentialManager struct {
    Cache sync.Map
}

func NewCredentialManager() *CredentialManager {
    return &CredentialManager{}
}
```

clamoriniere: The idea will be to create the `CredentialManager` in the `main()` func and pass it to the others components.

clamoriniere (commented):
Looks good.
Just a comment about how you can remove a "global variable".

pkg/config/creds.go
ahmed-mez: `credsCache` is a local variable and isn't accesses by any other component, it's just used by `GetCredentials`. Not sure I understand the benefit of creating and passing a `CredentialManager` struct to the other components in main? we will end up with multiple/identical structs instantiated while a public util `GetCredentials` can be enough imho

pkg/controller/utils/datadog/metrics_forwarder.go
```diff
@@ -489,31 +488,38 @@ func (mf *metricsForwarder) getDatadogAgent() (*datadoghqv1alpha1.DatadogAgent,
 }
 
 // getCredentials returns the Datadog API Key and APP Key, it returns an error if one key is missing
+// getCredentials tries to get the credentials from the CRD, then from operator configuration
 func (mf *metricsForwarder) getCredentials(dda *datadoghqv1alpha1.DatadogAgent) (string, string, error) {
+	apiKey, appKey, err := mf.getCredsFromDatadogAgent(dda)
+	if err != nil {
+		if errors.Is(err, ErrEmptyAPIKey) || errors.Is(err, ErrEmptyAPPKey) {
```
L3n41c, line 495: This means that we forbid a behaviour where the API key could be retrieved from the `datadog` CR and only the APP key fetch is fallen back to the operator config.
As soon as one of the two keys cannot be retrieved from the CR, the two are fetched from the operator config even if the other key in the CR is valid.

I don’t know if there a real need for such a setup, but a user who defines only an API key in its CRD might expect to have the API key of the CRD used and to have only the APP key fallen back to the operator one.

At least, this differs from the previous behaviour.

L3n41c (commented):
I let a comment about a change in the behaviour.

pkg/config/creds.go
L3n41c, line 24: The use of a `Map` seems a little bit overkill here since it will ever contain only a single element.
You replace it by:
```suggestion
var credsCache Creds
var sync.Mutex CredsMutex
```
I agree that in this case, we’ll have to handle a Mutex in `cacheCreds`, `getCredsFromCache` and `resetCredsCache`.

pkg/controller/utils/datadog/metrics_forwarder.go
ahmed-mez: TBH I did this on purpose, I'm thinking it's more natural to retrieve the app and api keys from the same place either the CRD or the operator config. Especially with the Datadog Monitor controller where the only source is the operator config.

pkg/config/creds.go
clamoriniere: for example during unittest. 2 tests running in // will use the same cache.

ahmed-mez: Yes, I applied your suggestion. Thanks!

pkg/config/creds_test.go
```diff
@@ -10,102 +10,153 @@ import (
...
+			wantFunc: func(t *testing.T, d *secrets.DummyDecryptor, cm *CredentialManager) {
+				d.AssertNotCalled(t, "Decrypt")
+				cachedCreds, cached := cm.getCredsFromCache()
+				assert.True(t, cached)
+				assert.EqualValues(t, Creds{APIKey: "foo", AppKey: "bar"}, cachedCreds)
+			},
+			resetFunc: func(cm *CredentialManager) {
+				os.Unsetenv("DD_API_KEY")
+				os.Unsetenv("DD_APP_KEY")
+				cm.resetCredsCache()
```
clamoriniere, line 43: I think you can remove the `cm.resetCredsCache()` since each test create a new `CredentialManager` instance.

pkg/config/creds.go
```diff
@@ -61,5 +78,28 @@ func getCredentials(decryptor secrets.Decryptor) (Creds, error) {
...
+		return cm.creds, true
+	}
+	return Creds{}, false
+}
+
+func (cm *CredentialManager) resetCredsCache() {
+	cm.credsMutex.Lock()
+	defer cm.credsMutex.Unlock()
+	cm.creds = Creds{}
 }
```
clamoriniere, line 105: can be removed, because you can avoid using it in the unittest

ahmed-mez: yes good point, done 👍

pkg/config/creds.go
```diff
@@ -18,15 +19,29 @@ type Creds struct {
 	AppKey string
 }
 
+// CredentialManager provides the credentials from the operator configuration
+type CredentialManager struct {
+	secretBackend secrets.Decryptor
+	creds         Creds
+	credsMutex    sync.Mutex
```
clamoriniere, line 26: I feel like the `credsMutex` can be remove if:
* `NewCredentialManager()` is also responsible to get the secret, and store it in `creds`
* `GetCredential()` only returns `creds`.

WDYT?

L3n41c: Is the issue error management and retry policy ?
If `NewCredentialManager()` is invoked only once whereas `GetCredential()` is called at several places, what will the behaviour be in case of a transient error in the secret backend ?

pkg/config/creds.go
```diff
@@ -45,10 +60,12 @@ func getCredentials(decryptor secrets.Decryptor) (Creds, error) {
...
+		creds := Creds{APIKey: apiKey, AppKey: appKey}
+		cm.cacheCreds(creds)
+		return creds, nil
 	}
 
-	decrypted, err := decryptor.Decrypt(encrypted)
+	decrypted, err := cm.secretBackend.Decrypt(encrypted)
 	if err != nil {
 		return Creds{}, err
 	}
```
L3n41c, line 71: The code seems structured to support the following case:
* `decrypted, err := cm.secretBackend.Decrypt(encrypted)` returns `err == nil` but
* `apiKey` is in `decrypted`, but not `appKey`.

Can this happen in practice ?
That would mean that one secret could be found, but not the other one.

In such a case, we would cache a decrypted api key but the app key would be cached encrypted and we would never recover automatically.

pkg/controller/utils/datadog/metrics_forwarder.go
L3n41c: In this case, should we put a log to warn the user if one of the two keys is set in the `datadog` CR but is not used because the other one isn’t present ?

It’s just that a user might expect the CR specific settings to override the generic operator ones.

L3n41c (commented):
I only have some interrogations regarding the fact that we try to handle API and APP keys as a whole.

pkg/config/creds.go
clamoriniere: Like we have in Datadog-agent and DCA. the process stop with an error?

pkg/config/creds.go
ahmed-mez: `cm.secretBackend.Decrypt(encrypted)` returns an error if it cannot decrypt all the elements in `encrypted`.
Does this answer to your question?

pkg/config/creds.go
```diff
@@ -18,15 +23,36 @@ type Creds struct {
...
+// NewCredentialManager returns a CredentialManager
+func NewCredentialManager() *CredentialManager {
+	return &CredentialManager{
+		secretBackend: secrets.NewSecretBackend(),
+		creds:         Creds{},
+		decryptorBackoff: wait.Backoff{
+			Steps:    5,
+			Duration: 10 * time.Millisecond,
+			Factor:   5.0,
+			Cap:      20 * time.Second,
```
ahmed-mez, line 43: This is is discussable of course

pkg/config/creds.go
L3n41c: Yes.
If `cm.secretBackend.Decrypt(encrypted)` cannot return some secrets but not all, it’s OK.

ahmed-mez: it's already the case 👍

pkg/config/creds.go
```diff
@@ -45,11 +71,17 @@ func getCredentials(decryptor secrets.Decryptor) (Creds, error) {
...
-	decrypted, err := decryptor.Decrypt(encrypted)
-	if err != nil {
+	decrypted := map[string]string{}
+	var decErr error
+	if err := retry.OnError(cm.decryptorBackoff, secrets.Retriable, func() error {
+		decrypted, decErr = cm.secretBackend.Decrypt(encrypted)
+		return decErr
+	}); err != nil {
 		return Creds{}, err
 	}
```
clamoriniere, line 86: maybe you can reverse the logic here:

```golang
if len(encrypted) > 0 {
   //...
}
creds := Creds{APIKey: apiKey, AppKey: appKey}
cm.cacheCreds(creds)
return creds, err
```

pkg/secrets/types.go
```diff
@@ -12,6 +12,41 @@ import (
...
+	return e.message
+}
+
+// IsRetriable returns wether the error is retriable
+func (e *DecryptorError) IsRetriable() bool {
+	return e.retriable
+}
+
+// NewDecryptorError returns a new DecryptorError
+func NewDecryptorError(message string, retriable bool) *DecryptorError {
```
clamoriniere, line 33: To follow golang 1.13 best practice I think the function should take a `err error` instead of `message string`

something like
```golang
type DecryptorError struct {
   error
   retriable bool
}

func NewDecryptorError(err error, retriable bool) *DecryptorError
```

Official doc: https://golang.org/pkg/errors/
interesting [blogpost](https://blog.tomarrell.com/post/introducing_wrapcheck_linter_for_go)

clamoriniere (commented):
looks good. I added some improvement propositions around error management

pkg/secrets/types.go
ahmed-mez: Thank you! PR updated

clamoriniere (approved):
💯

PR #258, opened 2021-03-10
Fix make deploy due to last-applied annotation size

clamoriniere:
### What does this PR do?

The `DatadogAgent` CRD can be install with `kubectl apply` with the
classic client side merge mechanism because the `last-apply` annotation
used by the `kubectl apply` command is generating a too long annotation
value for the CRD.

Current solution is to use the `server-side` merge management for
the kubectl apply command. Or use the `kubectl create` command.

### Motivation

Fix `make deploy` command

### Additional Notes

N/A
### Describe your test plan

On a kubernetes 1.18+ cluster, install the operator with the command
`make deploy`


vboulineau:
Is it not only useful on the CRD (`install`) then?

celenechang (approved):

clamoriniere:
> Is it not only useful on the CRD (`install`) then?

I can update the PR to only do it for the CRD installation.

PR #263, opened 2021-03-26
[datadogmonitor] support monitor options

celenechang:
### What does this PR do?

Support setting monitor options available in the Datadog Monitor API in a DatadogMonitor spec

### Motivation

Better user experience

### Additional Notes

### Describe your test plan

With a running Operator (with proper credentials), try applying the `examples/datadogmonitor/datadog-monitor-test.yaml` `DatadogMonitor` definition. Also try updating by changing the values and/or adding other options.


clamoriniere (approved):
💯

PR #271, opened 2021-04-07
Cleanup gitlab-ci config

clamoriniere:
### What does this PR do?

Remove gitlab workaround needed with previous gitlab version

### Motivation

Simplify gitlab configuration

### Additional Notes

### Describe your test plan

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -14,7 +14,6 @@ stages:
   - build
   - test
   - image
-  - intermediate
   - release
 
 build:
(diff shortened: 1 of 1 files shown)


CharlyF (approved):

.gitlab-ci.yml
```diff
@@ -75,6 +74,9 @@ publish_public_main:
     - if: '$CI_COMMIT_BRANCH == "main"'
       when: on_success
     - when: never
+  needs:
+    - build_image_arm64
```
vboulineau, line 78: Now that you removed the `intermediate` stage, this is not required anymore. (in all 3 release jobs)

clamoriniere: for my knowledge, if we don't provide a list of `needs` it means all the jobs from the previous "stage" need to succeed, correct?

vboulineau: Yes, which is what we want in this case, everything in `image` needs to be successful to run `publish`

PR #268, opened 2021-04-03
Add Features.LogCollection

clamoriniere:
### What does this PR do?

* Add LogCollection in `spec.features`.
* Add Patch step in DatadogAgent reconcile loop.
* Fixed unittest around orchestrator explorer.

### Additional Notes

Anything else we should know when reviewing?

### Describe your test plan

* Create a `DatadogAgent` with `spec.agent.log.enabled: true`.
  After a reconcile loop, the `DatadogAgent.spec.features.logCollection.enabled:true`
  should be added.
* Create a `DatadogAgent` with `spec.features.logCollection.enabled:true`.
  Log feature should be enabled in the agent.

PR #273, opened 2021-04-16
[DatadogAgent] update configuration options for system probe

celenechang:
### What does this PR do?

Update configuration of System Probe for `DatadogAgent` to parallel its configuration in the agent helm chart.

### Motivation

Better user experience

### Additional Notes

### Describe your test plan

Run the operator in a test cluster

Apply one of the examples in `examples/datadogagent/` (such as datadog-agent-all.yaml) using `kubectl apply -f`, making sure to update the credentials as needed.

Try enabling System Probe with `agent.features.networkMonitoring.Enabled`, `agent.systemProbe.enableOOMKill`, `agent.systemProbe.enableTCPQueueLength` OR `agent.security.runtime.enabled`. Make sure the System Probe container starts properly.

Comment out each of the four parameters above so that they will default to `false`. Make sure the System Probe container does not start.


sarina-dd (approved):
Docs LGTM

api/v1alpha1/datadogagent_types.go
```diff
@@ -518,6 +519,9 @@ type SystemProbeSpec struct {
 	// DebugPort Specify the port to expose pprof and expvar for system-probe agent
 	DebugPort int32 `json:"debugPort,omitempty"`
 
+	// EnableNetworkMonitoring enables Network Performance Monitoring
+	EnableNetworkMonitoring *bool `json:"enableNetworkMonitoring,omitempty"`
```
clamoriniere, line 523: For feature the Idea, is to put them under: `spec.features`  and try to apply always the same pattern

here for example
```yaml
spec:
  features:
     networkMonitoring:
        enable: true
```

the idea is to have a section for each feature. like this if this feature require later to add additional option we can aggregate in the section.

about the option naming. I don't know what should be the best name. We are already using `networkMonitoring` but the product is known as  NPM (NetworkPerformanceMonitoring). So maybe it can be good to as the network team what makes more sense for them.

clamoriniere: as example you have this PR: https://github.com/DataDog/datadog-operator/pull/268

celenechang: i made some updates, i hope it covers the bases but let me know if i missed something. (i checked with networks and they requested use of `networkMonitoring`)

clamoriniere (approved):

PR #275, opened 2021-04-19
Fix operator leader election on vintage k8s cluster

clamoriniere:
### What does this PR do?

Fix Operator leader election on vintage cluster by defaulting, 
the leader election resource to `configmaps` which is supported
on any k8s clusters version.

### Motivation

Fix leader-electio on vintage clusters.

### Additional Notes

### Describe your test plan

Deploy the operator on a kubernetes clustet v1.10.
Check that the Operator is able to instanciate the leader election object.

### DIFF
--- a/main.go
+++ b/main.go
@@ -73,9 +73,11 @@ func (ss *stringSlice) Set(value string) error {
 func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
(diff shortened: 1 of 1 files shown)


CharlyF (approved):
aligned with what we have in other controllers - Very nice.

PR #285, opened 2021-04-29
Remove automatic copy of all labels/annotations from DatadogAgent CRD - now only Datadog tags labels will be copied

vboulineau:
### What does this PR do?

Remove automatic copy of all labels/annotations from DatadogAgent CRD as it triggered copying of unwanted labels (like last-applied) and overwriting others (like kubernetes.io labels)

### Motivation

### Additional Notes

### Describe your test plan

Deploy Agent through Operator, set custom Labels/Annotations on DatadogAgent object, make sure these Labels/Annotations are not present on created DS/Deployment except Datadog tags labels.


vboulineau:
Do we want to keep the `agent.datadoghq.com/agentspechash: <hash>` annotation copied everywhere? I don't we think we need it, but maybe I missed a use case.

clamoriniere (approved):

PR #292, opened 2021-05-03
remove fallback logic for credentials

celenechang:
### What does this PR do?

Do not use (or look for) credentials set in the Operator for DatadogAgent. (Only depend on the DatadogAgent spec.)

### Motivation

The fallback wasn't working properly; removing for now to revisit whether to include this later.

### Additional Notes

### Describe your test plan

Make sure credentials are required in the DatadogAgent spec and that when included, it works properly


clamoriniere (approved):
👍

PR #311, opened 2021-05-21
Fix management of seccomp ConfigMap when not used

vboulineau:
### What does this PR do?

Currently the operator always mount the seccomp volumes, even if we use a seccomp profile from host (e.g. no install performed in init container), while the ConfigMap is not created, creating PODs referencing non-existing ConfigMaps.
Also, the ConfigMap was never deleted if setting was changed from using seccomp profile from ConfigMap to using profile from host.

Took the opportunity to change the way we do diffs in tests.

### Motivation

### Additional Notes

### Describe your test plan

Deploy the Agent with `system-probe` enabled. Then change the DatadogAgent set the `SecCompProfileName` attribute to `unconfined`. The new DaemonSet should start properly and the ConfigMap named `<dda.Name>-system-probe-seccomp` should be deleted.


clamoriniere (approved):

config/crd/bases/v1/datadoghq.com_datadogagents.yaml
```diff
@@ -3630,7 +3630,8 @@ spec:
                         type: object
                       secCompCustomProfileConfigMap:
                         description: SecCompCustomProfileConfigMap specify a pre-existing
-                          ConfigMap containing a custom SecComp profile.
+                          ConfigMap containing a custom SecComp profile. This ConfigMap
+                          must contain a file named system-probe-seccomp.json
```
jtappa, line 3634:
```suggestion
                          must contain a file named system-probe-seccomp.json.
```

config/crd/bases/v1beta1/datadoghq.com_datadogagents.yaml
```diff
@@ -3492,7 +3492,8 @@ spec:
                       type: object
                     secCompCustomProfileConfigMap:
                       description: SecCompCustomProfileConfigMap specify a pre-existing
-                        ConfigMap containing a custom SecComp profile.
+                        ConfigMap containing a custom SecComp profile. This ConfigMap
+                        must contain a file named system-probe-seccomp.json
```
jtappa, line 3496:
```suggestion
                        must contain a file named system-probe-seccomp.json.
```

docs/configuration.md
```diff
@@ -175,7 +175,7 @@ spec:
 | agent.systemProbe.env | The Datadog SystemProbe supports many environment variables. See also: https://docs.datadoghq.com/agent/docker/?tab=standard#environment-variables |
 | agent.systemProbe.resources.limits | Limits describes the maximum amount of compute resources allowed. More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/ |
 | agent.systemProbe.resources.requests | Requests describes the minimum amount of compute resources required. If Requests is omitted for a container, it defaults to Limits if that is explicitly specified, otherwise to an implementation-defined value. More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/ |
-| agent.systemProbe.secCompCustomProfileConfigMap | SecCompCustomProfileConfigMap specify a pre-existing ConfigMap containing a custom SecComp profile. |
+| agent.systemProbe.secCompCustomProfileConfigMap | SecCompCustomProfileConfigMap specify a pre-existing ConfigMap containing a custom SecComp profile. This ConfigMap must contain a file named system-probe-seccomp.json |
```
jtappa, line 178: I'm assuming this is autogenerated and would be fixed with above suggestion?

jtappa (changes requested):

docs/configuration.md
vboulineau: That's correct!

PR #324, opened 2021-06-16
Introducing defaulting logic

CharlyF:
### What does this PR do?

- This PR introduces a new field in the status that surfaces the fields that have been defaulted as part of the defaulting logic of the DatadogAgent (DDA) object.
We do not update the spec of the DDA anymore, thus an update of the operator will directly benefit the deployment of the DDA.
Started using more pointers in order not to end up with empty structures in the object rendered.

- Introduced the FeatureOverride method in order to take a second round of defaulting in case a certain feature is enabled (this is in order to accommodate our update work on simplifying Features)
  
- Fixing the command used in the init-volume and the volumes mounted in case the Runtime security is enabled (or not).

- Repackaged the defaults a bit.

### Motivation

Get ready for GA.

### Additional Notes

- Defaulted objects are put in the defaultOverride structure, initialised ones are just left in the internal DDA (which represents the union of the defaultOverride and the spec).
- One thing I did not fix is the new logic and structure of the Image needs to be updated in the patch section similarly to the Agent.Spec.Logs that moved to Features.Logs, because if the old image name is used we still default the tag value but with the current logic it will not be taken into account.
- We use the method `IsEqualStruct` extensively, which I introduced in order to properly compare the large DDA object that contains nested json and pointers.
- Main features and parts of the DDA have an Enable boolean because we want to support having a DDA without the DS for instance.
- Added proper liveness/readiness probes. Fixed one for the DCA that was overridden if one used the External Metrics Server. We need to merge the liveness response upstream on one endpoint (in our backlog already)

### Describe your test plan

Create DDAs with limited elements in the spec and verify that the DefaultOverride reflects the proper defaults.


controllers/datadogagent/agent.go
```diff
@@ -356,17 +356,12 @@ func buildAgentNetworkPolicy(dda *datadoghqv1alpha1.DatadogAgent, name string) *
...
-			port = *dda.Spec.Agent.Apm.HostPort
-		}
-
 		ingressRules = append(ingressRules, networkingv1.NetworkPolicyIngressRule{
 			Ports: []networkingv1.NetworkPolicyPort{
 				{
 					Port: &intstr.IntOrString{
 						Type:   intstr.Int,
-						IntVal: port,
+						IntVal: *dda.Spec.Agent.Apm.HostPort,
```
clamoriniere, line 364: are we sure now that `datadoghqv1alpha1.DatadogAgent` is never nil

controllers/datadogagent/agent.go
```diff
@@ -405,20 +400,24 @@ func newExtendedDaemonSetFromInstance(logger logr.Logger, dda *datadoghqv1alpha1
 	if err != nil {
 		return nil, "", err
 	}
+	strategy, err := getAgentDeploymentStrategy(dda)
+	if err != nil {
+		return nil, "", err
```
clamoriniere, line 405: maybe you can wrap the error to give more context
```golang
return nil, "", fmt.Errorf("unable to get Deployment strategy when creation new EDS instance, err: %w", err)

File: controllers/datadogagent/agent.go [hunk 3]
```diff
@@ -443,15 +442,19 @@ func newDaemonSetFromInstance(logger logr.Logger, dda *datadoghqv1alpha1.Datadog
 			MatchLabels: template.Labels,
 		}
 	}
+	strategy, err := getAgentDeploymentStrategy(dda)
+	if err != nil {
+		return nil, "", err
```
clamoriniere, line 447: same here about error wrapping

controllers/datadogagent/clusteragent.go
```diff
@@ -178,6 +178,10 @@ func newClusterAgentDeploymentFromInstance(logger logr.Logger, dda *datadoghqv1a
 	}
 
 	annotations := getDefaultAnnotations(dda)
+	for key, val := range dda.Annotations {
```
clamoriniere, line 181: I'm not sure we want to add dda.Annotations to the child deployment

@vboulineau, what is you opinion here? since I know that you have worked on this topic recently.

clamoriniere (commented):
Really like this implementation 👏 
It is great improvement for the Operation UX!

docs/configuration.md
```diff
@@ -71,9 +85,35 @@ spec:
...
+| agent.config.livenessProbe.timeoutSeconds | Number of seconds after which the probe times out. Defaults to 1 second. Minimum value is 1. More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes |
 | agent.config.logLevel | Set logging verbosity, valid log levels are: trace, debug, info, warn, error, critical, and off |
 | agent.config.podAnnotationsAsTags | Provide a mapping of Kubernetes Annotations to Datadog Tags. <KUBERNETES_ANNOTATIONS>: <DATADOG_TAG_KEY> |
 | agent.config.podLabelsAsTags | Provide a mapping of Kubernetes Labels to Datadog Tags. <KUBERNETES_LABEL>: <DATADOG_TAG_KEY> |
+| agent.config.readinessProbe.exec.command | Command is the command line to execute inside the container, the working directory for the command  is root ('/') in the container's filesystem. The command is simply exec'd, it is not run inside a shell, so traditional shell instructions ('|', etc) won't work. To use a shell, you need to explicitly call out to that shell. Exit status of 0 is treated as live/healthy and non-zero is unhealthy. |
```
cswatt, line 104:
```suggestion
| agent.config.readinessProbe.exec.command | Command is the command line to execute inside the container. The working directory for the command is root ('/') in the container's filesystem. The command is simply executed; it is not run inside a shell, so traditional shell instructions ('|', etc.) do not work. To use a shell, you need to explicitly call out to that shell. Exit status of 0 is treated as live/healthy and non-zero is unhealthy. |
```

docs/configuration.md
```diff
@@ -71,9 +85,35 @@ spec:
...
+| agent.config.readinessProbe.failureThreshold | Minimum consecutive failures for the probe to be considered failed after having succeeded. Defaults to 3. Minimum value is 1. |
```
cswatt, line 105:
```suggestion
| agent.config.readinessProbe.failureThreshold | Minimum consecutive failures for the probe to be considered failed after having succeeded. Defaults to `3`. Minimum value is `1`. |
```

docs/configuration.md
```diff
@@ -71,9 +85,35 @@ spec:
...
+| agent.config.readinessProbe.httpGet.host | Host name to connect to, defaults to the pod IP. You probably want to set "Host" in httpHeaders instead. |
```
cswatt, line 106:
```suggestion
| agent.config.readinessProbe.httpGet.host | Host name to connect to, defaults to the pod IP. You may want to set "Host" in `httpHeaders` (see next parameter) instead. |
```

docs/configuration.md
```diff
@@ -71,9 +85,35 @@ spec:
...
+| agent.config.readinessProbe.httpGet.httpHeaders | Custom headers to set in the request. HTTP allows repeated headers. |
+| agent.config.readinessProbe.httpGet.path | Path to access on the HTTP server. |
+| agent.config.readinessProbe.httpGet.port | Name or number of the port to access on the container. Number must be in the range 1 to 65535. Name must be an IANA_SVC_NAME. |
```
cswatt, line 109:
```suggestion
| agent.config.readinessProbe.httpGet.port | Name or number of the port to access on the container. Number must be in the range `1` to `65535`. Name must be an `IANA_SVC_NAME`. |
```

docs/configuration.md
```diff
@@ -71,9 +85,35 @@ spec:
...
+| agent.config.readinessProbe.httpGet.scheme | Scheme to use for connecting to the host. Defaults to HTTP. |
+| agent.config.readinessProbe.initialDelaySeconds | Number of seconds after the container has started before liveness probes are initiated. More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes |
+| agent.config.readinessProbe.periodSeconds | How often (in seconds) to perform the probe. Default to 10 seconds. Minimum value is 1. |
```
cswatt, line 112:
```suggestion
| agent.config.readinessProbe.periodSeconds | How often (in seconds) to perform the probe. Default to `10` seconds. Minimum value is `1`. |
```

docs/configuration.md
```diff
@@ -71,9 +85,35 @@ spec:
...
+| agent.config.readinessProbe.successThreshold | Minimum consecutive successes for the probe to be considered successful after having failed. Defaults to 1. Must be 1 for liveness and startup. Minimum value is 1. |
```
cswatt, line 113:
```suggestion
| agent.config.readinessProbe.successThreshold | Minimum consecutive successes for the probe to be considered successful after having failed. Defaults to `1`. Must be `1` for liveness and startup. Minimum value is `1`. |
```

docs/configuration.md
```diff
@@ -71,9 +85,35 @@ spec:
...
+| agent.config.readinessProbe.tcpSocket.host | Optional: Host name to connect to, defaults to the pod IP. |
+| agent.config.readinessProbe.tcpSocket.port | Number or name of the port to access on the container. Number must be in the range 1 to 65535. Name must be an IANA_SVC_NAME. |
```
cswatt, line 115:
```suggestion
| agent.config.readinessProbe.tcpSocket.port | Number or name of the port to access on the container. Number must be in the range `1` to `65535`. Name must be an `IANA_SVC_NAME`. |
```

docs/configuration.md
```diff
@@ -71,9 +85,35 @@ spec:
...
+| agent.config.readinessProbe.timeoutSeconds | Number of seconds after which the probe times out. Defaults to 1 second. Minimum value is 1. More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes |
```
cswatt, line 116:
```suggestion
| agent.config.readinessProbe.timeoutSeconds | Number of seconds after which the probe times out. Defaults to `1` second. Minimum value is `1`. More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes |
```

docs/configuration.md
```diff
@@ -233,6 +274,7 @@ spec:
 | clusterAgent.config.externalMetrics.port | If specified configures the metricsProvider external metrics service port. |
 | clusterAgent.config.externalMetrics.useDatadogMetrics | Enable usage of DatadogMetrics CRD (allow to scale on arbitrary queries). |
 | clusterAgent.config.externalMetrics.wpaController | Enable informer and controller of the watermark pod autoscaler. NOTE: The WatermarkPodAutoscaler controller needs to be installed. See also: https://github.com/DataDog/watermarkpodautoscaler. |
+| clusterAgent.config.healthPort | HealthPort of the agent container for internal liveness probe. Must be the same as the Liness/Readiness probes. |
```
cswatt, line 277:
```suggestion
| clusterAgent.config.healthPort | HealthPort of the Agent container for internal liveness probe. Must be the same as the Liveness/Readiness probes. |
```

Should it be "liveness" instead of liness?

docs/configuration.md
```diff
@@ -55,6 +68,7 @@ spec:
 | agent.config.dogstatsd.unixDomainSocket.enabled | Enable APM over Unix Domain Socket. See also: https://docs.datadoghq.com/developers/dogstatsd/unix_socket/ |
 | agent.config.dogstatsd.unixDomainSocket.hostFilepath | Define the host APM socket filepath used when APM over Unix Domain Socket is enabled. (default value: /var/run/datadog/statsd.sock). See also: https://docs.datadoghq.com/developers/dogstatsd/unix_socket/ |
 | agent.config.env | The Datadog Agent supports many environment variables. See also: https://docs.datadoghq.com/agent/docker/?tab=standard#environment-variables |
+| agent.config.healthPort | HealthPort of the agent container for internal liveness probe. Must be the same as the Liness/Readiness probes. |
```
cswatt, line 71:
```suggestion
| agent.config.healthPort | HealthPort of the Agent container for internal liveness probe. Must be the same as the Liveness/Readiness probes. |
```

Or is it supposed to be Liness? I had not heard of liness probes

docs/configuration.md
```diff
@@ -267,7 +310,34 @@ spec:
 | clusterChecksRunner.config.args | Args allows the specification of extra args to `Command` parameter |
 | clusterChecksRunner.config.command | Command allows the specification of custom entrypoint for Cluster Checks Runner container |
 | clusterChecksRunner.config.env | The Datadog Agent supports many environment variables. See also: https://docs.datadoghq.com/agent/docker/?tab=standard#environment-variables |
+| clusterChecksRunner.config.healthPort | HealthPort of the agent container for internal liveness probe. Must be the same as the Liness/Readiness probes. |
```
cswatt, line 313:
```suggestion
| clusterChecksRunner.config.healthPort | HealthPort of the Agent container for internal liveness probe. Must be the same as the Liveness/Readiness probes. |
```
again i'm not 100% sure this is supposed to be liveness vs. liness

docs/configuration.md
```diff
@@ -267,7 +310,34 @@ spec:
...
+| clusterChecksRunner.config.livenessProbe.exec.command | Command is the command line to execute inside the container, the working directory for the command  is root ('/') in the container's filesystem. The command is simply exec'd, it is not run inside a shell, so traditional shell instructions ('|', etc) won't work. To use a shell, you need to explicitly call out to that shell. Exit status of 0 is treated as live/healthy and non-zero is unhealthy. |
```
cswatt, line 314:
```suggestion
| clusterChecksRunner.config.livenessProbe.exec.command | Command is the command line to execute inside the container. The working directory for the command is root ('/') in the container's filesystem. The command is simply exec'd; it is not run inside a shell, so traditional shell instructions ('|', etc) do not work. To use a shell, you need to explicitly call out to that shell. Exit status of 0 is treated as live/healthy and non-zero is unhealthy. |
```

docs/configuration.md
```diff
@@ -267,7 +310,34 @@ spec:
...
+| clusterChecksRunner.config.livenessProbe.failureThreshold | Minimum consecutive failures for the probe to be considered failed after having succeeded. Defaults to 3. Minimum value is 1. |
```
cswatt, line 315:
```suggestion
| clusterChecksRunner.config.livenessProbe.failureThreshold | Minimum consecutive failures for the probe to be considered failed after having succeeded. Defaults to `3`. Minimum value is `1`. |
```

docs/configuration.md
```diff
@@ -267,7 +310,34 @@ spec:
...
+| clusterChecksRunner.config.livenessProbe.httpGet.host | Host name to connect to, defaults to the pod IP. You probably want to set "Host" in httpHeaders instead. |
```
cswatt, line 316:
```suggestion
| clusterChecksRunner.config.livenessProbe.httpGet.host | Host name to connect to, defaults to the pod IP. You may want to set "Host" in `httpHeaders` instead. |
```

docs/configuration.md
```diff
@@ -267,7 +310,34 @@ spec:
...
+| clusterChecksRunner.config.livenessProbe.httpGet.httpHeaders | Custom headers to set in the request. HTTP allows repeated headers. |
+| clusterChecksRunner.config.livenessProbe.httpGet.path | Path to access on the HTTP server. |
+| clusterChecksRunner.config.livenessProbe.httpGet.port | Name or number of the port to access on the container. Number must be in the range 1 to 65535. Name must be an IANA_SVC_NAME. |
```
cswatt, line 319:
```suggestion
| clusterChecksRunner.config.livenessProbe.httpGet.port | Name or number of the port to access on the container. Number must be in the range `1` to `65535`. Name must be an `IANA_SVC_NAME`. |
```

docs/configuration.md
```diff
@@ -267,7 +310,34 @@ spec:
...
+| clusterChecksRunner.config.livenessProbe.httpGet.scheme | Scheme to use for connecting to the host. Defaults to HTTP. |
+| clusterChecksRunner.config.livenessProbe.initialDelaySeconds | Number of seconds after the container has started before liveness probes are initiated. More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes |
+| clusterChecksRunner.config.livenessProbe.periodSeconds | How often (in seconds) to perform the probe. Default to 10 seconds. Minimum value is 1. |
```
cswatt, line 322:
```suggestion
| clusterChecksRunner.config.livenessProbe.periodSeconds | How often (in seconds) to perform the probe. Defaults to `10` seconds. Minimum value is `1`. |
```

docs/configuration.md
```diff
@@ -267,7 +310,34 @@ spec:
...
+| clusterChecksRunner.config.livenessProbe.successThreshold | Minimum consecutive successes for the probe to be considered successful after having failed. Defaults to 1. Must be 1 for liveness and startup. Minimum value is 1. |
```
cswatt, line 323:
```suggestion
| clusterChecksRunner.config.livenessProbe.successThreshold | Minimum consecutive successes for the probe to be considered successful after having failed. Defaults to `1`. Must be `1` for liveness and startup. Minimum value is `1`. |
```

docs/configuration.md
```diff
@@ -267,7 +310,34 @@ spec:
...
+| clusterChecksRunner.config.livenessProbe.tcpSocket.host | Optional: Host name to connect to, defaults to the pod IP. |
+| clusterChecksRunner.config.livenessProbe.tcpSocket.port | Number or name of the port to access on the container. Number must be in the range 1 to 65535. Name must be an IANA_SVC_NAME. |
```
cswatt, line 325:
```suggestion
| clusterChecksRunner.config.livenessProbe.tcpSocket.port | Number or name of the port to access on the container. Number must be in the range `1` to `65535`. Name must be an `IANA_SVC_NAME`. |
```

docs/configuration.md
```diff
@@ -267,7 +310,34 @@ spec:
...
+| clusterChecksRunner.config.livenessProbe.timeoutSeconds | Number of seconds after which the probe times out. Defaults to 1 second. Minimum value is 1. More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes |
```
cswatt, line 326:
```suggestion
| clusterChecksRunner.config.livenessProbe.timeoutSeconds | Number of seconds after which the probe times out. Defaults to `1` second. Minimum value is `1`. More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes |
```

docs/configuration.md
```diff
@@ -267,7 +310,34 @@ spec:
...
 | clusterChecksRunner.config.logLevel | Set logging verbosity, valid log levels are: trace, debug, info, warn, error, critical, and off |
+| clusterChecksRunner.config.readinessProbe.exec.command | Command is the command line to execute inside the container, the working directory for the command  is root ('/') in the container's filesystem. The command is simply exec'd, it is not run inside a shell, so traditional shell instructions ('|', etc) won't work. To use a shell, you need to explicitly call out to that shell. Exit status of 0 is treated as live/healthy and non-zero is unhealthy. |
```
cswatt, line 328:
```suggestion
| clusterChecksRunner.config.readinessProbe.exec.command | Command is the command line to execute inside the container. The working directory for the command is root ('/') in the container's filesystem. The command is simply exec'd; it is not run inside a shell, so traditional shell instructions ('|', etc.) do not work. To use a shell, you need to explicitly call out to that shell. Exit status of 0 is treated as live/healthy and non-zero is unhealthy. |
```

docs/configuration.md
```diff
@@ -267,7 +310,34 @@ spec:
...
+| clusterChecksRunner.config.readinessProbe.failureThreshold | Minimum consecutive failures for the probe to be considered failed after having succeeded. Defaults to 3. Minimum value is 1. |
```
cswatt, line 329:
```suggestion
| clusterChecksRunner.config.readinessProbe.failureThreshold | Minimum consecutive failures for the probe to be considered failed after having succeeded. Defaults to `3`. Minimum value is `1`. |
```

docs/configuration.md
```diff
@@ -267,7 +310,34 @@ spec:
...
+| clusterChecksRunner.config.readinessProbe.httpGet.host | Host name to connect to, defaults to the pod IP. You probably want to set "Host" in httpHeaders instead. |
```
cswatt, line 330:
```suggestion
| clusterChecksRunner.config.readinessProbe.httpGet.host | Host name to connect to, defaults to the pod IP. You may want to set "Host" in `httpHeaders` instead. |
```

docs/configuration.md
```diff
@@ -267,7 +310,34 @@ spec:
...
+| clusterChecksRunner.config.readinessProbe.httpGet.httpHeaders | Custom headers to set in the request. HTTP allows repeated headers. |
+| clusterChecksRunner.config.readinessProbe.httpGet.path | Path to access on the HTTP server. |
+| clusterChecksRunner.config.readinessProbe.httpGet.port | Name or number of the port to access on the container. Number must be in the range 1 to 65535. Name must be an IANA_SVC_NAME. |
```
cswatt, line 333:
```suggestion
| clusterChecksRunner.config.readinessProbe.httpGet.port | Name or number of the port to access on the container. Number must be in the range `1` to `65535`. Name must be an `IANA_SVC_NAME`. |
```

docs/configuration.md
```diff
@@ -267,7 +310,34 @@ spec:
...
+| clusterChecksRunner.config.readinessProbe.httpGet.scheme | Scheme to use for connecting to the host. Defaults to HTTP. |
+| clusterChecksRunner.config.readinessProbe.initialDelaySeconds | Number of seconds after the container has started before liveness probes are initiated. More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes |
+| clusterChecksRunner.config.readinessProbe.periodSeconds | How often (in seconds) to perform the probe. Default to 10 seconds. Minimum value is 1. |
```
cswatt, line 336:
```suggestion
| clusterChecksRunner.config.readinessProbe.periodSeconds | How often (in seconds) to perform the probe. Default to `10` seconds. Minimum value is `1`. |
```

docs/configuration.md
```diff
@@ -267,7 +310,34 @@ spec:
...
+| clusterChecksRunner.config.readinessProbe.successThreshold | Minimum consecutive successes for the probe to be considered successful after having failed. Defaults to 1. Must be 1 for liveness and startup. Minimum value is 1. |
```
cswatt, line 337:
```suggestion
| clusterChecksRunner.config.readinessProbe.successThreshold | Minimum consecutive successes for the probe to be considered successful after having failed. Defaults to `1`. Must be `1` for liveness and startup. Minimum value is `1`. |
```

docs/configuration.md
```diff
@@ -267,7 +310,34 @@ spec:
...
+| clusterChecksRunner.config.readinessProbe.tcpSocket.host | Optional: Host name to connect to, defaults to the pod IP. |
+| clusterChecksRunner.config.readinessProbe.tcpSocket.port | Number or name of the port to access on the container. Number must be in the range 1 to 65535. Name must be an IANA_SVC_NAME. |
```
cswatt, line 339:
```suggestion
| clusterChecksRunner.config.readinessProbe.tcpSocket.port | Number or name of the port to access on the container. Number must be in the range `1` to `65535`. Name must be an `IANA_SVC_NAME`. |
```

docs/configuration.md
```diff
@@ -267,7 +310,34 @@ spec:
...
+| clusterChecksRunner.config.readinessProbe.timeoutSeconds | Number of seconds after which the probe times out. Defaults to 1 second. Minimum value is 1. More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes |
```
cswatt, line 340:
```suggestion
| clusterChecksRunner.config.readinessProbe.timeoutSeconds | Number of seconds after which the probe times out. Defaults to `1` second. Minimum value is `1`. More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes |
```

cswatt (changes requested):
just small docs changes

api/v1alpha1/utils.go
```diff
@@ -42,3 +46,20 @@ func BoolToString(b *bool) string {
 
 	return "false"
 }
+
+// IsEqualStruct is a util fonction that returns whether 2 structures are the same
+// We compare the marchaled results to avoid traversing all fields and be agnostic of the struct.
+func IsEqualStruct(in interface{}, cmp interface{}) bool {
```
vboulineau, line 52: Looking at the PR, it looks like `IsEqualStruct` is only used to compare `userInput` and `emptyStruct`, like `v1alpha1.IsEqualStruct(dd.Spec.Agent, v1alpha1.DatadogAgentSpecAgentSpec{})`.
Doing JSON Marshalling is costly in terms of CPU and I wonder if we could do better here.

api/v1alpha1/datadogagent_types.go
```diff
@@ -42,15 +43,15 @@ type DatadogAgentSpec struct {
 	// The desired state of the Agent as an extended daemonset.
 	// Contains the Node Agent configuration and deployment strategy.
 	// +optional
-	Agent *DatadogAgentSpecAgentSpec `json:"agent,omitempty"`
+	Agent DatadogAgentSpecAgentSpec `json:"agent,omitempty"`
```
vboulineau, line 46: I understand the general move to non-pointer to pointer, but why moving these ones from pointer to non-pointer?

controllers/datadogagent/agent_test.go
```diff
@@ -897,6 +893,26 @@ func defaultSystemProbeEnvVars() []corev1.EnvVar {
 			Name:  datadoghqv1alpha1.DDSystemProbeNPMEnabled,
 			Value: "false",
 		},
+		{
+			Name:  datadoghqv1alpha1.DDSystemProbeConntrackEnabled,
```
vboulineau, line 897: It means it's no more possible to configure Agent through `datadog.yaml` even if booleans are not set in `DatadogAgent` CRD.
It's not a problem specific to this field, it has been there for a while, but with this PR we should take the time to clarify what we want to support there.

controllers/datadogagent/clusteragent.go
vboulineau: We should not do it due to: https://github.com/DataDog/datadog-operator/pull/285

controllers/datadogagent/common_configmap.go
```diff
@@ -68,13 +67,24 @@ func (r *Reconciler) updateIfNeededConfigMap(dda *datadoghqv1alpha1.DatadogAgent
 	// Copy possibly changed fields
 	updateCM := oldConfigMap.DeepCopy()
 	updateCM.Data = newConfigMap.Data
+	if len(updateCM.Annotations) == 0 {
```
vboulineau, line 70: It'd be clearer to check for `nil` as an empty, but initialized, map also has length == 0.

vboulineau (commented):
Great work, some comments. I think newer Operator with older CRDs will `panic` (the opposite is also probably true), so we should state that somewhere in the `README`.

controllers/datadogagent/agent.go
CharlyF: `isAPMEnabled` should take care of that

controllers/datadogagent/clusteragent.go
CharlyF: okay thanks and good catch. I initially added that because it was panic'ing, but as I rebased I included your change and missed the rational why we removed them 👍

api/v1alpha1/datadogagent_default.go
```diff
@@ -55,910 +67,1120 @@ const (
...
-		}
-		if !IsDefaultedImageConfig(&ad.Spec.Agent.Image) {
-			return false
-		}
+// DefaultDatadogAgent defaults the DatadogAgent
+func DefaultDatadogAgent(dda *DatadogAgent) *DatadogAgentStatus {
+	// instOverrideStatus contains all the defaults from the runtime.
+	// It is published in the status of the DatadogAgent
+	dso := &DatadogAgentStatus{
+		DefaultOverride: &DatadogAgentSpec{},
```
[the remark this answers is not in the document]
CharlyF: I still see this as overriding, I chose `DefaultOverride` to highlight that the spec was defaulted for empty values that are later on required and therefore overridden. The goal was to only display the diff of what the user set and what was "defaulted" there.

api/v1alpha1/datadogagent_default.go
```diff
@@ -55,910 +67,1120 @@ const (
...
+	// Readiness probe default config
+	defaultReadinessProbeInitialDelaySeconds int32 = 15
+	defaultReadinessProbePeriodSeconds       int32 = 15
+	defaultReadinessProbeTimeoutSeconds      int32 = 5
+	defaultReadinessProbeSuccessThreshold    int32 = 1
+	defaultReadinessProbeFailureThreshold    int32 = 6
+	defaultReadinessProbeHTTPPath                  = "/ready"
 )
 
 var defaultImagePullPolicy = corev1.PullIfNotPresent
```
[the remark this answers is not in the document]
CharlyF: Interesting, I intentionally moved all to the default file so it was all in one place and also avoided imports (maybe I hit a circular one too?). Anyways, I did not consider the vendoring aspect, that's a good point.
This does represent some work to move things around etc, I am in favour of merging and then creating a backlog task to address this refactoring, but let me know if I am missing something.

api/v1alpha1/utils.go
CharlyF: I agree, it's not ideal... Because of the intricacy of the whole structure (with pointers of other structures etc), I decided to go the marchalling route to make sure that I would compare apple to apple. 
I did look into a few alternatives with build tags, interfaces etc but that was the best I could find.
I did isolate the function on purpose though, in case we figure out something better.

api/v1alpha1/datadogagent_types.go
CharlyF: Great point - There are multiple reasons.
The goal was to have the high level feature fields (Feature being one on them, I hope it's not confusing) as non ptr structs so I could pass them down when building the defaulting instead of passing by reference the same struct and trying to get the diff of what was overridden (which would not work if the pointer is nil initially as we would work on a local copy).
Secondly, it helps distinguish if a section is disabled entirely (which will help for the next task of dealing with the Feature section to enable a lot of things depending on a small feature flag). So if the section is empty, we use the `Enabled` flag as false (and so is the internal structure) and can end up with an operator which does not have the agent for instance (which should be the case but is not possible at the moment IIRC).

controllers/datadogagent/agent_test.go
CharlyF: Sorry, I don't get this one. Are you talking about the config file of the agent? Do you mean that we don't mirror values to this file?
Happy to chat about how you think we could approach this and use maybe this change to fix it.

controllers/datadogagent/common_configmap.go
CharlyF: good point - I am also updating the label a few lines below.

CharlyF:
@cswatt I actually tried to use your suggestions but it conflicts with the code gen tool that we use. 
I am going to create a card so that we can see what we can do and I will share the patch there so your recs are not lost. Thanks again!

CharlyF:
Tried to use the operator 0.6.0 with the new CRD and also used the new controller with the old CRD and it went well.
Are we good to go @vboulineau @clamoriniere ?

api/v1alpha1/utils.go
vboulineau: I would expect https://github.com/kubernetes/apimachinery/blob/master/pkg/api/equality/semantic.go to work in this scenario. We could extend the `Semantic` var with more types if not.

controllers/datadogagent/agent_test.go
vboulineau: Any ENV variable that is set by default means that having a configuration in `datadog.yaml` (and nothing set in `DatadogAgent`) is not going to work, the value from `datadog.yaml` being ignored.
As this PR handles defaulting, I think it's a good place to clarify what we want to do for these cases.

api/v1alpha1/utils.go
CharlyF: I looked into it a bit - It would probably work indeed. Though as it uses reflects under the hood I would be curious to benchmark the change. I will make a card so we can evaluate changing this bit of logic.

controllers/datadogagent/agent_test.go
CharlyF: The goal of this change was to surface defaulted values in the status and stop mutating the spec of the DDA. 
But I agree with you, this problem should be tackled as part of the boarder defaulting initiative.
For the sake of unblocking iterations towards the GA, I am creating a task in our backlog to look into that. 
I still have to work on the `Features`, so all can be done in parallel once we merge this (which is a blocker for the rest).

vboulineau (approved):

PR #329, opened 2021-06-29
add host /etc/*-release files to agent container

celenechang:
### What does this PR do?

Mounts host /etc/*-release files to agent container

### Motivation

Feature request

### Additional Notes

### Describe your test plan

Exec onto agent container and run `ls /host/etc/` and/or `cat /host/etc/*-release` to make sure files are mounted correctly

### DIFF
--- a/api/v1alpha1/const.go
+++ b/api/v1alpha1/const.go
@@ -122,9 +122,9 @@ const (
 	DDExternalMetricsProviderAppKey              = "DD_EXTERNAL_METRICS_PROVIDER_APP_KEY"
 	DDAuthTokenFilePath                          = "DD_AUTH_TOKEN_FILE_PATH"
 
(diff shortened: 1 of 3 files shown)


vboulineau:
Is it necessary to always mount these files? They may not exist, could we handle it with https://github.com/DataDog/datadog-operator/blob/main/api/v1alpha1/datadogagent_types.go#L783-L789 only?

ahmed-mez (commented):
I kind of agree with Vincent, but in case we're adding these volumes let's also make the same change in the public helm chart.

celenechang:
Ah, yea we could do that. I figured since the `/etc/os-release` volume mount is already defined in the code I would add in `/etc/lsb-release`, but you're right that the latter is less likely to be present than the former.

There's still an open discussion around how to handle this on the public charts, so I'll just make the internal change needed for now

PR #332, opened 2021-07-08
Do not default CLC and DCA replica fields

ahmed-mez:
### What does this PR do?

Do not default CLC and DCA replicas

### Motivation

Allow using a WPA or a HPA to autoscale the DCA and CLC deployments

### Describe your test plan

Do not set the `replicas` field in the CR + deploy a WPA/HPA targeting the deployment:
- make sure the operator doesn't change deployment's spec and doesn't log any errors.
- make sure the HPA/WPA is able to autoscale the deployment
- apply some config changes to the runners in the DatadogAgent CR, make sure the deployment is updated accordingly but the replicas field managed by the HPA/WPA stays stable `kg deploy datadog-cluster-checks-runner -ojson | jq .spec.replicas`


sarina-dd (approved):
👍 docs

CharlyF (approved):

PR #342, opened 2021-07-28
Fix runtime security config mount

clamoriniere:
### What does this PR do?

Fix issue with "security-runtime" policyDir configuration. which was removing the default configuration. Now the configuration provided by `policyDir` are added to the `/etc/datadog-agent/runtime-security.d` folder.

### Motivation

give more flexibility to the runtime policies configuration.

### Additional Notes

N/A

### Describe your test plan

try to deploy the datadogagent with runtime-security enable. with 2 different configs

1. no custom config => the `default.policy` configuration should be present in `/etc/datadog-agent/runtime-security.d`

```yaml  
    security:

      runtime:
        enabled: true
```

2. provide a configMapName + items. => `default.policy` configuration and the `custom.policy` files should be present 

```yaml
    security:
      runtime:
        enabled: true
        policiesDir:
          configMapName: "runtime-config"
```

you can use the following configmap:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: runtime-config
data:
  custom.policy: |
    version: 1.0.0
    macros:
    rules:
```

### DIFF
--- a/controllers/datadogagent/agent_test.go
+++ b/controllers/datadogagent/agent_test.go
@@ -1635,7 +1635,7 @@ func runtimeSecurityAgentPodSpec(extraEnv map[string]string, extraDir string) co
 
 		volumeMountsBuilder.Add(&corev1.VolumeMount{
 			Name:      datadoghqv1alpha1.SecurityAgentRuntimePoliciesDirVolumeName,
-			MountPath: "/etc/datadog-agent/runtime-security.d",
+			MountPath: "/opt/datadog-agent/runtime-security.d",
 		})
 		volumeMountsBuilder.Add(&corev1.VolumeMount{
 			Name:      datadoghqv1alpha1.SecurityAgentRuntimeCustomPoliciesVolumeName,
@@ -3443,6 +3443,14 @@ func Test_newExtendedDaemonSetFromInstance_SecurityAgent_Runtime(t *testing.T) {
 			ReadOnly:  true,
 		},
 	}...)
+	securityAgentPodSpec.Containers[2].VolumeMounts = append(securityAgentPodSpec.Containers[2].VolumeMounts, []corev1.VolumeMount{
+		{
+			Name:      datadoghqv1alpha1.SecurityAgentRuntimePoliciesDirVolumeName,
+			MountPath: "/etc/datadog-agent/runtime-security.d",
+			ReadOnly:  true,
+		},
+	}...)
+
 	securityAgentPodSpec.InitContainers[1].VolumeMounts = append(securityAgentPodSpec.Containers[0].VolumeMounts, []corev1.VolumeMount{
 		{
 			Name:      datadoghqv1alpha1.SystemProbeSocketVolumeName,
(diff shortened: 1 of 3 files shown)


controllers/datadogagent/agent_test.go
```diff
@@ -3443,6 +3443,14 @@ func Test_newExtendedDaemonSetFromInstance_SecurityAgent_Runtime(t *testing.T) {
 			ReadOnly:  true,
 		},
 	}...)
+	securityAgentPodSpec.Containers[2].VolumeMounts = append(securityAgentPodSpec.Containers[2].VolumeMounts, []corev1.VolumeMount{
+		{
+			Name:      datadoghqv1alpha1.SecurityAgentRuntimePoliciesDirVolumeName,
+			MountPath: "/etc/datadog-agent/runtime-security.d",
```
CharlyF, line 3449: is it /opt ?
https://github.com/DataDog/datadog-operator/pull/342/files#diff-1e55b7aa7422a4461a33b141f491043abd3d73bf64215734d90850c38655e289R1638

controllers/datadogagent/utils_test.go
```diff
@@ -195,14 +195,15 @@ func Test_getVolumeMountsForSecurityAgent(t *testing.T) {
...
-			name: "compliance volumeMounts",
+			name: "runtime volumeMounts",
 			dda:  testutils.NewDatadogAgent("foo", "bar", "datadog/agent:7", &testutils.NewDatadogAgentOptions{SecuritySpec: securityRuntime}),
 			want: []v1.VolumeMount{
 				{Name: "logdatadog", ReadOnly: false, MountPath: "/var/log/datadog"},
 				{Name: "datadog-agent-auth", ReadOnly: true, MountPath: "/etc/datadog-agent/auth"},
 				{Name: "dsdsocket", ReadOnly: true, MountPath: "/var/run/datadog/statsd"},
 				{Name: "config", ReadOnly: false, MountPath: "/etc/datadog-agent"},
 				{Name: "hostroot", ReadOnly: true, MountPath: "/host/root"},
+				{Name: "runtimepoliciesdir", ReadOnly: true, MountPath: "/etc/datadog-agent/runtime-security.d"},
```
CharlyF, line 206: /opt ?

controllers/datadogagent/agent_test.go
clamoriniere: no here it is `/etc/` and `SecurityAgentRuntimePoliciesDirVolumeName` correspond to the emptyDir volume with aggregation of default files + custom file. (the magic appends inside an init container )

controllers/datadogagent/utils_test.go
clamoriniere: same here, it is on purpose that we mount in `/etc/` the emptyDir.

controllers/datadogagent/utils.go
```diff
@@ -1761,6 +1745,14 @@ func getVolumeMountsForSecurityAgent(dda *datadoghqv1alpha1.DatadogAgent) []core
 		}...)
 	}
 
+	if runtimeEnabled {
```
lebauce, line 1748: /etc/datadog-agent/runtime-security.d is not used by security-agent, only by system-probe

lebauce (approved):

CharlyF (approved):

controllers/datadogagent/utils.go
clamoriniere: I have added here, if later on it is the case.

PR #349, opened 2021-08-05
Fix cluster-agent defaulting to enable it by default

clamoriniere:
### What does this PR do?

Fix the default of the cluster-agent to all be defaulted to `enabled:true`.

### Motivation

In a specific when the section "spec.clusterAgent" is not provided
in the DatadogAgent instance, the cluster-agent is not defaulted to
`spec.clusterAgent.enabled:true`.

### Additional Notes

N/A

### Describe your test plan

Deploy the example: `datadog-agent-with-credential-secret.yaml`

The cluster-agent should be deployed. Defaulting value can be also
verified by checking the value of `dataddog.status.defaultOverride.clusterAgent.enabled`

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -29,7 +29,7 @@ spec:
         - --enable-leader-election
         - --pprof
         image: controller:latest
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
         name: manager
         env:
         - name: POD_NAME
--- a/controllers/datadogagent_controller_test.go
+++ b/controllers/datadogagent_controller_test.go
@@ -123,6 +123,7 @@ var _ = Describe("DatadogAgent Controller", func() {
 		It("It should create DaemonSet", func() {
 			options := &testutils.NewDatadogAgentOptions{
 				UseEDS:                       false,
+				ClusterAgentDisabled:         true,
 				APIKey:                       "xnfdsjgdjcxlg42rqmzxzvdsgjdfklg",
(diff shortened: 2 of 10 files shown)


config/manager/manager.yaml
```diff
@@ -29,7 +29,7 @@ spec:
         - --enable-leader-election
         - --pprof
         image: controller:latest
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
```
clamoriniere, line 32: ease local testing

controllers/datadogagent_controller_test.go
```diff
@@ -123,6 +123,7 @@ var _ = Describe("DatadogAgent Controller", func() {
 		It("It should create DaemonSet", func() {
 			options := &testutils.NewDatadogAgentOptions{
 				UseEDS:                       false,
+				ClusterAgentDisabled:         true,
```
clamoriniere, line 126: to not modify the current test suite

controllers/testutils/new.go
```diff
@@ -17,7 +17,7 @@ import (
 type NewDatadogAgentOptions struct {
 	ExtraLabels                  map[string]string
 	ExtraAnnotations             map[string]string
-	ClusterAgentEnabled          bool
+	ClusterAgentDisabled         bool
```
clamoriniere, line 20: convert Enabled to Disabled, because default is now "enabled".

ahmed-mez (approved):

PR #351, opened 2021-08-09
Improve defaulting when spec.Agent|ClusterAgent empty

clamoriniere:
### What does this PR do?

* Improve agent defaulting by enabling it if `spec.Agent` is empty.
* Improve ClusterAgent defaulting by enabling it if `spec.ClusterAgent`
  is empty.
* Fix credentials verification function.

### Motivation

DatadogAgent should valide even if only "spec.Credentials" and `spec.Features` are
provided.

### Additional Notes

N/A

### Describe your test plan

Deploy the example: `examples/datadogagent/datadog-agent-with-credential-secret.yaml`

Expected results:
* ClusterAgent should be deployed
* Agent should be deployed

### DIFF
--- a/api/v1alpha1/datadogagent_default.go
+++ b/api/v1alpha1/datadogagent_default.go
@@ -32,8 +32,12 @@ const (
 	defaultUseDogStatsDSocketVolume         bool   = false
 	defaultHostDogstatsdSocketName          string = "statsd.sock"
 	defaultHostDogstatsdSocketPath          string = "/var/run/datadog"
+	defaultAgentEnabled                     bool   = true
+	defaultClusterAgentEnabled              bool   = true
+	defaultClusterChecksRunnerEnabled       bool   = false
 	defaultApmEnabled                       bool   = false
 	defaultApmHostPort                      int32  = 8126
+	defaultExternalMetricsEnabled           bool   = false
 	defaultSystemProbeEnabled               bool   = false
 	defaultSystemProbeOOMKillEnabled        bool   = false
 	defaultSystemProbeTCPQueueLengthEnabled bool   = false
@@ -163,7 +167,7 @@ func FeatureOverride(dda *DatadogAgentSpec, dso *DatadogAgentSpec) {
 func DefaultDatadogAgentSpecAgent(agent *DatadogAgentSpecAgentSpec) *DatadogAgentSpecAgentSpec {
 	// If the Agent is not specified in the spec, disable it.
 	if IsEqualStruct(*agent, DatadogAgentSpecAgentSpec{}) {
-		agent.Enabled = NewBoolPointer(true)
+		agent.Enabled = NewBoolPointer(defaultAgentEnabled)
 
 		if !BoolValue(agent.Enabled) {
 			return agent
@@ -172,7 +176,7 @@ func DefaultDatadogAgentSpecAgent(agent *DatadogAgentSpecAgentSpec) *DatadogAgen
 
 	agentOverride := &DatadogAgentSpecAgentSpec{}
 	if agent.Enabled == nil {
-		agent.Enabled = NewBoolPointer(true)
+		agent.Enabled = NewBoolPointer(defaultAgentEnabled)
 		agentOverride.Enabled = agent.Enabled
 	}
 
@@ -929,7 +933,7 @@ func DefaultDatadogFeatureNetworkMonitoring(ft *DatadogFeatures) *NetworkMonitor
 // return the defaulted DatadogAgentSpecClusterAgentSpec to update the status
 func DefaultDatadogAgentSpecClusterAgent(clusterAgent *DatadogAgentSpecClusterAgentSpec) *DatadogAgentSpecClusterAgentSpec {
 	if IsEqualStruct(*clusterAgent, DatadogAgentSpecClusterAgentSpec{}) {
-		clusterAgent.Enabled = NewBoolPointer(true)
+		clusterAgent.Enabled = NewBoolPointer(defaultClusterAgentEnabled)
 
 		if !BoolValue(clusterAgent.Enabled) {
 			return clusterAgent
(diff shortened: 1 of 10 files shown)


ahmed-mez (approved):

api/v1alpha1/datadogagent_default.go
```diff
@@ -902,7 +909,10 @@ func DefaultDatadogFeaturePrometheusScrape(ft *DatadogFeatures) *PrometheusScrap
 func DefaultDatadogFeatureNetworkMonitoring(ft *DatadogFeatures) *NetworkMonitoringConfig {
 	if ft.NetworkMonitoring == nil {
 		ft.NetworkMonitoring = &NetworkMonitoringConfig{Enabled: NewBoolPointer(false)}
-		return ft.NetworkMonitoring
+
+		if !BoolValue(ft.NetworkMonitoring.Enabled) {
```
CharlyF, line 913: This seems like it'll always be the case, why adding it?

api/v1alpha1/datadogagent_default.go
```diff
@@ -920,15 +930,23 @@ func DefaultDatadogFeatureNetworkMonitoring(ft *DatadogFeatures) *NetworkMonitor
 func DefaultDatadogAgentSpecClusterAgent(clusterAgent *DatadogAgentSpecClusterAgentSpec) *DatadogAgentSpecClusterAgentSpec {
 	if IsEqualStruct(*clusterAgent, DatadogAgentSpecClusterAgentSpec{}) {
 		clusterAgent.Enabled = NewBoolPointer(true)
-		return clusterAgent
+
+		if !BoolValue(clusterAgent.Enabled) {
```
CharlyF, line 934: This will never be the case I believe

api/v1alpha1/datadogagent_default.go
```diff
@@ -999,7 +1017,10 @@ func DefaultDatadogAgentSpecClusterAgentConfig(dca *DatadogAgentSpecClusterAgent
 func DefaultExternalMetrics(conf *ClusterAgentConfig) *ExternalMetricsConfig {
 	if conf.ExternalMetrics == nil {
 		conf.ExternalMetrics = &ExternalMetricsConfig{Enabled: NewBoolPointer(false)}
-		return conf.ExternalMetrics
+
+		if !BoolValue(conf.ExternalMetrics.Enabled) {
```
CharlyF, line 1021: same here

api/v1alpha1/datadogagent_default.go
```diff
@@ -1019,7 +1040,10 @@ func DefaultExternalMetrics(conf *ClusterAgentConfig) *ExternalMetricsConfig {
 func DefaultAdmissionController(conf *ClusterAgentConfig) *AdmissionControllerConfig {
 	if conf.AdmissionController == nil {
 		conf.AdmissionController = &AdmissionControllerConfig{Enabled: NewBoolPointer(defaultAdmissionControllerEnabled)}
-		return conf.AdmissionController
+
+		if !BoolValue(conf.AdmissionController.Enabled) {
```
CharlyF, line 1044: same here

api/v1alpha1/datadogagent_default.go
```diff
@@ -1075,7 +1099,10 @@ func DefaultDatadogClusterAgentImage(dca *DatadogAgentSpecClusterAgentSpec, name
 func DefaultDatadogAgentSpecClusterChecksRunner(clusterChecksRunner *DatadogAgentSpecClusterChecksRunnerSpec) *DatadogAgentSpecClusterChecksRunnerSpec {
 	if IsEqualStruct(clusterChecksRunner, DatadogAgentSpecClusterChecksRunnerSpec{}) {
 		clusterChecksRunner.Enabled = NewBoolPointer(false)
-		return clusterChecksRunner
+
+		if !BoolValue(clusterChecksRunner.Enabled) {
```
CharlyF, line 1103: same here

controllers/datadogagent/clusteragent.go
```diff
@@ -763,7 +766,10 @@ func getClusterAgentMetricsProviderPort(config datadoghqv1alpha1.ClusterAgentCon
 }
 
 func getAdmissionControllerServiceName(dda *datadoghqv1alpha1.DatadogAgent) string {
-	if isClusterAgentEnabled(dda.Spec.ClusterAgent) && dda.Spec.ClusterAgent.Config.AdmissionController != nil && dda.Spec.ClusterAgent.Config.AdmissionController.ServiceName != nil {
+	if isClusterAgentEnabled(dda.Spec.ClusterAgent) &&
+		dda.Spec.ClusterAgent.Config != nil &&
```
CharlyF, line 770: why is the config necessary for the admission controller?

CharlyF (approved):
Just curious about extra checks but I might be missing something, looks good otherwise

api/v1alpha1/datadogagent_default.go
clamoriniere: if never we change the default `Enabled` value. 
I tried to use the same pattern on every `Default...` fonction

api/v1alpha1/datadogagent_default.go
clamoriniere: Agree, it was to follow the same pattern everywhere.

controllers/datadogagent/clusteragent.go
clamoriniere: because the function takes value in `dda.Spec.ClusterAgent.Config.AdmissionController`.
due to the bug in the `DefaultClusterAgent()` when `if IsEqualStruct(*clusterAgent, DatadogAgentSpecClusterAgentSpec{}) ` was `true`, the `dda.Spec.ClusterAgent.Config` was `nil`. 

I think it is ok to keep this change, because it is always better to check point value before using it.

PR #353, opened 2021-08-10
fix clusterrolebinding update

clamoriniere:
### What does this PR do?

ClusterRoleBinding can't be update if we change its RoleRef because this field
is immutable.

To fix the issue, we now do `Delete` and `Create`. 

This PR also refactor the code base to define only one function responsible of updating a ClusterRoleBinding.

### Motivation

In operator v0.7, the clusterchecksrunner has its owner clusterrole (not shared with the agent ds).
So to migrate from Operator v0.6, the ClusterRoleBinding need to be updated. but a simple `Update()`
doesn't work due to the `RoleRef` immutability 

### Additional Notes

N/A

### Describe your test plan

1. Deploy the operator v0.6.0
1. Deploy the DatadogAgent with clusterchecksrunner enabled.
1. Chek the status of the DatadogAgent, it should be `Active=true` like here
   ```console
   NAME            ACTIVE   AGENT                CLUSTER-AGENT     CLUSTER-CHECKS-RUNNER   AGE
   datadog-agent   True     Running (15/15/15)   Running (2/2/2)   Running (3/3/3)          3m
   ```
1. Update the operator to use v0.7.0(-rc.x)
1. Wait that the new operator instance get the leadership
1. The status of the DatadogAgent should still be `Active=true`
1. You can also check that the ClusterRoleBinding has been recreated. (timestamp)

### DIFF
--- a/controllers/datadogagent/common_rbac.go
+++ b/controllers/datadogagent/common_rbac.go
@@ -11,6 +11,7 @@ import (
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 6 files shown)


controllers/datadogagent/common_rbac.go
```diff
@@ -161,3 +162,26 @@ func (r *Reconciler) cleanupServiceAccount(logger logr.Logger, client client.Cli
...
+		serviceAccountName: serviceAccountName,
+	}
+	newClusterRoleBinding := buildClusterRoleBinding(dda, info, version)
+	if !apiequality.Semantic.DeepEqual(newClusterRoleBinding.Subjects, clusterRoleBinding.Subjects) || !apiequality.Semantic.DeepEqual(newClusterRoleBinding.RoleRef, clusterRoleBinding.RoleRef) {
+		logger.V(1).Info("updateIfNeededClusterRoleBinding", "clusterRoleBinding.name", clusterRoleBinding.Name, "serviceAccount", serviceAccountName, "roleName", roleName)
+		// ClusterRoleBinding can be updated, if we change the RoleRef in it, we need to delete and recreate
+		if err := r.client.Delete(context.TODO(), clusterRoleBinding); err != nil {
+			return reconcile.Result{}, err
+		}
+		if err := r.client.Create(context.TODO(), newClusterRoleBinding); err != nil {
```
davidor, line 179: Should we ignore the "not exists" error in the `Delete` above?

Not sure if the following could happen:
- Delete is successful
- Create fails because an error connection
- When the job is retried it will fail because the Delete will return "not found"

clamoriniere: If it was delete, we should not go again in the update function, but instead in the create function. So  I think it should be ok. Worth case it will take 2 reconcile
Loop

davidor: :+1:

davidor (approved):

PR #355, opened 2021-08-11
Standardize image defaulting

clamoriniere:
### What does this PR do?

Update defaulting images logic and have a dedicated package for defaulting.

The idea here is to update the Agent and Cluster-Agent image version with 
the lastest stable version each time we release the Operator. 

### Motivation

Keep up-to-date the agent and cluster-agent when deploying a new Operator version.

### Additional Notes

N/A

### Describe your test plan

Deploy the example `examples/datadogagent/datadog-agent-with-credential-secret.yaml` 

* Check that the Agent image tag is `7.30.0` 
* Check that the Cluster-Agent image tag is `1.14.0`

### DIFF
--- a/api/v1alpha1/datadogagent_default.go
+++ b/api/v1alpha1/datadogagent_default.go
@@ -241,7 +241,8 @@ func DefaultDatadogAgentSpecAgentImage(agent *DatadogAgentSpecAgentSpec, name, t
 		imgOverride.Name = agent.Image.Name
 	}
 
-	if agent.Image.Tag == "" {
+	// Only default Tag if not already present in the image.Name
+	if !defaulting.IsImageNameContainsTag(agent.Image.Name) && agent.Image.Tag == "" {
 		agent.Image.Tag = tag
 		imgOverride.Tag = agent.Image.Tag
 	}
@@ -1082,7 +1083,8 @@ func DefaultDatadogClusterAgentImage(dca *DatadogAgentSpecClusterAgentSpec, name
 		imgOverride.Name = dca.Image.Name
 	}
 
-	if dca.Image.Tag == "" {
+	// Only default Tag if not already present in the image.Name
+	if !defaulting.IsImageNameContainsTag(dca.Image.Name) && dca.Image.Tag == "" {
 		dca.Image.Tag = tag
 		imgOverride.Tag = dca.Image.Tag
 	}
@@ -1150,7 +1152,8 @@ func DefaultDatadogAgentSpecClusterChecksRunnerImage(clc *DatadogAgentSpecCluste
 		imgOverride.Name = clc.Image.Name
 	}
 
-	if clc.Image.Tag == "" {
+	// Only default Tag if not already present in the image.Name
+	if !defaulting.IsImageNameContainsTag(clc.Image.Name) && clc.Image.Tag == "" {
 		clc.Image.Tag = tag
 		imgOverride.Tag = clc.Image.Tag
 	}
--- a/api/v1alpha1/datadogagent_default.go
+++ b/api/v1alpha1/datadogagent_default.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/DataDog/datadog-operator/pkg/defaulting"
 	"github.com/DataDog/datadog-operator/pkg/utils"
 	edsdatadoghqv1alpha1 "github.com/DataDog/extendeddaemonset/api/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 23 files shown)


apigirl (commented):
no docs team review needed on this one

pkg/defaulting/images_test.go
```diff
@@ -0,0 +1,27 @@
...
+import "testing"
+
+func TestGetLatestAgentImage(t *testing.T) {
+	tests := []struct {
+		name string
+		want string
+	}{
+		{
+			name: "default",
+			want: "gcr.io/datadoghq/agent:7.30.0",
```
davidor, line 17: This test will fail every time the latest version is updated. Should we change the test to match a regex instead of a specific version to avoid that?

pkg/defaulting/images.go
```diff
@@ -0,0 +1,37 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2012 Datadog, Inc.
+
+package defaulting
```
davidor, line 6: I wonder if `defaulting` is too generic and has the risk of becoming another "utils" package where anything can be placed. Given that this is just about images, maybe it makes sense to define an `images` package instead?

pkg/defaulting/images.go
```diff
@@ -0,0 +1,37 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2012 Datadog, Inc.
```
davidor, line 4: `2012` in the copyright 👀 
Same for the file created in this PR: `pkg/defaulting/images_test.go`.

api/v1alpha1/datadogagent_default.go
```diff
@@ -21,8 +22,6 @@ import (
 // default values
 const (
 	defaultLogLevel                         string = "INFO"
-	defaultAgentImageTag                    string = "7.28.0"
-	defaultClusterAgentImageTag             string = "1.12.0"
```
davidor, line 25: Related to my other comment, I think the two constants below should belong to the new package created in this PR.

api/v1alpha1/datadogagent_default_test.go
```diff
@@ -620,7 +621,7 @@ func TestDefaultDatadogAgentSpecAgent(t *testing.T) {
 				UseExtendedDaemonset: NewBoolPointer(false),
 				Image: &ImageConfig{
 					Name:        "gcr.io/datadog/agent:6.26.0",
-					Tag:         defaultAgentImageTag,
+					Tag:         defaulting.AgentLatestVersion,
```
davidor, line 624: The name in the line above doesn't correspond to "latest".

api/v1alpha1/datadogagent_default.go
clamoriniere: yes

pkg/defaulting/images_test.go
clamoriniere: I initialy didn't want to commit this file :)

clamoriniere: test updated to support futur tag bump

pkg/defaulting/images.go
clamoriniere: In a near futur, the idea is to have a package `defaulting` to move `datadogagent_default.go` logic from `/api/v1alpha1` to this new package.
It will be useful, if we create a dedicate repo for the CRD definitions.

api/v1alpha1/datadogagent_default_test.go
clamoriniere: I keep `Latest` in the name, because it correspond to the "latest released" version. 
Do you think we should find another name?

pkg/defaulting/images.go
davidor: :+1:

api/v1alpha1/datadogagent_default_test.go
davidor: What I meant in this case is that the `Name` field in the line above already includes the version (`6.26.0`) and it does not correspond to `defaulting.AgentLatestVersion`. I think that `:6.26.0` shouldn't be there.

clamoriniere: Yes, I now understand this is a weird behaviour of the defaulting, that doesn't check if the `Name` contains the tag. I will try to fix it

clamoriniere: I have updated the defaulting logic to only set the "default tag" if the tag is not already present in the image name.
you can review the change in this [commit](https://github.com/DataDog/datadog-operator/pull/355/commits/46d0b8b03bad88baac20797e279d4be6f7533b29)

davidor (approved):

PR #359, opened 2021-08-20
[controllers/datadogagent] Ensure ClusterRoles and ClusterRoleBindings are deleted

davidor:
### What does this PR do?

We set the DatadogAgent object as the owner of the ClusterRoles and the ClusterRoleBindings that it creates. The reason to do that is to make sure that a cascade delete of the DatadogAgent will also delete its associated ClusterRoles and ClusterRoleBindings. However, that does not happen because DatadogAgent is namespace-scoped, while ClusterRoles and ClusterRoleBindings are not. When assigning the owner reference, Kubernetes raises a warning, and at delete time, it simply ignores that reference. The result is that ClusterRoles and ClusterRoleBindings are never deleted. Also, when a DatadogAgent is created with the same name as a previous one, the owner reference is not updated and still points to the old one.

This PR solves the issue. Instead of using owner references, it uses annotations to know if a given ClusterRole or ClusterRoleBinding was created by a particular DatadogAgent. The owned ClusterRoles and ClusterRoleBindings are now deleted in the finalizer.

A couple of things to notice that we should probably address in separate PRs:
- Upgrading the operator to a version that includes the fix will not remove old ClusterRoles and ClusterRoleBindings. We'll need additional tooling to do that.
- I think that, because of the names that the operator choses when creating ClusterRoles and ClusterRoleBindings, it does not support creating 2 different DatadogAgents with the same name in different namespaces. It's because the names do not encode the namespace, only the name.

### Describe your test plan

1) Create a DatadogAgent and wait until everything is deployed and ready.
2) Check with `kubectl get clusterroles` that these 3 are there `agentname-agent`, `agentname-cluster-agent`, `agentname-cluster-checks-runner` (Replace agentname with the real one). Check also that `kubectl get clusterrolebindings` returns 3 with the same names.
3) Delete the DatadogAgent created in step 1.
4) Check with `kubectl get clusterroles` and `kubectl get clusterrolebindings` that the ones mentioned in step 2 have been deleted.

### DIFF
--- a/controllers/datadogagent/controller_test.go
+++ b/controllers/datadogagent/controller_test.go
@@ -308,9 +308,6 @@ func TestReconcileDatadogAgent_Reconcile(t *testing.T) {
 				if !hasAllNodeLevelRbacResources(clusterRole.Rules) {
 					return fmt.Errorf("bad cluster role, should contain all node level rbac resources, current: %v", clusterRole.Rules)
 				}
-				if !CheckOwnerReference(datadogAgent, clusterRole) {
-					return fmt.Errorf("bad cluster role, should be owned by the datadog operator, current owners: %v", clusterRole.OwnerReferences)
-				}
 
 				return nil
 			},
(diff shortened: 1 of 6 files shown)


controllers/datadogagent/controller_test.go
```diff
@@ -353,9 +350,6 @@ func TestReconcileDatadogAgent_Reconcile(t *testing.T) {
 				if err := c.Get(context.TODO(), types.NamespacedName{Name: rbacResourcesName}, clusterRoleBinding); err != nil {
 					return err
 				}
-				if !CheckOwnerReference(datadogAgent, clusterRoleBinding) {
```
clamoriniere, line 356: should we check here instead the Annotation `part-of` ?

davidor: I pushed a new commit that replaces those owner refs checks with the `part-of` ones 👍

clamoriniere (approved):

PR #362, opened 2021-08-23
Fix runtime security volumes creation

clamoriniere:
### What does this PR do?

Bug discovered during the QA of #342.

* Fix defaulting of system-probe if security.runtime is enabled.
* Mount the security-runtime empty dir even if `policydir` is not set.

### Additional Notes

### Describe your test plan

Bug discovered with the following `DatadogAgent` configuration

```yaml
kind: DatadogAgent
metadata:
  name: datadog
  namespace: datadog
spec:
  credentials:
    apiKey: REDACTED
    appKey: REDACTED
  agent:
    clusterName: foo
    config:
      kubelet:
        tlsVerify: false
    image:
      name: "gcr.io/datadoghq/agent:latest"
    apm:
      enabled: true
    process:
      enabled: true
      processCollectionEnabled: true
    log:
      enabled: true
    systemProbe:
      bpfDebugEnabled: true
    security:
      runtime:
        enabled: true
```

### DIFF
--- a/api/v1alpha1/datadogagent_default.go
+++ b/api/v1alpha1/datadogagent_default.go
@@ -615,12 +615,18 @@ func DefaultDatadogAgentSpecAgentApmUDS(apm *APMSpec) *APMUnixDomainSocketSpec {
 // DefaultDatadogAgentSpecAgentSystemProbe defaults the System Probe
 // This method can be re-run as part of the FeatureOverride
 func DefaultDatadogAgentSpecAgentSystemProbe(agent *DatadogAgentSpecAgentSpec) *SystemProbeSpec {
+	sysOverride := &SystemProbeSpec{}
+
 	if agent.SystemProbe == nil {
 		agent.SystemProbe = &SystemProbeSpec{Enabled: NewBoolPointer(defaultSystemProbeEnabled)}
-		return agent.SystemProbe
+		sysOverride = agent.SystemProbe
 	}
 
-	sysOverride := &SystemProbeSpec{}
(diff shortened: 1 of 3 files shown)


api/v1alpha1/datadogagent_default.go
```diff
@@ -615,12 +615,18 @@ func DefaultDatadogAgentSpecAgentApmUDS(apm *APMSpec) *APMUnixDomainSocketSpec {
...
 func DefaultDatadogAgentSpecAgentSystemProbe(agent *DatadogAgentSpecAgentSpec) *SystemProbeSpec {
+	sysOverride := &SystemProbeSpec{}
+
 	if agent.SystemProbe == nil {
 		agent.SystemProbe = &SystemProbeSpec{Enabled: NewBoolPointer(defaultSystemProbeEnabled)}
-		return agent.SystemProbe
+		sysOverride = agent.SystemProbe
+	}
+
+	if agent.Security != nil && BoolValue(agent.Security.Runtime.Enabled) {
```
davidor, line 625: Should we also check that `agent.Security.Runtime != nil` before `BoolValue(...)`?

clamoriniere: `agent.Security.Runtime` is not a pointer link: https://github.com/DataDog/datadog-operator/blob/main/api/v1alpha1/datadogagent_types.go#L648

davidor: Right :+1:

davidor (approved):

PR #374, opened 2021-09-13
Introducing v2alpha1 

CharlyF:
### What does this PR do?

- Modifies structure of the repo to [support multi api groups](https://book.kubebuilder.io/migration/multi-group.html). 
- Converted through operator-sdk:
```
operator-sdk edit --multigroup=true
```
- Introduce the storage flag and update the controller-gen to 0.5.0.
- Fix 2 json flag in the types: `containerSymlinksPath` and `ConfigDirSpec.Items` using the [atomic list](https://github.com/kubernetes-sigs/structured-merge-diff/issues/115) and [here](https://github.com/pivotal-k8s/community/blob/df4289b7a057d264cc32adf7d982936221c65371/contributors/devel/sig-architecture/api-conventions.md#declaring-patch-strategy-and-object-topology).
### Motivation

We are going to introduce a new API that will carry breaking changes, this PR is the foundation of the migration.

### Additional Notes

### Describe your test plan

Transparent change for now, only issue would be if someone vendors the project.


kayayarai (approved):
LGTM from documentation perspective

clamoriniere (approved):
💯

config/crd/kustomization.yaml
```diff
@@ -5,6 +5,7 @@ resources:
 - bases/v1beta1/datadoghq.com_datadogagents.yaml
 - bases/v1beta1/datadoghq.com_datadogmetrics.yaml
 - bases/v1beta1/datadoghq.com_datadogmonitors.yaml
+- bases/datadoghq.com_datadogagents.yaml
```
vboulineau, line 8: Should not be present

PR #375, opened 2021-09-15
Add a service with local internal traffic policy for traces and dogstatsd.

L3n41c:
### What does this PR do?

Add a service which leverages [internal traffic policy](https://kubernetes.io/docs/concepts/services-networking/service-traffic-policy/) for traces and dogstatsd.

### Motivation

Ease the use of dogstatsd and APM as an application could use a kubernetes service instead of having to rely on the downward API to target the local node agent.

### Additional Notes

### Describe your test plan

Deploy the agent with the operator on kubernetes 1.22+ cluster and check that a new service targeting the nodes agent with local internal traffic policy has been created.


jtappa (approved):
👍🏻 from docs team

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.15 as builder
+FROM golang:1.16 as builder
```
clamoriniere, line 2: could you update the golang version of the gitlab runner and github action too 🙇

clamoriniere (commented):
IMO what is missing is a way in the DatadogAgent CR to force the creation of the service if a customer has enabled it on k8s cluster 1.21

clamoriniere:
Also maybe it will require to update the Network policy to allow the traffic on this service?

Dockerfile
L3n41c: This had already been done in a previous PR #387:
https://github.com/DataDog/datadog-operator/blob/5d59e5faa9454a984afdae533b0a5e8303fd1455/.gitlab-ci.yml#L1
https://github.com/DataDog/datadog-operator/blob/5d59e5faa9454a984afdae533b0a5e8303fd1455/.github/workflows/main.yml#L13

L3n41c:
@clamoriniere

> IMO what is missing is a way in the DatadogAgent CR to force the creation of the service if a customer has enabled it on k8s cluster 1.21

I’ve just updated the PR to implement here the equivalent of DataDog/helm-charts#434 which includes a way to force the service creation.

> Also maybe it will require to update the Network policy to allow the traffic on this service?

I don’t think so because Network Policies apply to pods. Reaching a pod directly or through a Service doesn’t make any difference for the Network policy.

controllers/datadogagent/service.go
```diff
@@ -146,6 +148,33 @@ func (r *Reconciler) manageAdmissionControllerService(logger logr.Logger, dda *d
...
+
+func (r *Reconciler) manageAgentService(logger logr.Logger, dda *datadoghqv1alpha1.DatadogAgent) (reconcile.Result, error) {
+	// Service Internal Traffic Policy exists in Kube 1.21 but it is enabled by default since 1.22
+	gitVersion := version.Get().GitVersion
+	if testGitVersion != "" {
+		gitVersion = testGitVersion
+	}
+
+	forceLocalServiceEnable := dda.Spec.Agent.LocalService != nil && dda.Spec.Agent.LocalService.ForceLocalServiceEnable != nil && *dda.Spec.Agent.LocalService.ForceLocalServiceEnable
+	if !utils.IsAboveMinVersion(gitVersion, "1.22-0") && !forceLocalServiceEnable {
```
clamoriniere, line 161: should it be '1.22.0" instead of "1.22-0" ?

L3n41c: It was `1.22-0` on purpose.
`1.22` is equivalent to `1.22.0`.
The `-0` suffix is meant to match pre-release version. See [doc here](https://github.com/Masterminds/semver#working-with-prerelease-versions).
Concretely, whereas `1.23.1` matches both `>= 1.22` and `>= 1.22.0`, `1.23.1-rc.1` wouldn’t match any of those two.
But `1.23.1-rc.1` matches both `>= 1.22-0` and `>= 1.22.0-0` (which are equivalent).

clamoriniere (approved):
👍

PR #386, opened 2021-09-30
Fix RBAC resource name too long

clamoriniere:
### What does this PR do?

We introduced the `namespace` as prefix of the clusterrole and
clusterrolebinding to allow multi DatadogAgent deployment in a cluster.
But if the DatadogAgent namespace/name is very long like:
datadog-agent/datadog-agent the final rbac name for `orchestrator-explorer`
RBAC was: `datadog-agent-datadog-agent-orchestrator-explorer-cluster-agent`
which is longer than the limit of 64 chars.

This PR reduce the size by renaming it to `datadog-agent-datadog-agent-orch-exp-dca`

This PR also update the value in `app.kubernetes.io/instances` for cluster scoped resources like:
* clusterrole
* clusterrolebinding

### Motivation

Fix agent deployment

### Additional Notes

### Describe your test plan

try to deploy the datadogAgent named `datadog-agent` in the namespace: `datadog-agent`
the install should work.

### DIFF
--- a/controllers/datadogagent/common_rbac.go
+++ b/controllers/datadogagent/common_rbac.go
@@ -31,7 +31,7 @@ type roleBindingInfo struct {
 func buildRoleBinding(dda *datadoghqv1alpha1.DatadogAgent, info roleBindingInfo, agentVersion string) *rbacv1.RoleBinding {
 	return &rbacv1.RoleBinding{
 		ObjectMeta: metav1.ObjectMeta{
-			Labels:    getDefaultLabels(dda, info.name, agentVersion),
+			Labels:    getDefaultLabels(dda, dda.Name, agentVersion),
 			Name:      info.name,
 			Namespace: dda.Namespace,
 		},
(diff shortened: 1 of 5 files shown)


celenechang (approved):

davidor:
I think this only solves the issue in some cases. We can have the same issue if the DDA has a long name or is deployed in a namespace with a long name.

controllers/datadogagent/common_rbac.go
```diff
@@ -31,7 +31,7 @@ type roleBindingInfo struct {
 func buildRoleBinding(dda *datadoghqv1alpha1.DatadogAgent, info roleBindingInfo, agentVersion string) *rbacv1.RoleBinding {
 	return &rbacv1.RoleBinding{
 		ObjectMeta: metav1.ObjectMeta{
-			Labels:    getDefaultLabels(dda, info.name, agentVersion),
+			Labels:    getDefaultLabels(dda, dda.Name, agentVersion),
```
davidor, line 34: Shouldn't these two be `NewPartOfLabelValue(dda).String()` like the rest?

clamoriniere: Here it is a role not a clusterrole

davidor: You're right, the namespace of the DDA is not needed in roles, but I wonder if we should just use the same format everywhere for consistency.

`buildAgentClusterAgentRole` (https://github.com/DataDog/datadog-operator/pull/386/files/d695397051a98d999df874dabfeba119ddfde1dd#diff-1ab2275c9afe43600275bda29e5e505bb4ecda385774675935de82f60e27a1b7R1363)  is a role and uses `NewPartOfLabelValue(dda).String()`. That'd need to be changed if we keep `dda.Name` here.

clamoriniere: good catch

davidor (approved):

PR #387, opened 2021-09-30
Update kubernetes dependencies version + golang

clamoriniere:
### What does this PR do?

* Update to golang 1.16
* Update k8s dependencies to `v0.20.11`
* Update github.com/ulikunitz/xz to `v0.5.8`

### Motivation

Remove vulnerabilities and get bug fixies

### Additional Notes

### Describe your test plan

Run the Operator and try to deploy the agent


go.mod
```diff
@@ -20,16 +20,17 @@ require (
...
 	go.uber.org/zap v1.15.0
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
-	k8s.io/api v0.20.2
-	k8s.io/apiextensions-apiserver v0.20.2
-	k8s.io/apimachinery v0.21.0
-	k8s.io/cli-runtime v0.20.2
-	k8s.io/client-go v0.20.2
+	k8s.io/api v0.20.11
+	k8s.io/apiextensions-apiserver v0.20.11
+	k8s.io/apimachinery v0.21.5
```
davidor, line 29: This is the only one with v0.21.X.
I guess it's OK because it was already like that. Pointing it out just in case.

davidor (approved):

go.mod
vboulineau: Probably a mistake, should be aligned, good catch

clamoriniere: I don't rememenber why it is like that. but it is also like this in the `datadog-agent` repo

clamoriniere: done

PR #407, opened 2021-11-23
Do not serve v2alpha1 as schema not defined yet

vboulineau:
### What does this PR do?

Keep `v1alpha1` as default version

### Motivation

### Additional Notes

### Describe your test plan

`make deploy`, it should still work with `v1alpha1` CRDs.


apis/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -26,6 +26,7 @@ type DatadogAgentStatus struct {
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
+//+kubebuilder:unservedversion
```
clamoriniere, line 29: Just to confirm: adding this comment is changing the status of `served` in the CRD yaml file definition?

clamoriniere: does it allows us to release a version from `main` without exposing v2alpha1 to user?

vboulineau: Yes, correct. So `v2alpha1` will be present in the CRD but it's unusable. The other option is to flag `v2alpha1` as `skipped` in which case it's not generated at all.

clamoriniere (approved):

PR #409, opened 2021-11-24
Merge custom and bundled compliance benchmarks

lebauce:
### What does this PR do?

Instead of replacing the /etc/datadog-agent/compliance.d directory with the content
of a configmap, merge the benchmarks bundled with the agent and the one provided
with the configmap.

### Motivation

This behaviour is the same than the security policies and it allows to both add new benchmarks
and replace existing ones.

### Additional Notes

### Describe your test plan

### DIFF
--- a/controllers/datadogagent/agent_test.go
+++ b/controllers/datadogagent/agent_test.go
@@ -365,6 +365,12 @@ func complianceSecurityAgentVolumes() []corev1.Volume {
 			},
 		},
 		{
+			Name: datadoghqv1alpha1.SecurityAgentComplianceConfigDirVolumeName,
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
+			},
+		},
+		{
 			Name: datadoghqv1alpha1.ConfigVolumeName,
 			VolumeSource: corev1.VolumeSource{
 				EmptyDir: &corev1.EmptyDirVolumeSource{},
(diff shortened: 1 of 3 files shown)


controllers/datadogagent/agent_test.go
```diff
@@ -2278,6 +2284,7 @@ func Test_newExtendedDaemonSetFromInstance_CustomConfigMaps(t *testing.T) {
 			},
 		},
 		{
+
```
clamoriniere, line 2287: this new line is not needed
```suggestion
```

controllers/datadogagent/utils.go
```diff
@@ -508,6 +508,21 @@ func getConfigInitContainers(spec *datadoghqv1alpha1.DatadogAgentSpec, volumeMou
...
+				Name:      datadoghqv1alpha1.SecurityAgentComplianceCustomConfigDirVolumeName,
+				MountPath: "/etc/datadog-agent-compliance-benchmarks",
+			},
+			corev1.VolumeMount{
+				Name:      datadoghqv1alpha1.SecurityAgentComplianceConfigDirVolumeName,
+				MountPath: "/opt/datadog-agent/compliance.d",
+			},
+		)
+		copyCommands = append(copyCommands, "cp -v /etc/datadog-agent-compliance-benchmarks/* /opt/datadog-agent/compliance.d/")
+	}
```
clamoriniere, line 524: by doing files copy from the configmap in an init-container, it will require to delete/recreate the pod to get any updated made in the configmap. Is it what we want?

clamoriniere (commented):
I added a small comment. just to confirm the behavior

controllers/datadogagent/utils.go
lebauce: You're right. This is not ideal but I don't think we modify this configmap very often (in fact, we haven't modified this configmap only once or twice in the last year)

clamoriniere (approved):

PR #412, opened 2021-11-30
feat: Introduce the Features Factory

clamoriniere:
### What does this PR do?

This draft PR is here to illustrate how the Feature Factory can improve the generically of the DatadogAgent controller logic. 

### Motivation

* Ease code evolution
* Better separation of concern: each feature logic are implemented in a separated package that allows better code ownership detection
* Ease unit-testing: since each feature should implement the same interface. it will ease the creation of a generic unit-test framework and allow to test each feature separately.

### Additional Notes

Some code entry point: 
* the [factory implementation](https://github.com/DataDog/datadog-operator/blob/clamoriniere/generic-reconciler/controllers/datadogagent/feature/factory.go): instantiate and configure Feature when needed.
* the [dependencies Store](https://github.com/DataDog/datadog-operator/blob/clamoriniere/generic-reconciler/controllers/datadogagent/dependencies/store.go): use to keep track of every resource dependencies.

Then the new Reconcile function is define here: https://github.com/DataDog/datadog-operator/blob/clamoriniere/generic-reconciler/controllers/datadogagent/reconcilerv2.go

### Describe your test plan


controllers/datadogagent/feature/types.go
```diff
@@ -0,0 +1,69 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Feature Feature interface
+// It returns `true` if the Feature is used, else it return `false`.
+type Feature interface {
+	// Configure use to configure the internal of a Feature
+	// It should return `true` if the feature is enabled, else `false`.
+	Configure(dda *v2alpha1.DatadogAgent) bool
```
celenechang, line 24: Is this function both altering the DatadogAgent spec and checking if the feature is enabled?

celenechang: From the ksmcore example it seems only the latter, in which case maybe we can call it `IsConfigured` or `IsEnabled`

clamoriniere: yes the idea it to extra the information needed from the `DatadogAgent`. like this we don't need to access again the resource after when processing the feature

controllers/datadogagent/dependencies/store.go
```diff
@@ -0,0 +1,227 @@
...
+	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// OperatorStoreLabelKey used to identified which resource is managed by the store.
+	OperatorStoreLabelKey = "operator.datadoghq.com/managed-by-store"
+	// OperatorGenHashAnnotationKey annotation key used on a Resource in order to compare 2 resources without know the spec struct.
+	OperatorGenHashAnnotationKey = "operator.datadoghq.com/gen-hash"
```
vboulineau, line 30: Not used, could be easily implemented in the store, no?

controllers/datadogagent/feature/dummy/feature.go
```diff
@@ -0,0 +1,66 @@
...
+	dummyLabelValue = "true"
+)
+
+type dummyFeature struct{}
+
+func (f *dummyFeature) Configure(dda *v2alpha1.DatadogAgent) bool {
+	return false
+}
+
+func (f *dummyFeature) ConfigureV1(dda *v1alpha1.DatadogAgent) bool {
```
vboulineau, line 38: Should not be necessary with conversion in place

controllers/datadogagent/feature/dummy/feature.go
```diff
@@ -0,0 +1,66 @@
...
+	return false
+}
+
+// ManageDependencies allows a feature to manage its dependencies.
+// Feature's dependencies should be added in the store.
+func (f *dummyFeature) ManageDependencies(store feature.DependenciesStoreClient) error {
+	return nil
```
vboulineau, line 45: With the store API, it means each deps need to do:
```
Get -> Check if nil -> Not nil: Update
                    -> Nil: Create
```

I think we should think about how to improve this experience before implementing features?

controllers/datadogagent/feature/types.go
```diff
@@ -0,0 +1,69 @@
...
+type BuildFunc func(options *Options) Feature
+
+// DependenciesStoreClient dependencies store client interface
+type DependenciesStoreClient interface {
+	AddOrUpdate(kind kubernetes.ObjectKind, obj client.Object)
+	Get(kind kubernetes.ObjectKind, namespace, name string) (client.Object, bool)
+}
+
+// PodFeatureConfiguration use to store a Pod Feature Configuration.
+type PodFeatureConfiguration struct {
```
vboulineau, line 60: I'm not sure to understand how this will be used, wasn't able to find any reference?

controllers/datadogagent/feature/dummy/feature.go
clamoriniere: Agree, But for now. I use it to validate the PR in https://github.com/DataDog/datadog-operator/pull/430

we can easily remove it later

controllers/datadogagent/feature/dummy/feature.go
clamoriniere: IMO dependencies are not so often shared between features.

I was thinking to create a function `GetOrCreate` that return the instance or create a new instance automatically.
I'm open to other idea

controllers/datadogagent/dependencies/store.go
clamoriniere: it already the case and used here:
* https://github.com/DataDog/datadog-operator/pull/412/files#diff-48d116016241a7d62a8614e21fc9b23f37962ca9d1bb54fa2ee5abd447709ed3R74
* https://github.com/DataDog/datadog-operator/pull/412/files#diff-48d116016241a7d62a8614e21fc9b23f37962ca9d1bb54fa2ee5abd447709ed3R146

controllers/datadogagent/feature/types.go
clamoriniere: I was a tentative to implement your proposition about not exposing the `PodTemplate` to feature function but only a subset of what can be configured.

controllers/datadogagent/feature/dummy/feature.go
vboulineau: I was thinking about `Roles` for instance (or `ConfigMap` if we decide to move all configs to `datadog.yaml`). Having a `GetOrCreate` may not be useful as you will still need to check if the object is newly created or not to fill some fieds.

I think we way want to have `Managers` on top of the store. Like the `RBACManager`, offering a high level API like `AddPermissions(serviceAccount,  roleName, permissions)` and when `Apply` is called on the store, it lazily generates the necessary objects.

About the comment below, your intention looks like a `PodTemplateManager` to me, with its helper functions to manage each piece.

clamoriniere: @vboulineau 
I have added a new commit with a `RBACManager` https://github.com/DataDog/datadog-operator/pull/412/commits/b0722e26dd2c1ca1aaccaf7b663a83c589e9a8d6

vboulineau (approved):

controllers/datadogagent/controller.go
```diff
@@ -113,6 +118,43 @@ func (r *Reconciler) internalReconcile(ctx context.Context, request reconcile.Re
...
+	features, err := feature.BuildFeaturesV1(instance, reconcilerOptionsToFeatureOptions(&r.options, logger))
+	if err != nil {
+		return result, fmt.Errorf("unable to build features, err: %w", err)
+	}
+
+	// -----------------------
+	// Manage dependencies
+	// -----------------------
+	depsStore := dependencies.NewStore(&dependencies.StoreOptions{SupportCilium: r.options.SupportCilium})
+	rbacManager := feature.NewResourcesManagers(depsStore)
```
celenechang, line 143: should this be generalized to `resourceManager` instead of `rbacManager` ?

PR #447, opened 2022-02-11
[defaults] Fix cluster agent token autogeneration

davidor:
### What does this PR do?

Currently, when a cluster agent token is not specified, it is not autogenerated properly and agents cannot connect to the cluster agent. This PR fixes the issue.

When a token is not specified, this PR autogenerates one and stores it in the "DefaultOverride" attribute of the spec.

### Additional Notes

When the secret that contains the token changes, the cluster agent and the agents are not restarted, so they keep using the old token. This issue is not fixed by this PR yet.

### Describe your test plan

- Deploy a datadog agent without specifying a token in the spec (`credentials.token`). Check the status of the agents (`agent status`) and verify that they can connect with the cluster agent. Also check that when the datadog agent is deleted, the secret with the token (has the same name as the datadog agent object) is also deleted. You can also check that in the status of the datadog agent, `defaultOverride.credentials.token` matches the token stored in the secret.
- Deploy a datadog agent specifying a token in the spec (`credentials.token`) and check the same except the `defaultOverride` part, which doesn't apply in this case.


apis/datadoghq/v1alpha1/datadogagent_default.go
```diff
@@ -133,32 +133,46 @@ func DefaultDatadogAgent(dda *DatadogAgent) *DatadogAgentStatus {
...
-			dso.ClusterAgent.GeneratedToken = apiutils.GenerateRandomString(32)
-		}
-	} else if dso.ClusterAgent != nil {
-		// cleanup the generated token if after an update the user specify one
-		// in the spec.Credentials.Token
-		dso.ClusterAgent.GeneratedToken = ""
+	defaultClusterAgentToken(dda, dso)
+}
+
+func defaultClusterAgentToken(dda *DatadogAgent, dso *DatadogAgentStatus) {
```
clamoriniere, line 154: IMO this change is not fully backward compatible. 
what happen if the `dso.ClusterAgent.GeneratedToken` was already defined? 

can we add another step in the function to get the token present in `status.ClusterAgent.GeneratedToken` if present ?

controllers/datadogagent/secret_agent.go
```diff
@@ -38,8 +38,10 @@ func newAgentSecret(name string, dda *datadoghqv1alpha1.DatadogAgent) (*corev1.S
 	// Agent credentials has two more fields
 	if creds.Token != "" {
 		data[datadoghqv1alpha1.DefaultTokenKey] = []byte(creds.Token)
-	} else if isClusterAgentEnabled(dda.Spec.ClusterAgent) && dda.Status.ClusterAgent != nil {
-		data[datadoghqv1alpha1.DefaultTokenKey] = []byte(dda.Status.ClusterAgent.GeneratedToken)
+	} else if isClusterAgentEnabled(dda.Spec.ClusterAgent) {
+		if dda.Status.DefaultOverride != nil && dda.Status.DefaultOverride.Credentials != nil && dda.Status.DefaultOverride.Credentials.Token != "" {
```
clamoriniere, line 42: nit: create a function the return the token, or check that it is defined. 

I saw the same code here: https://github.com/DataDog/datadog-operator/pull/447/files#diff-c46f55565558bd8d50d96ff9d37415ce9a6adbb913178786b3e413ce10440be4R160-R162

apis/datadoghq/v1alpha1/datadogagent_default.go
davidor: I added a new commit with this :+1:

controllers/datadogagent/secret_agent.go
davidor: I added a new commit with this :+1:

clamoriniere (approved):

PR #454, opened 2022-03-10
Fix DatadogAgent secret backend usage

celenechang:
### What does this PR do?

The secret backend feature for DatadogAgent was not working properly. This fixes the feature so that any credentials can use it. In addition, the Operator no longer creates a secret if it's not needed (i.e. it will not create an empty secret).

Other changes:
- create secret_common.go to hold common functions
- add new examples/datadogagent/datadog-agent-secret-backend.yaml
- update secret backed documentation

### Motivation

Fix

### Additional Notes

### Describe your test plan

Use the new `datadog-agent-secret-backend.yaml` example file to test the secret backend feature.

For completeness, can test the following cases (make sure the keys are read properly and the Agent can connect to the Cluster Agent):

(Create a secret containing credentials with `kubectl create secret generic test-secret --from-literal=api_key='<token>' --from-literal=app_key='<token>' --from-literal=token='<token>' `)

1. secret backend not used (check that Operator creates a secret based on provided credentials)
2. only API key uses secret backend
3. only App key uses secret backend
4. only token uses secret backend
5. all three (API key, App key, token) use secret backend
6. set API key using APISecret
7. set App key using APPSecret
9. enable ExternalMetrics with keys configured in ExternalMetricsCredentials, no secret backend
10. enable ExternalMetrics with keys configured in ExternalMetricsCredentials, with secret backend

### DIFF
--- a/controllers/datadogagent/clusteragent.go
+++ b/controllers/datadogagent/clusteragent.go
@@ -36,7 +36,7 @@ import (
 )
 
 func (r *Reconciler) reconcileClusterAgent(logger logr.Logger, dda *datadoghqv1alpha1.DatadogAgent, newStatus *datadoghqv1alpha1.DatadogAgentStatus) (reconcile.Result, error) {
-	result, err := r.manageClusterAgentDependencies(logger, dda, newStatus)
+	result, err := r.manageClusterAgentDependencies(logger, dda)
 	if utils.ShouldReturn(result, err) {
 		return result, err
 	}
@@ -191,13 +191,13 @@ func newClusterAgentDeploymentFromInstance(logger logr.Logger, dda *datadoghqv1a
 	return dca, hash, err
 }
 
-func (r *Reconciler) manageClusterAgentDependencies(logger logr.Logger, dda *datadoghqv1alpha1.DatadogAgent, newStatus *datadoghqv1alpha1.DatadogAgentStatus) (reconcile.Result, error) {
-	result, err := r.manageAgentSecret(logger, dda, newStatus)
+func (r *Reconciler) manageClusterAgentDependencies(logger logr.Logger, dda *datadoghqv1alpha1.DatadogAgent) (reconcile.Result, error) {
+	result, err := r.manageAgentSecret(logger, dda)
 	if utils.ShouldReturn(result, err) {
 		return result, err
 	}
 
-	result, err = r.manageExternalMetricsSecret(logger, dda, newStatus)
+	result, err = r.manageExternalMetricsSecret(logger, dda)
 	if utils.ShouldReturn(result, err) {
 		return result, err
 	}
--- a/controllers/datadogagent/clusteragent.go
+++ b/controllers/datadogagent/clusteragent.go
@@ -32,6 +32,7 @@ import (
 	"github.com/DataDog/datadog-operator/pkg/controller/utils/comparison"
 	"github.com/DataDog/datadog-operator/pkg/controller/utils/datadog"
 	"github.com/DataDog/datadog-operator/pkg/kubernetes"
+	"github.com/DataDog/datadog-operator/pkg/secrets"
 )
 
 func (r *Reconciler) reconcileClusterAgent(logger logr.Logger, dda *datadoghqv1alpha1.DatadogAgent, newStatus *datadoghqv1alpha1.DatadogAgentStatus) (reconcile.Result, error) {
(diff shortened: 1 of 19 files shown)


celenechang:
working on tests

jtappa (approved):
👍🏻 from docs team

controllers/datadogagent/clusteragent.go
```diff
@@ -661,10 +672,19 @@ func getEnvVarsForClusterAgent(logger logr.Logger, dda *datadoghqv1alpha1.Datado
...
-		Name:      datadoghqv1alpha1.DDAPIKey,
-		ValueFrom: getAPIKeyFromSecret(dda),
-	})
+	// This triggers use of the secret backend.
+	// Otherwise, read from the default or configured secret
+	if secrets.IsEnc(dda.Spec.Credentials.DatadogCredentials.APIKey) {
+		envVars = append(envVars, corev1.EnvVar{
+			Name:  datadoghqv1alpha1.DDAPIKey,
+			Value: dda.Spec.Credentials.DatadogCredentials.APIKey,
+		})
```
clamoriniere, line 681: here it means we expose API in an envvar, so anyone that list,get pod can see the api key. 
I would prefer to keep creating a the secret to store the API/APP key. But is is not easy to maintain up-to-date the secret, and redeploy agent when the operator envvar is update (new operator deployment).

maybe we can keep the implementation that you propose and put in our backlog how we can propose something more secure.

controllers/datadogagent/secret_agent.go
```diff
@@ -17,26 +15,21 @@ import (
...
 
 func (r *Reconciler) manageAgentSecret(logger logr.Logger, dda *datadoghqv1alpha1.DatadogAgent, newStatus *datadoghqv1alpha1.DatadogAgentStatus) (reconcile.Result, error) {
 	return r.manageSecret(logger, managedSecret{name: utils.GetDefaultCredentialsSecretName(dda), requireFunc: needAgentSecret, createFunc: newAgentSecret}, dda, newStatus)
 }
 
 func newAgentSecret(name string, dda *datadoghqv1alpha1.DatadogAgent) (*corev1.Secret, error) {
-	if err := checkCredentials(dda); err != nil {
-		return nil, err
-	}
-
```
clamoriniere, line 31: I think we still need to check that the API (and maybe APP) keys are provided as envvar or in the DatadogAgent.spec

controllers/datadogagent/clusteragent.go
celenechang: right, but this would only be stored if the credential set is in the format `ENC[...]` . could that still an issue if a user knows about the SECRET_BACKEND_COMMAND arg?

controllers/datadogagent/secret_agent.go
celenechang: yes, it is checked in the `needAgentSecret` function, which is required to be truthy before `newAgentSecret` is called

celenechang: i thought about that more, and i think the conditions to check whether the secret needs to be created are checked for all accounted for in the `sufficiency` functions. they're a combination of checking for spec credential existence, checking if it uses secret backend, and if there is an envvar fallback. happy to discuss more (perhaps IRL)

controllers/datadogagent/clusteragent.go
clamoriniere: I agree that the secret backend `ENC[...]` is used it is ok to have this value directly as env var.

but, do we need to manage differently the API/APP key and token if it use the secret backend? 
In both case, the secret can contains the the `raw` value or the `ENC["key-secret"]. So for simplicity, I would have keep the value in a secret.

docs/secret_management.md
```diff
@@ -169,9 +170,57 @@ spec:
...
+  credentials:
+    apiKey: ENC[api_key]
+    appKey: ENC[app_key]
+    useSecretBackend: true
+  agent:
+    env:
+      - name: DD_SECRET_BACKEND_COMMAND
+        value: "/readsecret.sh"
+      - name: DD_SECRET_BACKEND_ARGUMENTS
+        value: "/etc/secret-volume"
```
clamoriniere, line 196: For `v2.DatadogAgent` we could imaging to have specific fields for configuring the secret-backend

controllers/datadogagent/clusteragent.go
celenechang: i didn't consider that, and it makes sense. thanks, will update to behave this way instead

docs/secret_management.md
celenechang: i have a list of a couple changes to make for v2 regarding this feature, will add this too

clamoriniere (approved):
💯

PR #471, opened 2022-04-21
Add NodeLabelsAsTags and NamespaceLabelsAsTags configuration options

JacksonDavenport:
### What does this PR do?

This adds the configuration options to use `agent.config.nodeLabelsAsTags` and `agent.config.namespaceLabelsAsTags` like: 

```yaml
spec: 
  agent:
    config:
      nodeLabelsAsTags:
        key1: value1
      namespaceLabelsAsTags:
        key2: value2
```
To add the respective environment variables for `DD_KUBERNETES_NODE_LABELS_AS_TAGS` and `DD_KUBERNETES_NAMESPACE_LABELS_AS_TAGS`.

https://docs.datadoghq.com/agent/kubernetes/tag/?tab=containerizedagent#tag-autodiscovery

Additionally this removes the behavior where `DD_KUBERNETES_POD_LABELS_AS_TAGS` and `DD_KUBERNETES_POD_ANNOTATIONS_AS_TAGS` would default to `{}` when not otherwise set. 

### Motivation

Parity with Helm Chart, and simplify behavior to not set those 2 environment variables when not needed.
https://datadoghq.atlassian.net/browse/CONTECO-86

### Describe your test plan

Can test with `DatadogAgent` configuration with various:

```yaml
spec:
  agent:
    config:
      nodeLabelsAsTags:
        key1: value1
        key2: value2
      podLabelsAsTags: 
        key3: value3 
      podAnnotationsAsTags:
        key4: value4
      namespaceLabelsAsTags:
        key5: value5
  #(...)
```

And validate the resulting `DD_KUBERNETES_...` environment variables are created in proper JSON format.


apis/datadoghq/v1alpha1/datadogagent_types.go
```diff
@@ -779,12 +779,22 @@ type NodeAgentConfig struct {
 	// +optional
 	Checksd *ConfigDirSpec `json:"checksd,omitempty"`
 
-	// Provide a mapping of Kubernetes Labels to Datadog Tags.
+	// Provide a mapping of Kubernetes Namespace Labels to Datadog Tags.
+	// <KUBERNETES_LABEL>: <DATADOG_TAG_KEY>
+	// +optional
+	NamespaceLabelsAsTags map[string]string `json:"namespaceLabelsAsTags,omitempty"`
```
celenechang, line 785: do we want this env var set in containers besides the `agent` container?

JacksonDavenport: Good catch, looks like in the Helm chart these 4 are a part of the common environment variables distributed to each container. Now _technically_ `NodeLabelsAsTags` would probably be fine to just live in the `agent` container since that is setting the host-tag. But will look at looping all of those into the other containers through that same common env setup in the operator.

alai97 (approved):
Looks good for docs!

apis/datadoghq/v1alpha1/datadogagent_types.go
celenechang: thanks for the update. i'm a bit torn because it becomes less intuitive when its config is under `NodeAgentConfig` (`agent` container) but the envvar gets applied to all containers. the existing pattern with `PodLabelsAsTags` was already "broken" in that it only applied to the `agent` container, right?

maybe for v1 we keep it as it originally was (reverting your latest commit) and for v2 we can implement this properly. what do you think?

JacksonDavenport: I was a torn about that too. You're right the current pattern for the `PodLabelsAsTags` was only applying it for the `agent` container so that was "broken" before.  

This could get moved into the parent as {`agent.podLabelsAsTags`, `agent.nodeLabelsAsTags`, ...}, or if it would be cleaner {`agent.tagExtraction.podLabelsAsTags`, `agent.tagExtraction.nodeLabelsAsTags`, ...}. Currently in the Helm Chart these are just under the `datadog` parent, so I guess the closest equivalent would be to load them under {`features.podLabelsAsTags`, `features.nodeLabelsAsTags`, ...}.

celenechang: yea, i think ideally they would be moved to the parent, but then it would introduce a breaking change. and just putting `nodeLabelsAsTags` in the parent would be confusing. (`features` is intended to be for "products", more or less.)

JacksonDavenport: For now will remain as {`agent.config.podLabelsAsTags`, `agent.config.nodeLabelsAsTags`, ...} to remain as non-breaking change. With the caveat this will _not_ add those env vars to the other containers. Will adjust in v2, as the structure is going to change anyway.

controllers/datadogagent/utils_test.go
```diff
@@ -586,3 +586,84 @@ func Test_getReplicas(t *testing.T) {
 		})
 	}
 }
+
+func Test_EnvVarsForMetadataAsTags(t *testing.T) {
```
celenechang, line 590:
```suggestion
func Test_getEnvVarsForMetadataAsTags(t *testing.T) {
```

celenechang (approved):
just one small nit - lgtm!

PR #476, opened 2022-05-03
Add ComponentsEnabled struct in feature fwk

clamoriniere:
### What does this PR do?

Each "feature" can report the components that need to be deployed.

### Motivation

### Additional Notes

### Describe your test plan


controllers/datadogagent/feature/types.go
```diff
@@ -16,27 +17,69 @@ import (
...
+//   * nil: the feature doesn't need the corresponding component.
+type ComponentsEnabled struct {
+	ClusterAgent       *bool
+	Agent              *bool
+	ClusterCheckRunner *bool
+}
+
+// IsEnabled return true if the Feature need to be enabled
+func (cc *ComponentsEnabled) IsEnabled() bool {
+	return apiutils.BoolValue(cc.ClusterAgent) || apiutils.BoolValue(cc.Agent) || apiutils.BoolValue(cc.ClusterCheckRunner)
```
celenechang, line 33: the cluster checks runner would never be enabled without the cluster agent, right? maybe the last `||` is not needed

controllers/datadogagent/controller.go
```diff
@@ -131,19 +135,24 @@ func reconcilerOptionsToFeatureOptions(opts *ReconcilerOptions, logger logr.Logg
 func (r *Reconciler) reconcileInstance(ctx context.Context, logger logr.Logger, instance *datadoghqv1alpha1.DatadogAgent) (reconcile.Result, error) {
 	var result reconcile.Result
 
-	features, err := feature.BuildFeaturesV1(instance, reconcilerOptionsToFeatureOptions(&r.options, logger))
+	features, componentsEnabled, err := feature.BuildFeaturesV1(instance, reconcilerOptionsToFeatureOptions(&r.options, logger))
```
celenechang, line 138: i've been mulling over this and this pattern seems OK to me. a suggestion is to rename this to `requiredComponents` to better indicate that this is information to be used later (i presume in the reconcile<component> functions below), rather than something that's already been enabled.

controllers/datadogagent/feature/types.go
```diff
@@ -16,27 +17,69 @@ import (
...
 
+// ComponentsEnabled use to know which component need to be enabled for the feature
+// If set to:
+//   * true: the feature needs the corresponding component.
+//   * false: the corresponding component needs to ne disabled for this feature.
+//   * nil: the feature doesn't need the corresponding component.
+type ComponentsEnabled struct {
+	ClusterAgent       *bool
+	Agent              *bool
+	ClusterCheckRunner *bool
```
celenechang, line 28: i'm still wondering if we could/should have this be container-level rather than component-level. that is, keep the same structure but add a few lines for the agent containers. it would be the simplest form of the info we need..

celenechang: we discussed in IRL, feel free to ignore!

controllers/datadogagent/merger/fake/VolumeManager.go
```diff
@@ -0,0 +1,94 @@
...
+
+	commonv1 "github.com/DataDog/datadog-operator/apis/datadoghq/common/v1"
+	merger "github.com/DataDog/datadog-operator/controllers/datadogagent/merger"
+
+	v1 "k8s.io/api/core/v1"
+)
+
+// VolumeManager is an autogenerated mock type for the VolumeManager type
+type VolumeManager struct {
+	VolumeByC      map[commonv1.AgentContainerName][]*v1.Volume
```
celenechang, line 14: curious if this needs to be by container?

controllers/datadogagent/feature/types.go
clamoriniere: it was to have something generic.
my idea was: if a feature require the clustercheckrunner it should deploy the clusteragent too.

controllers/datadogagent/merger/fake/VolumeManager.go
clamoriniere: indeed we can remove by container for volume

PR #477, opened 2022-05-04
add DatadogAgent v2alpha1 defaulting code

celenechang:
### What does this PR do?

Add defaulting code for DatadogAgent v2alpha1 CRD

### Additional Notes

Currently the code is not called in the reconcile loop


docs/configuration.v2alpha1.md
```diff
@@ -87,8 +88,8 @@ spec:
 | features.logCollection.podLogsPath | PodLogsPath allows log collection from a pod log path. Default: `/var/log/pods` |
 | features.logCollection.tempStoragePath | TempStoragePath (always mounted from the host) is used by the Agent to store information about processed log files. If the Agent is restarted, it starts tailing the log files immediately. Default: `/var/lib/datadog-agent/logs` |
 | features.npm.collectDNSStats | CollectDNSStats enables DNS stat collection. Default: false |
+| features.npm.enableConntrack | EnableConntrack enable the system-probe agent to connect to the netlink/conntrack subsystem to add NAT information to connection data. See also: http://conntrack-tools.netfilter.org/ Default: false |
```
kayayarai, line 91:
```suggestion
| features.npm.enableConntrack | EnableConntrack enables the system-probe agent to connect to the netlink/conntrack subsystem to add NAT information to connection data. See also: http://conntrack-tools.netfilter.org/ Default: false |
```

kayayarai (approved):
Minor edits

apis/datadoghq/v2alpha1/datadogagent_default.go
```diff
@@ -0,0 +1,362 @@
...
+// DefaultDatadogAgent defaults the DatadogAgentSpec GlobalConfig and Features.
+func DefaultDatadogAgent(dda *DatadogAgent) {
+	defaultGlobalConfig(&dda.Spec)
+
+	defaultFeaturesConfig(&dda.Spec)
+}
+
+// defaultGlobalConfig sets default values in DatadogAgentSpec.Global.
+func defaultGlobalConfig(ddaSpec *DatadogAgentSpec) {
+	if ddaSpec.Global.ClusterName == nil {
```
vboulineau, line 86: Is it really a good idea? (I'm guessing we do the same in v1, but does not look legit)

apis/datadoghq/v2alpha1/datadogagent_default.go
```diff
@@ -0,0 +1,362 @@
...
+		defaultClusterNameCopy := defaultClusterName
```
vboulineau, line 87: I imagine that we have a lot of those, could just a `func()` that returns the pointer

apis/datadoghq/v2alpha1/datadogagent_default.go
```diff
@@ -0,0 +1,362 @@
...
+		ddaSpec.Global.ClusterName = &defaultClusterNameCopy
+	}
+
+	defaultCredentials(ddaSpec)
+}
+
+// defaultCredentials sets default values Credentials in the DatadogAgentSpec
+func defaultCredentials(ddaSpec *DatadogAgentSpec) {
+	if _, ok := ddaSpec.Override[ClusterAgentComponentName]; ok && *ddaSpec.Override[ClusterAgentComponentName].Disabled {
```
vboulineau, line 96: I'm not sure this works in all cases. For instance if Cluster Agent is not explicitly disabled but no activated feature requires the Cluster Agent?

Also, I believe it will continuously change the `Token`, especially if we don't store the `defaulted` part in the `status`, no? I am not sure it should be in the defaulting part.
Maybe it's more something that should be handled in the code directly when checking if there's already a Secret and so on.

docs/configuration.v2alpha1.md
```diff
@@ -123,6 +126,7 @@ spec:
 | global.endpoint.credentials.appKey | AppKey configures your Datadog application key. If you are using clusterAgent.metricsProvider.enabled = true, you must set a Datadog application key for read access to your metrics. |
 | global.endpoint.credentials.appSecret.keyName | KeyName is the key of the secret to use. |
 | global.endpoint.credentials.appSecret.secretName | SecretName is the name of the secret. |
+| global.endpoint.credentials.token | Token is the token for communication between the NodeAgent and ClusterAgent |
```
vboulineau, line 129: Not sure we need the token field is the common struct as it pushes it down everywhere.

vboulineau (commented):
I wonder if we should default `false` default value? I am saying that because it adds an unnecessary burden to anyone adding a new feature (to keep consistent code and having a mixup of features that are defaulted to false and not some others).

Also wondering if this should not be the place where the `image`, `registry`, etc. is defaulted?

apis/datadoghq/v2alpha1/datadogagent_default.go
vboulineau: This one will definitely conflict across clusters. Maybe we could generate a random ID, but I'd expect this code to be in the Cluster Agent then, to be consistent with the Helm chart.

docs/configuration.v2alpha1.md
celenechang: do you think it'd be better in ComponentOverride? i'm a bit concerned about confusion around which component one would need to set it in

apis/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -63,13 +63,13 @@ type DatadogFeatures struct {
 	NPM *NPMFeatureConfig `json:"npm,omitempty"`
 	// USM (Universal Service Monitoring) configuration.
 	USM *USMFeatureConfig `json:"usm,omitempty"`
-	// Dogstatsd configuration.
-	Dogstatsd *DogstatsdConfig `json:"dogstatsd,omitempty"`
+	// DogStatsD configuration.
+	DogStatsD *DogStatsDFeatureConfig `json:"dogStatsD,omitempty"`
```
vboulineau, line 67: I'd keep the JSON lower case, could be confusing otherwise

docs/configuration.v2alpha1.md
vboulineau: I don't think it should be in `ComponentOverride`, it should be in global but outside of `credentials`

apis/datadoghq/v2alpha1/datadogagent_types.go
celenechang: yea i changed that because the API complained about the difference in casing. i talked to agent core and we decided to change it back to what was originally there (`Dogstatsd / dogstatsd`)

docs/configuration.v2alpha1.md
celenechang: i see, ok will do that

apis/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -495,6 +497,9 @@ type GlobalConfig struct {
 	// Credentials defines the Datadog credentials used to submit data to/query data from Datadog.
 	Credentials *DatadogCredentials `json:"credentials,omitempty"`
 
+	// Token is the token for communication between the NodeAgent and ClusterAgent
+	Token *string `json:"token,omitempty"`
```
vboulineau, line 501: Maybe `ClusterAgentToken`, in case we have multiple one day

apis/datadoghq/v2alpha1/datadogagent_default.go
```diff
@@ -0,0 +1,280 @@
...
+// defaultFeaturesConfig sets default values in DatadogAgentSpec.Features.
+// Note: many default values are set in the Datadog Agent code and are not set here.
+func defaultFeaturesConfig(ddaSpec *DatadogAgentSpec) {
+	if ddaSpec.Features == nil {
+		ddaSpec.Features = &DatadogFeatures{}
+	}
+
+	// LogsCollection Feature
+	if ddaSpec.Features.LogCollection != nil && *ddaSpec.Features.LogCollection.Enabled {
+		if ddaSpec.Features.LogCollection.ContainerCollectUsingFiles == nil {
```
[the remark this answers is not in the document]
celenechang: thanks for the suggestion!

PR #430, opened 2022-01-23
migration KSMCore to feature pattern

clamoriniere:
### What does this PR do?

Based on https://github.com/DataDog/datadog-operator/pull/412, This pull request migrate the logic to configure KSM core to the feature factory pattern.

To achieve it. lots of refactoring was needed:
* create new dedicated package for api `const` values that is shared between v1alpha1 and v2alpha1
* create shared structure between v1alpha1 and v2alpha1
* move some functions present in the `controllers/datadogagent` package in dedicated package to avoid circular dependencies

### Motivation

Provide an example about how to structure the features implementations in the future.

### Additional Notes

Anything else we should know when reviewing?

### Describe your test plan

Write there any instructions and details you may have to test your PR.

PR #479, opened 2022-05-05
[features] migrate OOM kill

celenechang:
### What does this PR do?

Migrate OOM kill check to use FeatureFactory.

This requires #430 before merging.

### Motivation

### Additional Notes

### Describe your test plan


controllers/datadogagent/feature/types.go
```diff
@@ -46,21 +46,21 @@ func (rc *RequiredComponents) Merge(in *RequiredComponents) *RequiredComponents
 //   * false: the corresponding component needs to ne disabled for this feature.
 //   * nil: the feature doesn't need the corresponding component.
 type RequiredComponent struct {
-	Required           *bool
-	RequiredContainers []apicommonv1.AgentContainerName
+	IsRequired *bool
+	Containers []apicommonv1.AgentContainerName
```
celenechang, line 50: suggested rename, wdyt @clamoriniere ?

controllers/datadogagent/feature/oom_kill/feature.go
```diff
@@ -0,0 +1,103 @@
...
+
+func buildOOMKillFeature(options *feature.Options) feature.Feature {
+	oomKillFeat := &oomKillFeature{}
+
+	return oomKillFeat
+}
+
+type oomKillFeature struct {
+	enable bool
+	owner  metav1.Object
```
clamoriniere, line 36: Why do you need owner?

controllers/datadogagent/feature/types.go
clamoriniere: Much better, thanks

controllers/datadogagent/feature/oom_kill/feature.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// Configure is used to configure the feature from a v2alpha1.DatadogAgent instance.
+func (f *oomKillFeature) Configure(dda *v2alpha1.DatadogAgent) feature.RequiredComponents {
+	if dda.Spec.Features.OOMKill != nil && apiutils.BoolValue(dda.Spec.Features.OOMKill.Enabled) {
+		f.enable = true
+	}
+
+	return feature.RequiredComponents{
+		Agent: feature.RequiredComponent{IsRequired: &f.enable},
```
clamoriniere, line 45: here by setting the  `IsRequired: false` when `f.enable` is false make the `Agent` globally unwanted (with the current logic implemented in the Feature Factory). 

the idea was:
* set to `false`: we should not deploy this component
* nil: the feature don't need it, but don't care if it is deployed
* set to `true`: the feature need this component.

we can revisite the logic.

But I would suggest to no set it if `f.enable == false`

celenechang: right, thanks for catching that

PR #481, opened 2022-05-08
[features] migrate npm feature

celenechang:
### What does this PR do?

Migrate NPM feature to new framework
Adjust VolumeManager to handle case when more than one but fewer than all containers need a volumeMount

### Motivation

### Additional Notes

### Describe your test plan


controllers/datadogagent/feature/npm/feature.go
```diff
@@ -0,0 +1,149 @@
...
+func (f *npmFeature) ConfigureV1(dda *v1alpha1.DatadogAgent) (reqComp feature.RequiredComponents) {
+	f.owner = dda
+
+	if dda.Spec.Features.NetworkMonitoring != nil && *dda.Spec.Features.NetworkMonitoring.Enabled {
+		f.enable = true
+		reqComp = feature.RequiredComponents{
+			Agent: feature.RequiredComponent{
+				Required: &f.enable,
+				RequiredContainers: []apicommonv1.AgentContainerName{
+					apicommonv1.ProcessAgentContainerName,
```
clamoriniere, line 70: I think we need also `CoreAgent` container due to the remote-tagger

controllers/datadogagent/merger/volume.go
```diff
@@ -18,12 +18,14 @@ type VolumeManager interface {
 	AddVolume(volume *corev1.Volume, volumeMount *corev1.VolumeMount)
 	// Add the volume to the PodTemplate and add the volumeMount to container matching the containerName.
 	AddVolumeToContainer(volume *corev1.Volume, volumeMount *corev1.VolumeMount, containerName commonv1.AgentContainerName)
+	// Add the volume to the PodTemplate and add the volumeMount to container matching the containerNames.
+	AddVolumeToContainers(volume *corev1.Volume, volumeMount *corev1.VolumeMount, containerName []commonv1.AgentContainerName)
```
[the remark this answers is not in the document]
celenechang: i was looking for a way to add the volume to 1 < n < 5 containers. should we just have it binary, that is, add to 1 or add to all?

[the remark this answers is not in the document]
clamoriniere: Ok now I got it. this function is about to add container to a subset of container present in the pod.

controllers/datadogagent/merger/volume.go
```diff
@@ -59,13 +65,18 @@ func (impl *volumeManagerImpl) AddVolumeWithMergeFunc(volume *corev1.Volume, vol
 	return nil
 }
 
-func (impl *volumeManagerImpl) AddVolumeToContainerWithMergeFunc(volume *corev1.Volume, volumeMount *corev1.VolumeMount, containerName commonv1.AgentContainerName, volumeMergeFunc VolumeMergeFunction, volumeMountMergeFunc VolumeMountMergeFunction) error {
+func (impl *volumeManagerImpl) AddVolumeToContainersWithMergeFunc(volume *corev1.Volume, volumeMount *corev1.VolumeMount, containerNames []commonv1.AgentContainerName, volumeMergeFunc VolumeMergeFunction, volumeMountMergeFunc VolumeMountMergeFunction) error {
```
clamoriniere, line 68: A solution should be to call the function `AddVolumeToContainerWithMergeFunc()` in a loop on `containerNames`. It will not be optimum in term of tentative to add the `Volume`, but it can reduce the code

celenechang: yea i thought about that, but in the worst case scenario it would loop through the `impl.podTmpl.Spec.Containers` 4 x 5 times..

we could add the `return` into the `if` block which would help a bit. do you prefer this route?

clamoriniere: we can keep the function as it is too 👍

clamoriniere (approved):

PR #488, opened 2022-05-10
introduce the component package

clamoriniere:
### What does this PR do?

Introduce the `component` package with the `NewDefaultClusterAgentPodTemplateSpec` function
that aims to create a default cluster-agent `PodTemplateSpec`

### Motivation

Refactore code for the feature factory

### Additional Notes

### Describe your test plan

it is code refactoring unit-tests and test should be green.


controllers/datadogagent/component/clusteragent/default.go
```diff
@@ -0,0 +1,130 @@
...
+						ContainerPort: 5005,
+						Name:          "agentport",
+						Protocol:      "TCP",
+					},
+				},
+				Env:          envVars,
+				VolumeMounts: volumeMounts,
+				Command:      nil,
+				Args:         nil,
+				SecurityContext: &corev1.SecurityContext{
```
celenechang, line 77: should we make this configurable as well?

clamoriniere: this is the default. so it is not configurable.

The Idea it to create later another function that will use `DatadogAgent.Spec.Override` to override the PodTemplate if needed. Or it will be configurable in a `Feature`

celenechang: sorry commented too quickly, after reviewing more i understand

PR #496, opened 2022-05-19
fix(ksm-core): set CLC Runner enabled properly

clamoriniere:
### What does this PR do?

Fix how we configure the `kubernetes_state_core` check feature:
* It should run in cluster-check-runner only if they are enabled.

### Motivation

### Additional Notes

### Describe your test plan

Create an DatadogAgent with ksm-core enable, but without CLC runner enabled:
* the check should run in the DCA

Then update the DatadogAgent with CLC runner enabled:
* the check should now run in the clc-runner.

### DIFF
--- a/controllers/datadogagent/feature/kubernetesstatecore/feature.go
+++ b/controllers/datadogagent/feature/kubernetesstatecore/feature.go
@@ -73,11 +73,11 @@ func (f *ksmFeature) Configure(dda *v2alpha1.DatadogAgent) feature.RequiredCompo
 
 		if dda.Spec.Features.ClusterChecks != nil && apiutils.BoolValue(dda.Spec.Features.ClusterChecks.Enabled) {
 			f.clusterChecksEnabled = true
-			output.ClusterCheckRunner.IsRequired = &f.clusterChecksEnabled
 
 			if apiutils.BoolValue(dda.Spec.Features.ClusterChecks.UseClusterChecksRunners) {
 				f.rbacSuffix = common.CheckRunnersSuffix
 				f.serviceAccountName = v2alpha1.GetClusterChecksRunnerServiceAccount(dda)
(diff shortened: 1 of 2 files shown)


celenechang (commented):
thank you! i think we need a similar change for `ConfigureV1` depending on `if dda.Spec.ClusterChecksRunner.Enabled`

controllers/datadogagent/feature/kubernetesstatecore/feature.go
```diff
@@ -96,12 +96,14 @@ func (f *ksmFeature) ConfigureV1(dda *v1alpha1.DatadogAgent) feature.RequiredCom
...
 		output.ClusterAgent.IsRequired = &f.enable
 
-		if dda.Spec.ClusterAgent.Config != nil && apiutils.BoolValue(dda.Spec.ClusterAgent.Config.ClusterChecksEnabled) {
-			f.clusterChecksEnabled = true
+		// tried to be align with the function "needClusterChecksRunner" https://github.com/DataDog/datadog-operator/blob/c07038fbf0b8993798ec5693cf6f191d7c369feb/controllers/datadogagent/clusterchecksrunner.go#L83-L92
+		if (apiutils.BoolValue(dda.Spec.ClusterChecksRunner.Enabled) || apiutils.BoolValue(dda.Spec.ClusterAgent.Config.ClusterChecksEnabled)) && apiutils.BoolValue(dda.Spec.Features.KubeStateMetricsCore.ClusterCheck) {
+			output.ClusterCheckRunner.IsRequired = apiutils.NewBoolPointer(true)
+
 			f.rbacSuffix = common.CheckRunnersSuffix
+			f.clusterChecksEnabled = true
```
celenechang, line 104: we should allow this to be `true` even if `dda.Spec.ClusterChecksRunner.Enabled` is `false`, right?

celenechang (approved):

PR #508, opened 2022-05-27
Do not create a pdb if agent has a single replica

### LINKED ISSUE #401: poddisruptionbudget agent causes node to be unable to drain
```
evicting pod datadog/datadog-cluster-agent-689f579b4f-dmrcn
error when evicting pods/"datadog-cluster-agent-689f579b4f-dmrcn" -n "datadog" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
```

**Describe what happened:**
Trying to automatically drain my nodes and patch them. During the cluster upgrade I can see that the nodes just hang in `Ready,SchedulingDisabled` state.

I manually run:
```shell
kubectl drain aks-standard-12457058-vmss00000d --delete-emptydir-data --ignore-daemonsets
```
and it gives me an error saying that the pdb(poddisruptionbudget) is creating the issue.
In the end I had to add `--disable-eviction=true` to my drain command to work around it.

I don't see any need to have a pdb when running the agent as a daemonset. The daemonset controller will handle it this logic.
If running in daemon mode I would´t create a pdb.

If you don't think it should be disabled by default at least provide a variable in CRD so I can disable it.

Drain your nodes.

**Additional environment details (Operating System, Cloud provider, etc):**
N/A

phillebaba:
This PR fixes #401 which is caused by a PDB being created even when their is only on replica of the agent.

### DIFF
--- a/controllers/datadogagent/poddisruptionbudget.go
+++ b/controllers/datadogagent/poddisruptionbudget.go
@@ -34,8 +34,11 @@ type (
 )
 
 func (r *Reconciler) manageClusterAgentPDB(logger logr.Logger, dda *datadoghqv1alpha1.DatadogAgent) (reconcile.Result, error) {
-	cleanUpCondition := !isClusterAgentEnabled(dda.Spec.ClusterAgent)
-	return r.managePDB(logger, dda, getClusterAgentPDBName(dda), buildClusterAgentPDB, cleanUpCondition)
+  if *dda.Spec.ClusterAgent.Replicas > 1 {
+    cleanUpCondition := !isClusterAgentEnabled(dda.Spec.ClusterAgent)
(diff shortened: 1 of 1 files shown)


gnuletik:
Thanks for this PR @phillebaba!

Can you add the `bug` label to this PR, so github-action's bot can approve it?

Thanks!

gnuletik:
@phillebaba Ah yes, that's probably locked to maintainers, my bad!

CharlyF:
Hi @phillebaba - I cherry-picked your contribution (for some reason I could not get the CI to run) and added a test case too.
PR https://github.com/DataDog/datadog-operator/pull/535
Thanks a lot, feel free to let us know if you are still facing issues.

PR #514, opened 2022-06-02
Implement global settings

khewonc:
### What does this PR do?

Implements some of the global settings

### Motivation

### Additional Notes

* Adds network policy and service managers for `NetworkPolicy` and `LocalService`
* Passes in resource manager to reconcile functions (`controllers/datadogagent/controller_reconcile_v2.go`) to create network policy and service and to store k8s version
* Passes the entire DDA (not only the global spec) and resource manager to `ApplyGlobalSettings` for network policy and service creation and to check whether certain features are used, e.g. ClusterChecksRunners
* Stores Kubernetes version info in store to determine when to create the local agent service

### Describe your test plan

Testing a few settings:

```yaml
spec:
  global:
    clusterName: "kind"
    kubelet:
      tlsVerify: false
    logLevel: "DEBUG"
    podAnnotationsAsTags:
      custom-tag2: custom-tag2
    podLabelsAsTags:
      custom-tag1: custom-tag1
    tags:
        - hostTag1
        - hostTag2
    networkPolicy:
      create: true
      flavor: kubernetes
```

* logLevel
* clusterName
* tags

```
===============
Agent (v7.37.1)
===============

  Log Level: DEBUG
...
  Hostnames
  =========
    cluster-name: kind
    host_aliases: [kind-worker-kind]
    hostname: kind-worker-kind
    socket-fqdn: datadog-agent-42zhj
    socket-hostname: datadog-agent-42zhj
    host tags:
      cluster_name:kind
      hostTag1
      hostTag2
      kube_cluster_name:kind
```

* podAnnotationsAsTags
* podLabelsAsTags

```
== Source workloadmeta-kubernetes_pod ==
Tags: [container_id:<container-id> custom-tag1:custom-value3 custom-tag2:custon-value2 display_container_name:redis_redis-xxxxxxxxxx-xxxxx env:$(TEST) image_id:docker.io/library/redis@sha256:<image-id> image_name:redis image_tag:latest kube_container_name:redis kube_deployment:redis kube_namespace:system kube_ownerref_kind:replicaset kube_ownerref_name:redis-xxxxxxxxxx kube_qos:Guaranteed kube_replica_set:redis-xxxxxxxxxx pod_name:redis-xxxxxxxxxx-xxxxx pod_phase:running short_image:redis]
===
```

* kubelet

```
> kubectl describe po <agent-pod>
Containers:
  agent:
    Environment:
      DD_KUBELET_TLS_VERIFY:                     false
```

* localService

```
> kubectl get svc
NAME                               TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)    AGE
datadog-agent                      ClusterIP   ##.##.##.###   <none>        8125/UDP   121m
```

* networkPolicy

```
> kubectl get networkpolicy
NAME                            POD-SELECTOR                                                                                AGE
datadog-agent                   app.kubernetes.io/instance=agent,app.kubernetes.io/part-of=system-datadog                   122m
datadog-cluster-agent           app.kubernetes.io/instance=cluster-agent,app.kubernetes.io/part-of=system-datadog           122m
datadog-cluster-checks-runner   app.kubernetes.io/instance=cluster-checks-runner,app.kubernetes.io/part-of=system-datadog   122m
```


controllers/datadogagent/override/global.go
```diff
@@ -6,16 +6,212 @@
...
+				},
+			}
+		}
+		manager.EnvVar().AddEnvVar(&corev1.EnvVar{
+			Name:      apicommon.DDKubeletHost,
+			ValueFrom: kubeletHostValueFrom,
+		})
+
+		// tlsverify defaults to true in agent code
+		if !*config.Kubelet.TLSVerify {
```
celenechang, line 151: maybe we should explicitly set this to default True in the defaulting code, and then do
```suggestion
		if !apiutils.BoolValue(config.Kubelet.TLSVerify) {
```

controllers/datadogagent/override/global.go
```diff
@@ -6,16 +6,212 @@
...
+	for _, c := range manager.PodTemplateSpec().Spec.InitContainers {
+		contList = append(contList, apicommonv1.AgentContainerName(c.Name))
+	}
+	for _, c := range manager.PodTemplateSpec().Spec.Containers {
+		contList = append(contList, apicommonv1.AgentContainerName(c.Name))
+	}
+	return contList
+}
+
+func getAgentCAPath(kubeletConfig *apicommonv1.KubeletConfig) string {
```
celenechang, line 210: i'm thinking it'd be cleaner if we took care of this [logic](https://github.com/DataDog/datadog-operator/blob/f19ecf5d53950e4593b64faab9b8bdcbf66e4008/apis/datadoghq/common/v1/types.go#L62) in the defaulting code as well

controllers/datadogagent/override/global.go
```diff
@@ -6,16 +6,212 @@
...
+				// ccr
+			}
+		}
+	}
 
-	// set image registry
+	// LocalService contains configuration to customize the internal traffic policy service.
+	gitVersion := resourcesManager.Store().GetVersionInfo()
+	// Service Internal Traffic Policy exists in Kube 1.21 but it is enabled by default since 1.22
+	if utils.IsAboveMinVersion(gitVersion, "1.21-0") {
```
celenechang, line 121: could we have these git version strings as constants at the top?

maycmlee (commented):
Nothing for docs to review.

controllers/datadogagent/merger/network_policy.go
```diff
@@ -0,0 +1,182 @@
...
+		// In order to not ask end-users to inject NetworkPolicy on the
+		// agent in the agent namespace, the agent must be allowed to
+		// probe any pod.
+		egress = []netv1.NetworkPolicyEgressRule{
+			{
+				Ports: append([]netv1.NetworkPolicyPort{}, ddIntakePort()),
+			},
+		}
+		ingress = []netv1.NetworkPolicyIngressRule{}
+	case v2alpha1.ClusterAgentComponentName:
```
CharlyF, line 91: Can we also make sure cross Cluster Agent communication works, similar to how it was introduced in our helm chart https://github.com/DataDog/helm-charts/pull/641

controllers/datadogagent/override/global.go
```diff
@@ -6,16 +6,201 @@
...
+			})
+
+			if config.Kubelet.HostCAPath != "" {
+				kubeletVol, kubeletVolMount := volume.GetVolumes(apicommon.KubeletCAVolumeName, config.Kubelet.HostCAPath, config.Kubelet.AgentCAPath, true)
+				manager.VolumeMount().AddVolumeMountToContainers(
+					&kubeletVolMount,
+					[]apicommonv1.AgentContainerName{
+						apicommonv1.CoreAgentContainerName,
+						apicommonv1.ProcessAgentContainerName,
+						apicommonv1.TraceAgentContainerName,
```
celenechang, line 159: just noting that in case one of these containers doesn't exist, nothing will happen

celenechang (approved):

controllers/datadogagent/controller_reconcile_v2.go
```diff
@@ -131,27 +126,36 @@ func (r *Reconciler) reconcileInstanceV2(ctx context.Context, logger logr.Logger
...
 
 	if requiredComponents.ClusterChecksRunner.IsEnabled() {
-		result, err = r.reconcileV2ClusterChecksRunner(logger, features, instance, newStatus)
+		result, err = r.reconcileV2ClusterChecksRunner(logger, features, instance, resourceManagers, newStatus)
 		if utils.ShouldReturn(result, err) {
 			return r.updateStatusIfNeededV2(logger, instance, newStatus, result, err)
 		}
 	}
 
+	// ------------------------------
```
davidor, line 150: I'm not sure about the implications of changing the order of this call. I'll need to take a deeper look at the code.
Why is this needed?

controllers/datadogagent/merger/service.go
```diff
@@ -0,0 +1,80 @@
...
+type serviceManagerImpl struct {
+	store dependencies.StoreClient
+}
+
+// AddService creates or updates service
+func (m *serviceManagerImpl) AddService(name, namespace string, spec *corev1.ServiceSpec) error {
+	obj, _ := m.store.GetOrCreate(kubernetes.ServicesKind, namespace, name)
+	service, ok := obj.(*corev1.Service)
+	if !ok {
+		return fmt.Errorf("unable to get from the store the Service %s", name)
```
davidor, line 45: It would be good to include the namespace in the error message too.

controllers/datadogagent/merger/service.go
```diff
@@ -0,0 +1,80 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
davidor, line 1: I think it'd be good to add tests for the functions exposed in this file.

We can take as an example the tests that we have for the other files in this package (`envvars_test.go`, `rbac_test.go`, etc.)

controllers/datadogagent/merger/service.go
```diff
@@ -0,0 +1,80 @@
...
+	if !ok {
+		return fmt.Errorf("unable to get from the store the Service %s", name)
+	}
+	service.Spec = *spec
+
+	return m.store.AddOrUpdate(kubernetes.ServicesKind, service)
+}
+
+// BuildAgentLocalService creates a local service for the node agent
+func (m *serviceManagerImpl) BuildAgentLocalService(dda metav1.Object, nameOverride string) error {
```
davidor, line 53: I'm not sure if this function belongs here. Notice that this function only needs the `m` parameter to call `AddService`.

Each of the files in this package deals with a specific kubernetes object (secret, volume, etc.) and just stores or sets them in the appropriate resource and "merges" them, but this one is different because it also needs the agent object to create the resource. I think the service should be constructed either in the `component` package or the current caller of this function, and then call `AddService()`.

controllers/datadogagent/override/global.go
```diff
@@ -6,26 +6,199 @@
 package override
 
 import (
+	"encoding/json"
```
davidor, line 9: I think it'd be nice to have an integration test with an agent that has these "global" options set, but it's something we can leave for a future PR if you prefer.

controllers/datadogagent/override/global.go
```diff
@@ -6,26 +6,199 @@
...
+			case v2alpha1.NetworkPolicyFlavorCilium:
+				// TODO
+			}
+		}
+	}
+
+	if componentName == v2alpha1.NodeAgentComponentName {
+		// Tags contains a list of tags to attach to every metric, event and service check collected.
+		if config.Tags != nil {
+			tags, _ := json.Marshal(config.Tags)
```
davidor, line 84: I don't think we should ignore errors in this `json.Marshal` call and the others in this function. These config values can come from the user and nothing guarantees that they're going to be valid json.

controllers/datadogagent/merger/network_policy.go
```diff
@@ -0,0 +1,203 @@
...
+type networkPolicyManagerImpl struct {
+	store dependencies.StoreClient
+}
+
+// AddKubernetesNetworkPolicy creates or updates a kubernetes network policy
+func (m *networkPolicyManagerImpl) AddKubernetesNetworkPolicy(name, namespace string, podSelector metav1.LabelSelector, policyTypes []netv1.PolicyType, ingress []netv1.NetworkPolicyIngressRule, egress []netv1.NetworkPolicyEgressRule) error {
+	obj, _ := m.store.GetOrCreate(kubernetes.NetworkPoliciesKind, namespace, name)
+	policy, ok := obj.(*netv1.NetworkPolicy)
+	if !ok {
+		return fmt.Errorf("unable to get from the store the NetworkPolicy %s", name)
```
davidor, line 47: It'd be good to include the namespace in this error.

controllers/datadogagent/merger/network_policy.go
```diff
@@ -0,0 +1,203 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
davidor, line 1: Can we add tests for this file?

controllers/datadogagent/merger/network_policy.go
```diff
@@ -0,0 +1,203 @@
...
+
+	policy.Spec.PodSelector = podSelector
+	policy.Spec.PolicyTypes = append(policy.Spec.PolicyTypes, policyTypes...)
+	policy.Spec.Ingress = append(policy.Spec.Ingress, ingress...)
+	policy.Spec.Egress = append(policy.Spec.Egress, egress...)
+	return m.store.AddOrUpdate(kubernetes.NetworkPoliciesKind, policy)
+}
+
+// BuildAgentKubernetesNetworkPolicy creates the base node agent kubernetes network policy
+func (m *networkPolicyManagerImpl) BuildKubernetesNetworkPolicy(dda metav1.Object, componentName v2alpha1.ComponentName) error {
```
davidor, line 58: See my comment about `BuildAgentLocalService` in the `service.go` file.
I think the same applies to this function.

controllers/datadogagent/controller_reconcile_v2.go
khewonc: The local service and network policy weren't getting created when I tested in a cluster. I think it's because this [Apply call](https://github.com/DataDog/datadog-operator/blob/377f96fe19a6a1fd4a9d54590e814fd1065a7944/controllers/datadogagent/controller_reconcile_v2.go#L121) to create the objects in the store was getting called before the reconcile for each component, which is where the global settings are applied. Moving this section to after the component reconciles allowed the local service and network policy to be created.

controllers/datadogagent/override/global.go
khewonc: 👍 I created a card in the backlog for this.

controllers/datadogagent/override/global.go
```diff
@@ -6,34 +6,220 @@
...
-		manager.EnvVar().AddEnvVarToContainer(apicommonv1.ProcessAgentContainerName, kubeletEnvVar)
 	}
 
-	// set image registry
+	if componentName == v2alpha1.NodeAgentComponentName {
+		// Tags contains a list of tags to attach to every metric, event and service check collected.
+		if config.Tags != nil {
+			tags, marshalErr := json.Marshal(config.Tags)
+			if marshalErr != nil {
+				logger.Info("Failed to unmarshal json input", "error", err)
```
davidor, line 92: This one should be `marshalErr` instead of `err`. Same for the other 2 `json.Marshal` calls below.

controllers/datadogagent/override/global.go
```diff
@@ -6,34 +6,220 @@
...
+		// LocalService contains configuration to customize the internal traffic policy service.
+		gitVersion := resourcesManager.Store().GetVersionInfo()
+		if utils.IsAboveMinVersion(gitVersion, minLocalServiceVersion) {
+			if utils.IsAboveMinVersion(gitVersion, minDefaultLocalServiceVersion) || (config.LocalService != nil && apiutils.BoolValue(config.LocalService.ForceEnableLocalService)) {
+				if config.LocalService != nil && config.LocalService.NameOverride != nil {
+					err = resourcesManager.ServiceManager().AddService(component.BuildAgentLocalService(dda, *config.LocalService.NameOverride))
+				} else {
+					err = resourcesManager.ServiceManager().AddService(component.BuildAgentLocalService(dda, ""))
+				}
+				if err != nil {
```
davidor, line 136: This could evaluate to true if we had an error in a previous block of code.
Might be worth to define a new `errAddService` or similar to avoid confusion.

controllers/datadogagent/controller_reconcile_v2.go
davidor: Thanks for clarifying. It think it should be OK :+1:

PR #521, opened 2022-06-21
[v2alpha1] Fix issues so agent and DCA boot without errors

davidor:
### What does this PR do?

Fixes the issues that were preventing the agent and the DCA to boot when using v2 (`KUSTOMIZE_CONFIG=config/test-v2`).

Notice that there might be other issues when enabling certain features. The goal of this PR is just to make the agent and the DCA start correctly when using the default config.

### Describe your test plan

Deploy with v2 (`KUSTOMIZE_CONFIG=config/test-v2 make deploy`) and check that the daemonset agent and the DCA start correctly.

### DIFF
--- a/controllers/datadogagent/component/agent/default.go
+++ b/controllers/datadogagent/component/agent/default.go
@@ -24,8 +24,9 @@ import (
 
 // NewDefaultAgentDaemonset return a new default agent DaemonSet
 func NewDefaultAgentDaemonset(dda metav1.Object, requiredContainers []common.AgentContainerName) *appsv1.DaemonSet {
-	podTemplate := NewDefaultAgentPodTemplateSpec(dda, requiredContainers)
 	daemonset := component.NewDaemonset(dda, apicommon.DefaultAgentResourceSuffix, component.GetAgentName(dda), component.GetAgentVersion(dda), nil)
+	podTemplate := NewDefaultAgentPodTemplateSpec(dda, requiredContainers, daemonset.GetLabels())
+
 	daemonset.Spec.Template = *podTemplate
 	return daemonset
 }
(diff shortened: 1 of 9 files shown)


controllers/datadogagent/component/agent/default.go
```diff
@@ -283,3 +307,29 @@ func volumeMountsForSystemProbe() []corev1.VolumeMount {
 		component.GetVolumeMountForAuth(),
 	}
 }
+
+// GetDefaultAgentClusterRolePolicyRules returns the default policy rules for
+// the Agent
+func GetDefaultAgentClusterRolePolicyRules() []rbacv1.PolicyRule {
```
celenechang, line 313: should we put this in `feature/enabledefault/rbac.go` instead?

controllers/datadogagent/override/global.go
```diff
@@ -15,6 +18,13 @@ import (
 func ApplyGlobalSettings(manager feature.PodTemplateManagers, config *v2alpha1.GlobalConfig) *corev1.PodTemplateSpec {
 	// TODO(operator-ga): implement ApplyGlobalSettings
 
+	if config != nil && config.Kubelet != nil && config.Kubelet.TLSVerify != nil {
```
celenechang, line 21: fine for now, but when we add call to default DatadogAgent then globalConfig shouldn't be `nil`

controllers/datadogagent/component/agent/default.go
davidor: Yes, let's do that :+1: I forgot that we already had an `rbac.go` in other features.

celenechang (approved):

PR #522, opened 2022-06-21
CWS: fix runtime policies loading, when no config map is defined

paulcacheux:
### What does this PR do?

When loading the agent through the operator with a default config like:
```yaml
kind: DatadogAgent
metadata:
  name: datadog
spec:
  credentials:
    apiKey: XX
    appKey: YY
  agent:
    security:
      runtime:
        enabled: true
```
The CWS policies directory is empty. The reason explaining this issue is that the policies dir is always mounted, even when no override is defined by the user. This means that in the default case we are overriding the default rules with a mount to a non-existant/empty folder

### Motivation

### Additional Notes

### Describe your test plan

### DIFF
--- a/controllers/datadogagent/utils.go
+++ b/controllers/datadogagent/utils.go
@@ -1317,20 +1317,16 @@ func getVolumesForAgent(dda *datadoghqv1alpha1.DatadogAgent) []corev1.Volume {
 		}
 	}
 
-	if isRuntimeSecurityEnabled(&dda.Spec) {
+	if isRuntimeSecurityEnabled(&dda.Spec) && dda.Spec.Agent.Security.Runtime.PoliciesDir != nil {
 		volumes = append(volumes,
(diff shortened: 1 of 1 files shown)


spikat (approved):

lebauce (approved):

CharlyF:
Seems like we still create the empty volume at the pod spec level when no custom policies are used. Should we get rid of this too or the volume mount in the system probe is enough?
https://github.com/DataDog/datadog-operator/blob/3114084f79a6c2d6feb647cd4cec499679271df7/controllers/datadogagent/utils.go#L1320-L1333

CharlyF:
> What this PR does is enough to fix the issue, if you feel that the volume definition would be even better I can include it in the second `if`

Just want to make sure we are covering all grounds.
Thanks for the contrib! 🚀

CharlyF (approved):

PR #526, opened 2022-06-22
[feature] CWS feature migration

khewonc:
### What does this PR do?

Migrate CWS feature

### Motivation

### Additional Notes

### Describe your test plan


celenechang (approved):

controllers/datadogagent/feature/cws/feature.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// Configure is used to configure the feature from a v2alpha1.DatadogAgent instance.
+func (f *cwsFeature) Configure(dda *v2alpha1.DatadogAgent) (reqComp feature.RequiredComponents) {
+	if dda.Spec.Features != nil && dda.Spec.Features.CWS != nil && apiutils.BoolValue(dda.Spec.Features.CWS.Enabled) {
+		cws := dda.Spec.Features.CWS
+
+		if apiutils.BoolValue(cws.SyscallMonitorEnabled) {
+			f.syscallMonitorEnabled = true
+		}
```
davidor, line 49: The `if` could be simplified: `f.syscallMonitorEnabled = apiutils.BoolValue(cws.SyscallMonitorEnabled)`

controllers/datadogagent/feature/cws/feature.go
```diff
@@ -0,0 +1,180 @@
...
+
+func buildCWSFeature(options *feature.Options) feature.Feature {
+	cwsFeat := &cwsFeature{}
+
+	return cwsFeat
+}
+
+type cwsFeature struct {
+	configMapConfig       *apicommonv1.ConfigMapConfig
+	configMapName         string
```
davidor, line 38: I think we don't need this attribute, because `apicommonv1.ConfigMapConfig` has a `Name`.

davidor (changes requested):
I tried this with the v2 reconciler and it doesn't work because the debugfs volume is not mounted in the security-agent:
```
2022-07-06 11:04:42 UTC | SYS-PROBE | ERROR | (cmd/system-probe/api/module/loader.go:59 in Register) | new module `security_runtime` error: debugfs is not mounted and is needed for eBPF-based checks, run "sudo mount -t debugfs none /sys/kernel/debug" to mount debugfs
```

controllers/datadogagent/feature/cws/feature_test.go
```diff
@@ -0,0 +1,352 @@
...
+		{
+			Name:          "v1alpha1 CWS not enabled",
+			DDAv1:         ddav1CWSDisabled.DeepCopy(),
+			WantConfigure: false,
+		},
+		{
+			Name:          "v1alpha1 CWS enabled",
+			DDAv1:         ddav1CWSEnabled,
+			WantConfigure: true,
+			Agent: &test.ComponentTest{
```
davidor, line 327: Nit: you can use the new helpers that we merged some days ago to avoid duplicating some code: https://github.com/DataDog/datadog-operator/blob/d0669c0682b56d716054ab45010de48b490bb925/controllers/datadogagent/feature/test/testsuite.go#L65

With that, you can remove the `createEmptyFakeManager` function defined at the top of this file, and here use: `Agent: test.NewDefaultComponentTest().WithWantFunc(cwsAgentNodeWantFunc)`. Same for the test below.

davidor:
As discussed offline, the error that I mentioned in my previous comment was already present in 0.8.x. Given that it's not related to the feature migration itself, I think we can merge this PR and address the error in a different one.

PR #547, opened 2022-07-08
[reconciler_v2] Keep agent status updated according to DaemonSet

davidor:
### What does this PR do?

Fixes a bug in the reconciler V2. The agent status was not updated always.

### Describe your test plan

The agent status should be updated according to the status of the daemonset.

### DIFF
--- a/controllers/datadogagent/controller_reconcile_v2_common.go
+++ b/controllers/datadogagent/controller_reconcile_v2_common.go
@@ -154,7 +154,13 @@ func (r *Reconciler) createOrUpdateDaemonset(parentLogger logr.Logger, dda *data
 		// check if same hash
 		needUpdate := !comparison.IsSameSpecMD5Hash(hash, currentDaemonset.GetAnnotations())
 		if !needUpdate {
-			// no need to update to stop here the process
+			// Even if the DaemonSet is still the same, it's status might have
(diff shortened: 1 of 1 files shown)


controllers/datadogagent/controller_reconcile_v2_common.go
```diff
@@ -154,7 +154,13 @@ func (r *Reconciler) createOrUpdateDaemonset(parentLogger logr.Logger, dda *data
 		// check if same hash
 		needUpdate := !comparison.IsSameSpecMD5Hash(hash, currentDaemonset.GetAnnotations())
 		if !needUpdate {
-			// no need to update to stop here the process
+			// Even if the DaemonSet is still the same, it's status might have
+			// changed (for example, the number of pods ready). This call is
+			// needed to keep the agent status updated.
+			now := metav1.NewTime(time.Now())
```
clamoriniere, line 160: should we pass `now` as input of the `createOrUpdateDaemonset` function?

davidor: I'm not sure. The function already has many params and `now` is easily accessible from here.
If it made testing easier, for example, we could do it. But I don't see a big benefit now. No strong opinion.

celenechang (approved):

PR #548, opened 2022-07-08
[tests] Add e2e test suite for reconciler V2

davidor:
### What does this PR do?

This PR adds a new e2e test suite that runs with `V2APIEnabled` set to true and a new Makefile target to run it: `integration-tests-v2`.

The suite is very similar to the existing one (that will replace once we migrate). The only difference is that `V2APIEnabled` is set to true and that the DatadogAgent CRD is patched as we do in `config/test-v2/storagev2_in_datadogagents.yaml`.

I've included a very basic test just as an example. The goal of this PR is to agree on the basics so we can add more tests in following PRs.

For now the tests are run with `USE_EXISTING_CLUSTER` (ref: https://book.kubebuilder.io/reference/envtest.html#configuring-your-test-control-plane) set to false (it's the default and what we're doing for V1). In the future, we should evaluate using a real cluster to help us catch more errors.

### Describe your test plan

Skip

### DIFF
--- a/controllers/suite_v2_test.go
+++ b/controllers/suite_v2_test.go
@@ -21,7 +21,6 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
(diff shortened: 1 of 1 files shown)


controllers/suite_v2_test.go
```diff
@@ -0,0 +1,152 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build integration_v2 && !integration
+// +build integration_v2,!integration
+
+// Note: This is very similar to "suite_test.go". The only differences are that
```
davidor, line 9: For reviewers, this is a big file, but it's mostly identical to the existing one (`suite_test.go`) and it will replace it once the migration is done.

controllers/datadogagent/dependencies/store.go
```diff
@@ -205,6 +206,13 @@ func (ds *Store) Apply(ctx context.Context, k8sClient client.Client) []error {
 				continue
 			}
 
+			// Special case for Service. The cluster IPs should be immutable, so don't
```
davidor, line 209: This is something that we were already doing for V1:
https://github.com/DataDog/datadog-operator/blob/24d3c0966e8be89a93fe6f6e529f71703b0e240a/controllers/datadogagent/service.go#L290

The integration tests fail without this.

davidor:
Rebased on top of main to fix conflicts.

PR #549, opened 2022-07-11
[reconciler_v2] Add integration tests with features enabled

davidor:
### What does this PR do?

Adds integration tests for the reconciler V2. It uses the test suite introduced in https://github.com/DataDog/datadog-operator/pull/548 to test multiple scenarios with different features enabled.

I've opened this PR as draft because it contains commits from https://github.com/DataDog/datadog-operator/pull/548. That one needs to be merged first.

These tests would have caught the bugs fixed in this PR: https://github.com/DataDog/datadog-operator/pull/543

One of the tests fails, because the operator crashes when enabling the log collection feature. The reason is that the feature relies on some default that's not being applied now.

### Describe your test plan

Skip.


davidor:
Rebased on top of main.

davidor:
This is ready for review now.

The test that checks the log collection feature was failing, but it was fixed after merging #545

/cc @celenechang

controllers/datadogagent_controller_v2_test.go
```diff
@@ -13,99 +13,257 @@ import (
...
-	Context("Basic deployment", func() {
-		namespace := "default"
-		name := "foo"
-		key := types.NamespacedName{
-			Namespace: namespace,
-			Name:      name,
-		}
+// These tests verify that a DatadogAgent deployment is successful.
+//
+// The function that checks is a deployment is successful is
```
celenechang, line 33:
```suggestion
// The function that checks if a deployment is successful is
```

davidor: Fixed :+1:

PR #576, opened 2022-07-26
[datadogagent/finalizer] Delete dependencies for V2 agents

davidor:
### What does this PR do?

Implements a finalizer for V2 agents. The finalizer removes all the resources associated with the agent.
Before this PR, cluster roles and cluster role bindings were not being deleted.

### Describe your test plan

When deleting a V2 agent, check that its associated resources are also deleted.


davidor:
Rebased on top of `main` to fix conflicts.

celenechang (approved):
thought about it for a while and i'm ok with this approach, feels future proof.

PR #580, opened 2022-07-28
[feature/cws] Fix tagger auth error

davidor:
### What does this PR do?

Fixes a bug in the CWS. There was an error in the system-probe container because it couldn't fetch the token for the remote tagger connection.

I did these changes based on what we're already doing in the helm charts.

### Describe your test plan

Enable the CWS feature and check that there are no errors in the system-probe container.


clamoriniere (approved):
Nice catch

khewonc (approved):

PR #435, opened 2022-01-31
Add conversion Webhook for v1alpha to v2alpha1

vboulineau:
### What does this PR do?

A brief description of the change being made with this pull request.

### Motivation

What inspired you to submit this pull request?

### Additional Notes

Anything else we should know when reviewing?

### Describe your test plan

Write there any instructions and details you may have to test your PR.

PR #590, opened 2022-08-05
fix config links and apiVersion of example

jonmoter:
### What does this PR do?

#435 renamed `docs/configuration.md` to `docs.configuration.v1alpha1.md`, but that broke links to that page. This fixes the links.

I also noticed that the example manifest in `configuration.v2alpha1.md` still referred to `v1alpha1`, which I assume is a copy-paste typo.

### Motivation

I clicked on the "Configuration" link in the README and it took me to a 404 page.

### Describe your test plan

n/a, doc change only

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ However, the Datadog Operator is still in beta, so it is not yet a recommended w
 
 ## Getting started
 
-See the [Getting Started][5] dedicated documentation to learn how to deploy the Datadog operator and your first Agent, and Configuration ([v1alpha1][12] or [v2alpha1][13]) to see examples and a list of all configuration keys.
+See the [Getting Started][5] dedicated documentation to learn how to deploy the Datadog operator and your first Agent, and [Configuration][12] to see examples and a list of all configuration keys.
 
 ## Functionalities
 
@@ -54,7 +54,6 @@ See the [How to Contribute page][9].
 [10]: https://catalog.redhat.com/software/operators/detail/5e9874986c5dcb34dfbb1a12
 [11]: https://operatorhub.io/operator/datadog-operator
 [12]: https://github.com/DataDog/datadog-operator/blob/main/docs/configuration.v1alpha1.md
-[13]: https://github.com/DataDog/datadog-operator/blob/main/docs/configuration.v2alpha1.md
 
 ## Release
 
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ However, the Datadog Operator is still in beta, so it is not yet a recommended w
 
 ## Getting started
 
-See the [Getting Started][5] dedicated documentation to learn how to deploy the Datadog operator and your first Agent, and [Configuration][12] to see examples and a list of all configuration keys.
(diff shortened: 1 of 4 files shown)


jonmoter:
Another pair of fixes to the configuration docs:

1. In `v2alpha1`, the `credentials` parameters now exist under `global`, but the example didn't show that.
2. In both example config pages, the text says it's setting `agent.image.name`, but that's not set in the example manifest right below it. Changed that to set the `clusterName`.

docs/getting_started.md
```diff
@@ -85,7 +85,7 @@ helm delete my-datadog-operator
 [2]: https://kubernetes.io/docs/tasks/tools/install-kubectl/
 [3]: https://artifacthub.io/packages/helm/datadog/datadog-operator
 [4]: https://app.datadoghq.com/account/settings#api
-[5]: https://github.com/DataDog/datadog-operator/blob/main/docs/configuration.md
+[5]: https://github.com/DataDog/datadog-operator/blob/main/docs/configuration.v1alpha1.md
```
urseberry, line 88: How did you decide that this should be v1 versus v2?

urseberry (approved):

docs/getting_started.md
jonmoter: The embedded YAML above refers to `v1alpha1` on line 66, so I figured it made more sense to link to that version.

jonmoter: IMHO, the docs are in an awkward state where `v2` seems to be the latest and greatest, but most of the examples still refer to `v1`. I dunno whether you want to change all the examples over to v2, or have parallel sets of instructions and examples for v1 and v2, or some other solution.

I wanted to keep the changes in my PR to be minimal, and just fix a few things that were clearly broken or incorrect.

README.md
```diff
@@ -28,7 +28,7 @@ However, the Datadog Operator is still in beta, so it is not yet a recommended w
 
 ## Getting started
 
-See the [Getting Started][5] dedicated documentation to learn how to deploy the Datadog operator and your first Agent, and [Configuration][12] to see examples and a list of all configuration keys.
+See the [Getting Started][5] dedicated documentation to learn how to deploy the Datadog operator and your first Agent, and Configuration ([v1alpha1][12] or [v2alpha1][13]) to see examples and a list of all configuration keys.
```
celenechang, line 31: `v2alpha1` is not ready for public use yet - until then, can we leave that off of this doc?

celenechang (commented):
thanks so much for the contribution @jonmoter ! just made one small request

README.md
jonmoter: Ah, good to know. Could I make a suggestion then that we put a warning on top of the `v2alpha1` markdown page saying something like "WIP, not ready for public use"?

'Cause when I poked around in the repo, I assumed that was ready for use.

celenechang: yes definitely, we can do that - thanks for pointing it out

jonmoter:
Okay, I've updated the change so I'm no longer linking to `v2alpha1` off the main README.md

PR #592, opened 2022-08-09
[override] Implement component overrides

davidor:
### What does this PR do?

This implements the overrides defined in a `v2alpha1.DatadogAgentComponentOverride` object.

The overrides are mostly about the pod template specs, but they also affects deployments, and the required components.

The overrides at container level were already implemented in a previous PR: https://github.com/DataDog/datadog-operator/pull/584

### Describe your test plan

Define a datadog agent object with overrides and check that everything works as expected.


controllers/datadogagent/override/podtemplatespec.go
```diff
@@ -6,22 +6,169 @@
...
+	// v2alpha1.DatadogAgentComponentOverride, like "Replicas" or "Disabled",
+	// that are not related with a pod template spec. The overrides for those
+	// attributes are not applied in this function.
+
+	if override == nil {
+		return
+	}
+
+	// TODO: seccomprootpath, seccompcustomprofile, seccomprofilename.
+	// They should only apply to system-probe, and I think there's some setup
```
davidor, line 31: fyi reviewers

clamoriniere: should it be part of a feature "deployment security", and I think it is at container lever that it need to be done.

which change we need to do at Pod level?

controllers/datadogagent/override/podtemplatespec.go
```diff
@@ -6,22 +6,177 @@
...
+		manager.EnvVar().AddEnvVar(&corev1.EnvVar{
+			Name:  env.Name,
+			Value: env.Value,
+		})
+	}
+
+	overrideCustomConfigs(manager, override.CustomConfigurations, componentName, ddaName)
+
+	// Note: override.ExtraConfd.ConfigData requires creating a configmap, so it cannot be handled here
+	if override.ExtraConfd != nil && override.ExtraConfd.ConfigMap != nil {
```
clamoriniere, line 63: could it be more generic like: override.AdditionalVolumes. 
and the customer add a configmap itself. and then it need to add a VolumeMount in the containerOverride?

wdyt?

clamoriniere (commented):
Small comment to maybe to include `override.confd`, because the customer can already do te same with "override.Volumes"

controllers/datadogagent/override/podtemplatespec.go
```diff
@@ -6,22 +6,173 @@
...
+func PodTemplateSpec(manager feature.PodTemplateManagers, override *v2alpha1.DatadogAgentComponentOverride, componentName v2alpha1.ComponentName, ddaName string) {
+	// Notice that there are several attributes in a
+	// v2alpha1.DatadogAgentComponentOverride, like "Replicas" or "Disabled",
+	// that are not related with a pod template spec. The overrides for those
+	// attributes are not applied in this function.
+
+	if override == nil {
+		return
+	}
+
```
celenechang, line 30: i think we need to allow for `createRbac` override

controllers/datadogagent/override/podtemplatespec.go
```diff
@@ -6,22 +6,173 @@
...
+			manager.PodTemplateSpec().Spec.Containers[i].Image = overriddenImage(container.Image, override.Image)
+		}
+
+		for i, initContainer := range manager.PodTemplateSpec().Spec.InitContainers {
+			manager.PodTemplateSpec().Spec.InitContainers[i].Image = overriddenImage(initContainer.Image, override.Image)
+		}
+	}
+
+	for _, env := range override.Env {
+		manager.EnvVar().AddEnvVar(&corev1.EnvVar{
```
celenechang, line 50: should we handle possibility of a conflict here?

controllers/datadogagent/override/podtemplatespec.go
celenechang: my :2cents: we have `override.Volumes` below; i think it might be beneficial to keep `override.ExtraConfd` to make this option transparent and be aligned with the helm chart

controllers/datadogagent/override/requiredcomponents.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/DataDog/datadog-operator/controllers/datadogagent/feature"
+)
+
+// RequiredComponents overrides a feature.RequiredComponents according to the given override options
+func RequiredComponents(reqComponents *feature.RequiredComponents, overrides map[v2alpha1.ComponentName]*v2alpha1.DatadogAgentComponentOverride) {
+	for component, overrideOpts := range overrides {
+		if apiutils.BoolValue(overrideOpts.Disabled) {
+			switch component {
+			case v2alpha1.NodeAgentComponentName:
+				reqComponents.Agent.IsRequired = apiutils.NewBoolPointer(false)
```
celenechang, line 20: is there ever a case where we would accept the node agent to be disabled?

controllers/datadogagent/override/requiredcomponents.go
```diff
@@ -0,0 +1,28 @@
...
+			case v2alpha1.ClusterAgentComponentName:
+				reqComponents.ClusterAgent.IsRequired = apiutils.NewBoolPointer(false)
```
celenechang, line 22: i wonder if we should log a warning if this conflicts with enabled features

controllers/datadogagent/override/requiredcomponents.go
clamoriniere: if a customer want only to use DCA feature like `external metrics`

controllers/datadogagent/override/requiredcomponents.go
clamoriniere: IMO it should be added in the DatadogAgent.status. maybe a new condition for storing the info.

docs/configuration.v2alpha1.md
```diff
@@ -12,25 +12,23 @@
 ## All configuration options
 
 The following table lists the configurable parameters for the `DatadogAgent`
-resource. For example, if you wanted to set a value for `global.clusterName`,
+resource. For example, if you wanted to set a value for `clusterName`,
```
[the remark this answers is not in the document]
celenechang: the problem is we have a header generator for this file and the CI job for the docs is currently broken. we should update the script in a separate PR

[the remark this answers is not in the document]
davidor: I'll fix it in a separate PR.

kayayarai (approved):

docs/configuration.v2alpha1.md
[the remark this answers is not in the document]
davidor: https://github.com/DataDog/datadog-operator/pull/604

controllers/datadogagent/controller_reconcile_dca.go
```diff
@@ -34,17 +40,64 @@ func (r *Reconciler) reconcileV2ClusterAgent(logger logr.Logger, features []feat
...
+	}
+	if err := rbacManager.DeleteRoleByComponent(string(datadoghqv2alpha1.ClusterAgentComponentName), dda.Namespace); err != nil {
+		return reconcile.Result{}, err
+	}
+	if err := rbacManager.DeleteClusterRoleByComponent(string(datadoghqv2alpha1.ClusterAgentComponentName)); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	newStatus.ClusterAgent = nil
+	return reconcile.Result{Requeue: true}, nil
```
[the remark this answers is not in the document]
davidor: There's no error in this case, so I think we shouldn't set requeue to true.

davidor:
Rebased on top of main to fix conflicts.

davidor:
As discussed offline, we'll fix the issues with the config file overrides in a future PR.

celenechang (approved):

PR #594, opened 2022-08-11
update metrics forwarder manager to use an interface for the status condition

celenechang:
### What does this PR do?

Update metrics forwarder manager to use an interface in lieu of the custom status condition. This is an intermediary step to support a metrics forwarder v2 that uses the metav1.Condition rather than the custom v1alpha1.DatadogAgentCondition.

### Motivation

### Additional Notes

### Describe your test plan

Deploy the operator with `operatorMetricsEnabled=true` and test a v1 DatadogAgent custom resource. Make sure metrics report as expected.

### DIFF
--- a/pkg/controller/utils/datadog/metrics_forwarder.go
+++ b/pkg/controller/utils/datadog/metrics_forwarder.go
@@ -84,7 +84,7 @@ type metricsForwarder struct {
 	decryptor           secrets.Decryptor
 	creds               sync.Map
 	baseURL             string
-	status              *v1alpha1.DatadogAgentCondition
+	status              ConditionInterface
 	credsManager        *config.CredentialManager
 	sync.Mutex
 }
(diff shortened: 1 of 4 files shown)


pkg/controller/utils/datadog/metrics_forwarder.go
```diff
@@ -184,13 +184,17 @@ func (mf *metricsForwarder) stop() {
 	close(mf.stopChan)
 }
 
-func (mf *metricsForwarder) getStatus() *v1alpha1.DatadogAgentCondition {
+// ConditionInterface is a generic interface to encompass DatadogAgentCondition (v1alpha1) and metav1.Condition (v2alpha1)
+type ConditionInterface interface {
```
clamoriniere, line 188: I like the idea, but do we only look at conditions in the status? Should we name it `StatusInterface`

celenechang: we talked separately, since it only deals with condition here going to keep the name as is. (in the next PR maybe i'll rename the function to `getCondition`)

clamoriniere: make sense

clamoriniere: do you plan to define the interface in a following PR?

celenechang: as far as i know there's nothing to define, since the condition doesn't have any actions?

clamoriniere (approved):

PR #596, opened 2022-08-11
[feature] Live container feature migration

khewonc:
### What does this PR do?

### Motivation

### Additional Notes

### Describe your test plan


controllers/datadogagent/override/global.go
```diff
@@ -187,43 +187,42 @@ func ApplyGlobalSettings(logger logr.Logger, manager feature.PodTemplateManagers
...
-			manager.VolumeMount().AddVolumeMountToContainers(
-				&criVolMount,
-				[]apicommonv1.AgentContainerName{
-					apicommonv1.CoreAgentContainerName,
-					apicommonv1.ProcessAgentContainerName,
-					apicommonv1.SecurityAgentContainerName,
-				},
+			runtimeVol, runtimeVolMount = volume.GetVolumes(apicommon.CriSocketVolumeName, *config.CriSocketPath, criSocketMountPath, true)
+		} else {
+			runtimeVol, runtimeVolMount = volume.GetVolumes(
```
khewonc, line 209: Per offline discussion, moving default runtime socket path to component defaults.

controllers/datadogagent/component/agent/default.go
```diff
@@ -292,6 +294,7 @@ func volumeMountsForTraceAgent() []corev1.VolumeMount {
 		component.GetVolumeMountForAuth(),
 		component.GetVolumeMountForConfig(),
 		component.GetVolumeMountForDogstatsdSocket(true),
+		component.GetVolumeMountForRuntimeSocket(true),
```
davidor, line 297: question: I thought that the live container feature should only affect the process container, why do we need the volume mount in the other containers too? Is it because of a different feature?

khewonc: In the operator v1 code, the runtime socket volume mount looked to be added to multiple containers by default: [core agent](https://github.com/DataDog/datadog-operator/blob/235ea457bc2b9e8fecceb28035b37a04eeccbecc/controllers/datadogagent/utils.go#L1442), [process agent](https://github.com/DataDog/datadog-operator/blob/235ea457bc2b9e8fecceb28035b37a04eeccbecc/controllers/datadogagent/utils.go#L1604), [security agent](https://github.com/DataDog/datadog-operator/blob/235ea457bc2b9e8fecceb28035b37a04eeccbecc/controllers/datadogagent/utils.go#L1788). In the helm chart, it was also added to the [trace agent container](https://github.com/DataDog/helm-charts/blob/20b27a7b408bf836d2d6fdf3d620f216fc021885/charts/datadog/templates/_container-trace-agent.yaml#L74), so I added the volume mount to all those containers to be safe. I didn't look into specific features to see whether the runtime socket was needed though

davidor (approved):

PR #601, opened 2022-08-23
update metrics forwarder to support v2

celenechang:
### What does this PR do?

Update Metrics Forwarder to support v2 as well as maintain support for v1 DatadogAgent.

### Motivation

### Additional Notes

### Describe your test plan

Deploy the operator with the v1 DatadogAgent CRD enabled and set flag `operatorMetricsEnabled=true` . Apply the DatadogAgent, for instance `examples/datadogagent/datadog-agent-all.yaml`. Make sure that operator metrics are reporting to Datadog.

Do the same with the v2 CRD, using `examples/v2alpha1/min.yaml`.


docs/how-to-contribute.md
```diff
@@ -47,26 +47,12 @@ KUSTOMIZE_CONFIG=config/test-v2 make deploy
...
-    credentials:
-      apiSecret:
-        secretName: datadog-secret
-        keyName: api-key
-      appSecret:
-        secretName: datadog-secret
-        keyName: app-key
-```
+Create a secret that contains an `api-key` and an `app-key`. By default the Operator is installed in the
+`system` namespace and only watches resources in the this namespace. As a result, the secret and deployment need to be done in the same namespace.
```
buraizu, line 51:
```suggestion
`system` namespace, and only watches resources in this namespace. As a result, the secret and deployment must be within the same namespace.
```

buraizu (changes requested):

apis/datadoghq/common/v1/agent_types.go
```diff
@@ -36,3 +39,65 @@ type AgentImageConfig struct {
...
+
+	LastUpdate  *metav1.Time `json:"lastUpdate,omitempty"`
+	CurrentHash string       `json:"currentHash,omitempty"`
+
+	// GeneratedToken corresponds to the generated token if any token was provided in the Credential configuration when ClusterAgent is
+	// enabled.
+	// +optional
+	GeneratedToken string `json:"generatedToken,omitempty"`
+
+	// Status corresponds to the ClusterAgent deployment computed status.
```
clamoriniere, line 96: DeploymentStatus can be use also for the clustercheckrunner so it can be good to remove the notion of ClusterAgent in the description

clamoriniere (commented):
the PR looks good. 
small comment about struct decription

clamoriniere (approved):
💯

docs/how-to-contribute.md
```diff
@@ -47,26 +47,12 @@ KUSTOMIZE_CONFIG=config/test-v2 make deploy
...
-        secretName: datadog-secret
-        keyName: api-key
-      appSecret:
-        secretName: datadog-secret
-        keyName: app-key
-```
+Create a secret that contains an `api-key` and an `app-key`. By default the Operator is installed in the
+`system` namespace, and only watches resources in this namespace. As a result, the secret and deployment must be within the same namespace.
+
+Then apply the `examples/v2alpha1/min.yaml` file which contains the mininum information needed to deploy the agent and related services.
```
buraizu, line 53:
```suggestion
Then apply the `examples/v2alpha1/min.yaml` file which contains the mininum information needed to deploy the Agent and related services.
```

buraizu (changes requested):

buraizu (approved):

PR #607, opened 2022-09-01
update v2 status

celenechang:
### What does this PR do?

Small updates to v2alpha1 DatadogAgent status

### Motivation

### Additional Notes

### Describe your test plan

### DIFF
--- a/controllers/datadogagent/controller_reconcile_v2_common.go
+++ b/controllers/datadogagent/controller_reconcile_v2_common.go
@@ -92,7 +92,7 @@ func (r *Reconciler) createOrUpdateDeployment(parentLogger logr.Logger, dda *dat
 		now := metav1.NewTime(time.Now())
 		err = kubernetes.UpdateFromObject(context.TODO(), r.client, updateDeployment, currentDeployment.ObjectMeta)
 		if err != nil {
-			updateStatusFunc(nil, newStatus, now, metav1.ConditionFalse, "UpdateFailed", "Unable to update Deployment")
+			updateStatusFunc(nil, newStatus, now, metav1.ConditionFalse, "UpdateSucceeded", "Unable to update Deployment")
 			return reconcile.Result{}, err
 		}
 		event := buildEventInfo(updateDeployment.Name, updateDeployment.Namespace, deploymentKind, datadog.UpdateEvent)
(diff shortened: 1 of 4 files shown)


controllers/datadogagent/controller_reconcile_v2_common.go
```diff
@@ -92,12 +92,12 @@ func (r *Reconciler) createOrUpdateDeployment(parentLogger logr.Logger, dda *dat
 		now := metav1.NewTime(time.Now())
 		err = kubernetes.UpdateFromObject(context.TODO(), r.client, updateDeployment, currentDeployment.ObjectMeta)
 		if err != nil {
-			updateStatusFunc(nil, newStatus, now, metav1.ConditionFalse, "CreateFailed", "Unable to Update Deployment")
+			updateStatusFunc(nil, newStatus, now, metav1.ConditionFalse, "UpdateFailed", "Unable to update Deployment")
```
clamoriniere, line 95: I think in that case the condition should be `True` 
```suggestion
			updateStatusFunc(nil, newStatus, now, metav1.ConditionTrue, "UpdateFailed", "Unable to update Deployment")
```

clamoriniere: Also in the `else` you should set the opposite

controllers/datadogagent/controller_reconcile_v2_common.go
```diff
@@ -92,12 +92,12 @@ func (r *Reconciler) createOrUpdateDeployment(parentLogger logr.Logger, dda *dat
...
 			return reconcile.Result{}, err
 		}
 		event := buildEventInfo(updateDeployment.Name, updateDeployment.Namespace, deploymentKind, datadog.UpdateEvent)
 		r.recordEvent(dda, event)
-		updateStatusFunc(updateDeployment, newStatus, now, metav1.ConditionTrue, "DeploymentUpdated", "Deployment updated")
+		updateStatusFunc(updateDeployment, newStatus, now, metav1.ConditionTrue, "UpdateSucceeded", "Deployment updated")
```
clamoriniere, line 100: I feel like it should be `metav1.ConditionFalse` because we are in the error case `if err !=nil` so it means that the "update" might have failed

clamoriniere (commented):
Not link to the PR directly but, I think the conditions set at the end of the reconcile should be review.

controllers/datadogagent/controller_reconcile_v2_common.go
```diff
@@ -92,18 +92,18 @@ func (r *Reconciler) createOrUpdateDeployment(parentLogger logr.Logger, dda *dat
 		now := metav1.NewTime(time.Now())
 		err = kubernetes.UpdateFromObject(context.TODO(), r.client, updateDeployment, currentDeployment.ObjectMeta)
 		if err != nil {
-			updateStatusFunc(nil, newStatus, now, metav1.ConditionFalse, "CreateFailed", "Unable to Update Deployment")
+			updateStatusFunc(nil, newStatus, now, metav1.ConditionFalse, "UpdateSucceeded", "Unable to update Deployment")
```
clamoriniere, line 95: `UpdateSucceeded` and `CreateSucceeded` could defined at `const string`

clamoriniere (approved):

PR #610, opened 2022-09-09
fix dca default token

celenechang:
### What does this PR do?

Generate a cluster agent token when one is not provided
Allow for a user to specify the token's presence in a secret

### Motivation

Better OOTB support

### Describe your test plan

- Make sure v2alpha1 can be deployed when not setting a cluster agent token string or specifying a token secret


cswatt (approved):

davidor:
I think there might be a simpler way to do this. The problem that I see is that the `newStatus *v2alpha1.DatadogAgentStatus` is defined in the `Configure` func of the `Feature` interface, but the only feature that needs it is the "enabledefault" feature, and it's for something very specific, making sure that the auto-generated token is stored in the status.

I think that if we do that work in the `reconcileInstanceV2` function of the Reconciler, we can avoid changing the `Feature` interface and passing the status to all the methods. We could do something like this instead, before the "Start reconcile Components" section:

```go
	userSpecifiedClusterAgentToken := instance.Spec.Global.ClusterAgentToken != nil || instance.Spec.Global.ClusterAgentTokenSecret != nil
	if !userSpecifiedClusterAgentToken {
		ensureAutoGeneratedTokenInStatus(instance, newStatus, resourceManagers, logger)
	}
```

```go
func ensureAutoGeneratedTokenInStatus(instance *datadoghqv2alpha1.DatadogAgent, newStatus *datadoghqv2alpha1.DatadogAgentStatus, resourceManagers feature.ResourceManagers, logger logr.Logger) {
	if instance.Status.ClusterAgent != nil && instance.Status.ClusterAgent.GeneratedToken != "" {
		// Already there. Nothing to do.
		return
	}

	// The secret should have been created in the "enabledefault" feature.
	tokenSecret, exists := resourceManagers.Store().Get(
		kubernetes.SecretsKind, instance.Namespace, datadoghqv2alpha1.GetDefaultDCATokenSecretName(instance),
	)
	if !exists {
		logger.V(1).Info("expected autogenerated token was not created by the \"enabledefault\" feature")
		return
	}

	autoGeneratedToken := tokenSecret.(*v1.Secret).Data[apicommon.DefaultTokenKey]
	if newStatus == nil {
		newStatus = &datadoghqv2alpha1.DatadogAgentStatus{}
	}
	if newStatus.ClusterAgent == nil {
		newStatus.ClusterAgent = &commonv1.DeploymentStatus{}
	}
	// Persist generated token for subsequent reconcile loops
	newStatus.ClusterAgent.GeneratedToken = string(autoGeneratedToken)
}

```

Would something like this work?

controllers/datadogagent/feature/enabledefault/feature.go
```diff
@@ -42,10 +45,17 @@ func buildDefaultFeature(options *feature.Options) feature.Feature {
...
+		dF.logger = options.Logger
+	}
+
+	return dF
 }
 
 type defaultFeature struct {
-	owner metav1.Object
+	owner  metav1.Object
+	logger logr.Logger
```
davidor, line 58: This is no longer needed.

davidor (approved):

PR #618, opened 2022-09-23
Fix conversion from v1 to v2

CharlyF:
### What does this PR do?

Without this, the conversion of v1alpha1 to v2alpha1 is not fully working, the apiserver will return:

```
0920 22:10:47.131288       1 watcher.go:327] failed to prepare current and previous objects: conversion webhook for datadoghq.com/v2alpha1, Kind=DatadogAgent returned invalid object: must have the same name: datadog !=
```

### Motivation

Support converting v1 to v2.

### Additional Notes

### Describe your test plan


CharlyF:
After doing some testing, it turns out that `ConvertFrom` in the v1alpha1 package is called even if there is no trace of v1 in the cluster. In order not to block the release we will merge this and look to better support this in the future.

clamoriniere (approved):

PR #650, opened 2022-11-23
Add additional configuration for CWS (security profiles and network)

paulcacheux:
### What does this PR do?

### Motivation

### Additional Notes

### Describe your test plan

### DIFF
--- a/apis/datadoghq/v2alpha1/datadogagent_types.go
+++ b/apis/datadoghq/v2alpha1/datadogagent_types.go
@@ -240,11 +240,28 @@ type CWSFeatureConfig struct {
 	// +optional
 	SyscallMonitorEnabled *bool `json:"syscallMonitorEnabled,omitempty"`
 
+	Network          *CWSNetworkConfig          `json:"network,omitempty"`
+	SecurityProfiles *CWSSecurityProfilesConfig `json:"securityProfiles,omitempty"`
+
 	// CustomPolicies contains security policies.
 	// The content of the ConfigMap will be merged with the policies bundled with the agent.
 	// Any policies with the same name as those existing in the agent will take precedence.
 	// +optional
 	CustomPolicies *CustomConfig `json:"customPolicies,omitempty"`
+}
+
+type CWSNetworkConfig struct {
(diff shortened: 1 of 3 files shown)


cswatt (approved):

apis/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -240,13 +240,30 @@ type CWSFeatureConfig struct {
...
 	// CustomPolicies contains security policies.
 	// The content of the ConfigMap will be merged with the policies bundled with the agent.
 	// Any policies with the same name as those existing in the agent will take precedence.
 	// +optional
 	CustomPolicies *CustomConfig `json:"customPolicies,omitempty"`
 }
 
+type CWSNetworkConfig struct {
+	// Enabled enables Cloud Workload Security Network detections.
+	// Default: false
```
khewonc, line 255: Should this default be `true` based on `defaultCWSNetworkEnabled` above?

khewonc:
> how to add unit tests for this ?

You can add tests in the `feature_test.go` file. It's a little disorganized, but something like [this section](https://github.com/DataDog/datadog-operator/blob/d1ab8a94e6692adc02aef08c101cb3a2f5f696f8/controllers/datadogagent/feature/cws/feature_test.go#L67-L81) would be the DDA spec to test, with [this function](https://github.com/DataDog/datadog-operator/blob/d1ab8a94e6692adc02aef08c101cb3a2f5f696f8/controllers/datadogagent/feature/cws/feature_test.go#L118) listing out the expect pod template spec based on the DDA spec. Some features that were added later have tests that were reorganized to be more modular, in case you'd prefer this style: [example](https://github.com/DataDog/datadog-operator/blob/main/controllers/datadogagent/feature/clusterchecks/feature_test.go)

apis/datadoghq/v2alpha1/datadogagent_types.go
paulcacheux: Yes it should, fixed in last commit

khewonc (approved):

PR #652, opened 2022-11-29
Fix to handle partial image overrides input correctly

levan-m:
### What does this PR do?

Proposes a fix for an issue report in this Slack message - summarizing below.

When `override.nodeAgent.image` isn't fully provided the image URI in agent pod/deployment is corrupt. 
For
```
  override:
    nodeAgent:
      image:
        tag: 7.41.0-rc.5
```
Generated image URI is
```
    image: gcr.io/datadoghq/:7.41.0-rc.5
```

Same happens with DCA and CCR as well.

### Motivation

Fix will allow users to provide only partial overrides and use previuos defaults.

### Additional Notes

n/a

### Describe your test plan

* Added unit test for missing scenarios.
* Tested locally with default operator setup and combinations of overrides, making sure final image URI is as expected.

### DIFF
--- a/controllers/datadogagent/override/podtemplatespec_test.go
+++ b/controllers/datadogagent/override/podtemplatespec_test.go
@@ -43,7 +43,7 @@ func TestPodTemplateSpec(t *testing.T) {
 		{
 			name: "override image name and tag",
 			existingManager: func() *fake.PodTemplateManagers {
-				return fakePodTemplateManagers("docker.io/datadog/agent:7.38.0", t)
+				return fakePodTemplateManagersWithImageOverride("docker.io/datadog/agent:7.38.0", t)
 			},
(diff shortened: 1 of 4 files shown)


controllers/datadogagent/override/podtemplatespec_test.go
```diff
@@ -443,3 +514,21 @@ func TestPodTemplateSpec(t *testing.T) {
 		})
 	}
 }
+
+func fakePodTemplateManagers(image string, t *testing.T) *fake.PodTemplateManagers {
```
CharlyF, line 518: maybe you could rename that method to reflect that it's dedicated to give a podTemplateManager with images overidden?

levan-m: Renamed

CharlyF (approved):

PR #664, opened 2022-12-14
Levan m/update image override logic

levan-m:
…update documentation (related to #652).

### What does this PR do?

This is a follow-up fix of #652 to take full image name (`name:tag` or `repo/name:tag`) in `image.name`  verbatim and not apply any override logic.

### Motivation

#652 introduced logic to use default registry for certain inputs including full image names which wasn't right/documented approach.

### Additional Notes

`common.go`, `common_test.go` are reverted to pre-#652 version.

### Describe your test plan


hestonhoffman (approved):

CharlyF (approved):

apis/datadoghq/common/common.go
```diff
@@ -54,20 +51,11 @@ func GetDefaultReadinessProbe() *corev1.Probe {
...
-		// If the image name corresponds to a full URI we return it, otherwise
-		// we prefix name with passed registry or default one if former is empty/nil.
-		if len(strings.Split(imageSpec.Name, "/")) > 2 {
-			return imageSpec.Name
-		} else if registry != nil && *registry != "" {
-			return fmt.Sprintf("%s/%s", *registry, imageSpec.Name)
-		} else {
-			return fmt.Sprintf("%s/%s", DefaultImageRegistry, imageSpec.Name)
-		}
+		return imageSpec.Name
```
davidor, line 54: This function is also called from code used in `controllers/datadogagent` which I believe is still called in operator v1.0.0 with V2 disabled. I'm not sure if this breaks something, just wanted to mention it just in case.

controllers/datadogagent/override/podtemplatespec_test.go
```diff
@@ -41,9 +40,9 @@ func TestPodTemplateSpec(t *testing.T) {
 			},
 		},
 		{
-			name: "override image name and tag",
+			name: "01: given URI, override image name, tag, JMX",
```
davidor, line 43: nit: I'm not sure about numbered tests. When tests are updated you need to make sure that everything is still sorted. In my experience this can result in missing numbers, unsorted lists, etc.

apis/datadoghq/common/common.go
levan-m: This change basically reverts what was added in #652, so technically only change is added empty string check on line L59. Which I assume is needed since `registry` passed from [podtemplatespec](https://github.com/DataDog/datadog-operator/pull/664/files#diff-822a6227b8c46f98a0c6a5cb578d4439d84b09c84f332f28bfcfdd9a980d5aaaR152) won't be `nil`.

davidor (approved):

controllers/datadogagent/override/podtemplatespec_test.go
levan-m: I totally agree that this requires renumbering and did it couple of times, but with over dozen somewhat similar tests it made debugging and navigating in the IDE easier and found the additional effort worth it. For most other tests numbering is likely overkill but for this it was worth additional effort so I'd rather keep it.

davidor: :+1:

PR #673, opened 2022-12-25
Update policy/v1beta1 to policy/v1

### LINKED ISSUE #619: policy/v1beta1 no longer available in K8S 1.25
```
{"level":"ERROR","ts":"2022-09-28T19:38:58Z","logger":"setup","msg":"Problem running manager","error":"failed to wait for datadogagent caches to sync: no matches for kind \"PodDisruptionBudget\" in version \"policy/v1beta1\""}
```

**Describe what happened:**
The operator is failing due to a deprecated API version.

The operator to start

Helm install operator on a 1.25.x cluster

**Additional environment details (Operating System, Cloud provider, etc):**
KOPS deployed cluster on AWS

This appears to be the only non `v1` used in the code base.  The `v1` has been available since 1.21, and that version of K8S is already marked as EOL.  It does mean limiting the backward support of the operator to some extent.

adammck:
### What does this PR do?

This PR updates references to the `k8s.io/api/policy/v1beta1` package to `/v1`, and (in order to perform that update) removes references to `v1beta1.PodSecurityPolicy`, which was [deprecated](https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/) in k8s 1.21, and was [removed](https://kubernetes.io/blog/2022/08/04/upcoming-changes-in-kubernetes-1-25/) in 1.25.

Fixes #619.

### Motivation

I tried to use this operator to install agents in my cluster, and it didn't work. It said:

```
{
  "level": "ERROR",
  "ts": "2022-12-25T02:22:15Z",
  "logger": "controller-runtime.source",
  "kind": "PodDisruptionBudget.policy",
  "error": "no matches for kind \"PodDisruptionBudget\" in version \"policy/v1beta1\"",
  "msg": "if kind is a CRD, it should be installed before calling Start"
}
```

PodDisruptionBudget [graduated to stable](https://kubernetes.io/blog/2021/04/08/kubernetes-1-21-release-announcement/#graduated-to-stable) in 1.21, and was [removed](https://kubernetes.io/blog/2022/08/04/upcoming-changes-in-kubernetes-1-25/) from the v1beta1 api in 1.25, which I assume is when this problem started. Upgrading the package fixes the problem.

### Additional Notes

The current release of this operator doesn't work at all with the latest kube release (1.25), so even if you don't like my PR, please do something 🙏

### Describe your test plan

Caveat emptor, but I ran the tests 🤷🏾‍♀️

### DIFF
--- a/LICENSE-3rdparty.csv
+++ b/LICENSE-3rdparty.csv
@@ -323,7 +323,6 @@ core,"k8s.io/api/node/v1",Apache-2.0
 core,"k8s.io/api/node/v1alpha1",Apache-2.0
 core,"k8s.io/api/node/v1beta1",Apache-2.0
 core,"k8s.io/api/policy/v1",Apache-2.0
-core,"k8s.io/api/policy/v1beta1",Apache-2.0
 core,"k8s.io/api/rbac/v1",Apache-2.0
 core,"k8s.io/api/rbac/v1alpha1",Apache-2.0
 core,"k8s.io/api/rbac/v1beta1",Apache-2.0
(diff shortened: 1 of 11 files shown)


levan-m:
Thanks for submitting the PR! We can't merged it right now because:
* We are adding support for both versions of PDB, so Operator can work with Kubernetes versions <1.21.
* We are investigating whether PSP is needed at all and whether it can be removed from Operator without any impact. If this can be done we can use this (updated) PR to remove PSP. Otherwise we will have to limit use of PSP to only Kubernetes <1.25.

Kerwood:
Whats the status on this PR ?
This is much needed.

levan-m:
We merged two changes #683, #688 addressing this issue in `v1.0.0` and they'll be included in the next RC.

I'll close this PR. Please use https://github.com/DataDog/datadog-operator/issues/619 for further communication about this issue.

PR #683, opened 2023-01-06
Manage either `v1` or `v1beta1` PDB based on current Kubernetes API version

levan-m:
### What does this PR do?

* Support both `v1` and `v1beta1` version of PDB in the Operator.
* Based on Kubernetes API version information determine which version of PDB to create.
PDBs are used only with v1 controller now.
* With this change operator will use PDB `v1` for Kubernetes >=1.21; and `v1beta1` for earlier versions.

### Motivation

Operator used `v1beta1` version which isn't available in 1.25 preventing Operator to run on newer Kubernetes versions.

### Additional Notes

Some implementation details:
* We have a new struct `PlatformInfo` which has `versionInfo` and two maps for 'preferred' and 'other' versions. Current PDB selection opts for `preferred` version.
* It's added to `Reconciler`, `DatadogAgentReconciler` and `Store` structs.
* `PlatformInfo` provides a boolean check if PDB is `v1beta1`; it also provides functions to create specific version of PDB `client.Object/List`.
* `PlatformInfo` is created in `setup.go`. We are reusing existing discovery client to extract API version - mostly used this logic from DCA [code](https://github.com/DataDog/datadog-agent/blob/main/pkg/collector/corechecks/cluster/orchestrator/discovery/collector_discovery.go#L46-L91).
* Updated `poddisruptionbudget.go` to manage either `v1` or `v1beta1` resources based on the `PlatformInfo`.
* Updated `controller_test.go` to run PDB tests for both PDB versions; defaulting all tests to PDB `v1` except those targeting `v1beta1` tests.

### Describe your test plan
* Tested on 1.20, 1.24, 1.25 version of Kubernetes running on Kind cluster. 
* Tried both agent CRD `v2alpha1` and `v1alpha1` with v2 and v1 controllers respectively.
* Confirmed when using v1/`v2alpha1` and DCA replicas >1, PDB is created in 1.20, 1.24 with versions `v1beta1` and `v1` respectively. 
* 1.25 isn't isn't working since with this revision since it's blocked by PSP; After removing PSP confirmed v1/`v2alpha1` creates PDB `v1`.
* v2/`v2alpha1` doesn't create PDB, so only did sanity runs to confirm no regression happens.

### DIFF
--- a/controllers/datadogagent/poddisruptionbudget.go
+++ b/controllers/datadogagent/poddisruptionbudget.go
@@ -9,6 +9,7 @@ import (
 	"context"
 
 	"github.com/go-logr/logr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
@@ -52,13 +53,12 @@ func (r *Reconciler) manageClusterChecksRunnerPDB(logger logr.Logger, dda *datad
 }
 
 func (r *Reconciler) managePDB(logger logr.Logger, dda *datadoghqv1alpha1.DatadogAgent, pdbName string, builder pdbBuilder, builderv1beta1 pdbV1Beta1Builder, cleanUp bool) (reconcile.Result, error) {
-	if cleanUp {
-		return r.cleanupPDB(dda, pdbName)
-	}
-
 	useV1Beta1 := r.platformInfo.UseV1Beta1PDB()
 	if useV1Beta1 {
 		pdbV1Beta1 := &policyv1beta1.PodDisruptionBudget{}
+		if cleanUp {
+			return r.cleanupPDB(dda, pdbName, pdbV1Beta1)
+		}
 		err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: dda.Namespace, Name: pdbName}, pdbV1Beta1)
 		if err != nil {
 			if errors.IsNotFound(err) {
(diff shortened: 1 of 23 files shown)


controllers/datadogagent/poddisruptionbudget.go
```diff
@@ -33,37 +35,74 @@ type (
 	pdbBuilder func(dda *datadoghqv1alpha1.DatadogAgent) *policyv1.PodDisruptionBudget
 )
 
+type (
+	pdbV1Beta1Builder func(dda *datadoghqv1alpha1.DatadogAgent) *policyv1beta1.PodDisruptionBudget
```
vboulineau, line 39: Both should be in a single `type ()` or not use `()`

vboulineau (approved):
LGTM, only small comments, can be merged

controllers/datadogagent/poddisruptionbudget.go
```diff
@@ -33,37 +35,74 @@ type (
...
+	return r.managePDB(logger, dda, getClusterChecksRunnerPDBName(dda), buildClusterChecksRunnerPDBV1, buildClusterChecksRunnerPDBV1Beta1, cleanUpCondition)
 }
 
-func (r *Reconciler) managePDB(logger logr.Logger, dda *datadoghqv1alpha1.DatadogAgent, pdbName string, builder pdbBuilder, cleanUp bool) (reconcile.Result, error) {
-	if cleanUp {
-		return r.cleanupPDB(dda, pdbName)
-	}
+func (r *Reconciler) managePDB(logger logr.Logger, dda *datadoghqv1alpha1.DatadogAgent, pdbName string, builder pdbBuilder, builderv1beta1 pdbV1Beta1Builder, cleanUp bool) (reconcile.Result, error) {
+	useV1Beta1 := r.platformInfo.UseV1Beta1PDB()
+	if useV1Beta1 {
```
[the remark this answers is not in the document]
levan-m: This is a good call-out and I agree with you! I anticipate refactor of `poddisruptionbudget.go` when we implement PDBs in v2 since current implementation takes `v1alpha1` CRD, that refactor will take care of two version of PDB as well.

controllers/datadogagent/poddisruptionbudget.go
levan-m: Fixed.

PR #688, opened 2023-01-17
Avoid PodSecurityPolicy cleanup if the resource is not supported by Kubernetes API

levan-m:
### What does this PR do?
This change addresses the Kubernetes 1.25 compatibility issue caused by Operator attempting to handle `PodDisruptionPolicy` resource not available in aforementioned version. 

Operator 1.0.0 deployed on v1.25 would log following error on every reconcile
```
{"level":"ERROR","ts":"2023-01-17T21:19:23Z","logger":"controller.datadogagent","msg":"Reconciler error","reconciler group":"datadoghq.com","reconciler kind":"DatadogAgent","name":"datadog","namespace":"system","error":"no matches for kind \"PodSecurityPolicy\" in version \"policy/v1beta1\"","errorCauses":[{"error":"no matches for kind \"PodSecurityPolicy\" in version \"policy/v1beta1\""}]}
```

### Motivation

Make Operator 1.0.0 run on v1.25.

### Additional Notes

This change ***does not*** solve PSP usage across Operator. 

PSP resource isn't utilized/implemented (see [this](https://github.com/DataDog/datadog-operator/blob/main/controllers/datadogagent/merger/pod_security.go#L121-L134)) and only places attempting to handle PSP were resource cleanup routines. This PR changes the logic producing list of Kinds for the cleanup - it includes PSP only if it's supported by current Kubernetes version.

Once PSP resource implementation is done additional work will be need to control PSP creation. 

### Describe your test plan

Tested Kubernetes version 1.24, 1.25 using Kind cluster. Deployed agent locally, installed DDA v2 with DCA, CCR enabled - confirmed no errors are thrown.

### DIFF
--- a/pkg/kubernetes/platforminfo.go
+++ b/pkg/kubernetes/platforminfo.go
@@ -96,7 +96,7 @@ func (platformInfo *PlatformInfo) supportsPSP() bool {
 	if platformInfo.apiOtherVersions == nil || platformInfo.apiPreferredVersions == nil {
 		return true
 	}
-	_, ok1 := platformInfo.apiOtherVersions["PodSecurityPolicy"]
-	_, ok2 := platformInfo.apiPreferredVersions["PodSecurityPolicy"]
-	return ok1 || ok2
+	_, otherExists := platformInfo.apiOtherVersions["PodSecurityPolicy"]
+	_, preferredExists := platformInfo.apiPreferredVersions["PodSecurityPolicy"]
+	return otherExists || preferredExists
 }
--- a/pkg/kubernetes/platforminfo.go
+++ b/pkg/kubernetes/platforminfo.go
@@ -87,3 +87,16 @@ func (platformInfo *PlatformInfo) CreatePDBObjectList() client.ObjectList {
 		return &policyv1.PodDisruptionBudgetList{}
 	}
 }
+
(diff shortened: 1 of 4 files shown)


pkg/kubernetes/platforminfo.go
```diff
@@ -87,3 +87,16 @@ func (platformInfo *PlatformInfo) CreatePDBObjectList() client.ObjectList {
...
+func (platformInfo *PlatformInfo) GetAgentResourcesKind(withCiliumResources bool) []ObjectKind {
+	return getResourcesKind(withCiliumResources, platformInfo.supportsPSP())
+}
+
+func (platformInfo *PlatformInfo) supportsPSP() bool {
+	if platformInfo.apiOtherVersions == nil || platformInfo.apiPreferredVersions == nil {
+		return true
+	}
+	_, ok1 := platformInfo.apiOtherVersions["PodSecurityPolicy"]
+	_, ok2 := platformInfo.apiPreferredVersions["PodSecurityPolicy"]
```
CharlyF, line 100: nit: can we name these `otherExists` and `preferredExists` or something clearer than ok?

CharlyF (approved):

pkg/kubernetes/platforminfo.go
levan-m: Updated.

PR #697, opened 2023-01-30
Add plugin ARM/CGO release tooling

clamoriniere:
### What does this PR do?

This PR modify the kubectl plugin release process to:
* add ARM architecture support.
* CGO compilation

inspired by: https://github.com/DataDog/extendeddaemonset/pull/153

### Motivation

Support ARM user and more kubectl authentication mechanism.

### Additional Notes

### Describe your test plan

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,61 @@
+name: build
+
+on:
+  push:
+
+jobs:
+  build-linux-binary:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: gcc install
+        run: sudo apt install gcc-aarch64-linux-gnu
+      - name: Get tag
+        uses: little-core-labs/get-git-tag@v3.0.2
+        id: tag
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.18
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v3
+        with:
+          version: latest
+          args: build --skip-validate --config .goreleaser-for-linux.yaml
+  build-darwin-binary:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Get tag
+        uses: little-core-labs/get-git-tag@v3.0.2
+        id: tag
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.17
+      - name: Build
+        uses: goreleaser/goreleaser-action@v3
+        with:
+          args: build --skip-validate --config .goreleaser-for-darwin.yaml
+  build-windows-binary:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Get tag
+        uses: little-core-labs/get-git-tag@v3.0.2
+        id: tag
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.17
+      - name: Build
+        uses: goreleaser/goreleaser-action@v3
+        with:
+          args: build --skip-validate --config .goreleaser-for-windows.yaml
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,4 @@
-name: build
+name: Build
 
 on:
   push:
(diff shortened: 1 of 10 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,61 @@
+name: build
```
CharlyF, line 1:
```suggestion
name: Build
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,61 @@
...
+
+on:
+  push:
```
CharlyF, line 4: should we leave that if it's empty?

.github/workflows/build.yml
```diff
@@ -0,0 +1,61 @@
...
+          fetch-depth: 0
+      - name: gcc install
+        run: sudo apt install gcc-aarch64-linux-gnu
+      - name: Get tag
+        uses: little-core-labs/get-git-tag@v3.0.2
+        id: tag
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.18
```
CharlyF, line 22: any reason why we use 1.18 and not 1.17 like in the darwin and windows?

hack/release/generate-plugin-manifest.sh
```diff
@@ -18,18 +18,27 @@ PLUGIN_NAME=kubectl-datadog
...
 
-DARWIN_AMD64=$(grep $PLUGIN_NAME $OUTPUT_FOLDER/datadog-operator_${VERSION}_checksums.txt  | grep "darwin_amd64" | awk '{print $1}')
-WINDOWS_AMD64=$(grep $PLUGIN_NAME $OUTPUT_FOLDER/datadog-operator_${VERSION}_checksums.txt  | grep "windows_amd64" | awk '{print $1}')
-LINUX_AMD64=$(grep $PLUGIN_NAME $OUTPUT_FOLDER/datadog-operator_${VERSION}_checksums.txt  | grep "linux_amd64" | awk '{print $1}')
+DARWIN_AMD64=$(grep $PLUGIN_NAME $OUTPUT_FOLDER/checksums.txt  | grep "darwin_amd64" | awk '{print $1}')
+WINDOWS_AMD64=$(grep $PLUGIN_NAME $OUTPUT_FOLDER/checksums.txt  | grep "windows_amd64" | awk '{print $1}')
+LINUX_AMD64=$(grep $PLUGIN_NAME $OUTPUT_FOLDER/checksums.txt  | grep "linux_amd64" | awk '{print $1}')
+DARWIN_ARM64=$(grep $PLUGIN_NAME $OUTPUT_FOLDER/checksums.txt  | grep "darwin_arm64" | awk '{print $1}')
+WINDOWS_ARM64=$(grep $PLUGIN_NAME $OUTPUT_FOLDER/checksums.txt  | grep "windows_arm64" | awk '{print $1}')
+LINUX_ARM64=$(grep $PLUGIN_NAME $OUTPUT_FOLDER/checksums.txt  | grep "linux_arm64" | awk '{print $1}')
```
CharlyF, line 26: you want to ditch `${VERSION}` because you don't expect that the bins will be built on the same machines?

CharlyF (approved):
left a couple questions but thanks for working on this.

.github/workflows/build.yml
clamoriniere: I think it is to have have it enable for any push

.github/workflows/build.yml
clamoriniere: I forgot to update in darwin and windows I will fix it

.github/workflows/release.yaml
```diff
@@ -1,50 +1,161 @@
...
+        id: tag
+      - name: Prepare ./dist folder
+        run: |
+          mkdir -p ./dist
+          cat ./tmp-build/darwin/checksums.txt >> ./dist/checksums.txt
+          cat ./tmp-build/linux/checksums.txt >> ./dist/checksums.txt
+          cat ./tmp-build/windows/checksums.txt >> ./dist/checksums.txt
+          mv ./tmp-build/darwin/*.zip ./dist
+          mv ./tmp-build/linux/*.zip ./dist
+          mv ./tmp-build/windows/*.zip ./dist
```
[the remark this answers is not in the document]
clamoriniere: the artefact in an github release is flat. that is why I added everything at the same level. Maybe it works with sub folder too.

hack/release/generate-plugin-manifest.sh
clamoriniere: it is because the version is note really useful in this context. 
and after we will need to [merge several chechsums.txt files in once](https://github.com/DataDog/datadog-operator/blob/204dffc11f6c404656069bb957f48fee477224dc/.github/workflows/release.yaml#L150-L159), so it was easier to have a "static" file name

PR #730, opened 2023-03-08
Update install-yq.sh to support arm64

clamoriniere:
### What does this PR do?

the current script used to install `yq` doesn't work on arm64. This change use `uname` to know the CPU arch and download the corresponding binary

### Motivation

Fix releasing process on ARM64

### Additional Notes

### Describe your test plan


hack/install-yq.sh
```diff
@@ -26,4 +27,4 @@ curl -Lo ${BINARY} https://github.com/mikefarah/yq/releases/download/$VERSION/$B
 
 chmod +x $BINARY
 mkdir -p $ROOT/bin
-mv $BINARY $ROOT/bin/$PLATFORM/yq
\ No newline at end of file
+mv $BINARY $ROOT/bin/$PLATFORM/yq
```
AliDatadog, line 30:
```suggestion
mv $BINARY $ROOT/bin/$PLATFORM/yq

```

CharlyF (approved):
looks good - Thanks!

PR #765, opened 2023-04-18
Add feature enabled metrics to metrics forwarder

khewonc:
### What does this PR do?

Add a feature enabled metric for all enabled features.

Metric format: `datadog.operator.<feature>.enabled`, e.g. `datadog.operator.ksm.feature.enabled`
Tags:
* `cr_name`: DDA name
* `cr_namespace`: DDA namespace

Value:
* `1.0` if feature enabled
* No metric sent if feature not enabled

### Motivation

### Additional Notes

### Describe your test plan


levan-m:
Regarding 
> No metric sent if feature not enabled

Can we derive percentages if we only record 'enabled' metrics? Does it make sense to record `0.0` as well?

controllers/datadogagent/controller_reconcile_v2.go
```diff
@@ -95,6 +95,8 @@ func (r *Reconciler) reconcileInstanceV2(ctx context.Context, logger logr.Logger
 	newStatus := instance.Status.DeepCopy()
 
 	features, requiredComponents := feature.BuildFeatures(instance, reconcilerOptionsToFeatureOptions(&r.options, logger))
+	// update list of enabled features for metrics forwarder
+	r.updateMetricsForwardersFeatures(instance, features)
```
levan-m, line 99: Should we do this towards the end of the reconcile function? If something is enabled and reconciler starts failing the metrics reported would be inaccurate.

CharlyF (approved):

controllers/datadogagent/controller_reconcile_v2.go
CharlyF: that's a good point - Though I think we could actually keep the submission here to reflect that the feature is enabled, but later on, improve the code to have the value be `0` if for some reason it errors out (either alone or with a global error). WDYT?

levan-m: That's an interesting perspective, that way we would track intent of enabling feature and that's a reasonable thing to do. So, I could go either way.

PR #778, opened 2023-04-28
Clean up dependencies for currently reconciled DDA

khewonc:
### What does this PR do?

Applies cleanup method to dependencies associated with the currently reconciled DDA only to prevent other DDA's dependencies from being deleted.

### Motivation

### Additional Notes

### Describe your test plan

1. Spin up a DDA, delete it, and ensure that any dependencies managed by the store are deleted. ConfigMaps like `xxx-leader-election` aren't created using the store so those will not be affected.
2. Spin up a DDA with the external metrics feature enabled and check that the corresponding service was created. Update the DDA to disable the external metrics feature and check that the service is deleted.


CharlyF (approved):

controllers/datadogagent/dependencies/store.go
```diff
@@ -341,7 +341,7 @@ func (ds *Store) DeleteAll(ctx context.Context, k8sClient client.Client) []error
 	return deleteObjects(ctx, k8sClient, objsToDelete)
 }
 
-func listObjectToDelete(objList client.ObjectList, cacheObjects map[string]client.Object) ([]client.Object, error) {
+func listObjectToDelete(objList client.ObjectList, cacheObjects map[string]client.Object, ddaNs, ddaName string) ([]client.Object, error) {
```
levan-m, line 344: Can we use `ds.owner` instead of passing these additional values?

levan-m (approved):

levan-m (approved):

PR #780, opened 2023-05-01
[datadogmonitor] Update reconcile loop so monitor resource is source of truth

celenechang:
### What does this PR do?

- Update datadog-api-client-go version
- Add `RestrictedRoles` monitor parameter
- Update DatadogMonitor reconciler so that the resource content is the source of truth. That is, if the corresponding monitor in the app drifts (due to a user manually editing in the UI, for instance), then the data will be resynced according to the k8s resource.
- Similarly, if the monitor is deleted in the app but the resource is still present, it will be recreated in the app

### Motivation

Better user experience

### Additional Notes

### Describe your test plan

Ensure the following behave as expected:
- [ ] create monitor in app by applying manifest
- [ ] edit the monitor manifest and apply
- [ ] edit monitor in app; should be resynced within a couple minutes
- [ ] delete monitor locally; should be deleted in the app
- [ ] delete monitor in app; should be recreated
- [ ] add RestrictedRoles parameter to resource


docs/datadog_monitor.md
```diff
@@ -50,7 +50,8 @@ To deploy a `DatadogMonitor` with the Datadog Operator, use the [`datadog-operat
         - "test:datadog"
     ```
 
-    For additional examples, see [examples/datadog-monitor](../examples/datadogmonitor). Note that only metric alerts, query alerts, and service checks are supported.
+    For additional examples, see [examples/datadog-monitor](../examples/datadogmonitor).
+    *Note*: Not all monitor types are supported. See [here](../controllers/datadogmonitor/controller.go) for the current list.
```
maycmlee, line 54:
```suggestion
    **Note**: Not all monitor types are supported. See [here](../controllers/datadogmonitor/controller.go) for the current list.
```

maycmlee (approved):
Just a small edit

controllers/datadogmonitor/controller_test.go
```diff
@@ -939,3 +940,58 @@ func testAuditMonitor() *datadoghqv1alpha1.DatadogMonitor {
 		},
 	}
 }
+
+func Test_checkMonitorDefSync(t *testing.T) {
+	now := time.Unix(1612244495, 0)
+	metaNow := metav1.NewTime(now)
+	fiveMinAgo := now.Add(-time.Minute * 5)
+	metaFiveMinAgo := metav1.NewTime(fiveMinAgo)
+	m := genericMonitor(12345)
```
levan-m, line 949: nit - `1612244495` is set in this function too, this is not obvious from the diff.

controllers/datadogmonitor/controller.go
```diff
@@ -224,36 +242,47 @@ func (r *Reconciler) update(logger logr.Logger, datadogMonitor *datadoghqv1alpha
...
 		status.SyncStatus = datadoghqv1alpha1.SyncStatusGetError
-		return err
+		return m, err
 	}
+	return m, nil
+}
 
+// Helper functions to modify an API monitor object
+func checkMonitorDefSync(m datadogapiclientv1.Monitor, status *datadoghqv1alpha1.DatadogMonitorStatus) bool {
+	monitorModifiedTime := m.GetModified()
```
levan-m, line 276: Why not compare monitor parameters (just 6, [here](https://github.com/DataDog/datadog-operator/blob/19b5666dbcbf3dfc1add46f2306b9da48a73da95/controllers/datadogmonitor/monitor.go#L24-L29)) for equality instead of relying on modification time (and window)?

apis/datadoghq/v1alpha1/datadogmonitor_types.go
```diff
@@ -20,6 +20,10 @@ type DatadogMonitorSpec struct {
 	Priority int64 `json:"priority,omitempty"`
 	// Query is the Datadog monitor query
 	Query string `json:"query,omitempty"`
+	// RestrictedRoles is a list of unique role identifiers to define which roles are allowed to edit the monitor.
+	// `restricted_roles` is the successor of `locked`. For more information about `locked` and `restricted_roles`,
+	// see the [monitor options docs](https://docs.datadoghq.com/monitors/guide/monitor_api_options/#permissions-options).
+	RestrictedRoles []string `json:"restricted_roles,omitempty"`
```
levan-m, line 26: nit - too late not but splitting this seemingly unrelated changes into two PRs would make reviews faster

celenechang:
closing this PR in favor of a new one with a different approach

PR #791, opened 2023-05-19
[datadogmonitor] add periodic forced sync of API monitor

celenechang:
### What does this PR do?

Add periodic sync (currently defaulted to 1h) of monitor to ensure that definition of API monitor does not drift from DatadogMonitor resource, which is considered the source of truth for Operator-managed monitors.

Refactor code so that if a monitor is deleted, the operator will recreate it.

Rename `SyncStatus*` to `MonitorSyncStatus*` for clarity.

### Motivation

Better user experience

### Additional Notes

### Describe your test plan

- [ ] create monitor in app by applying manifest
- [ ]  edit the monitor manifest and apply
- [ ]  edit monitor in app; should be resynced within a couple minutes
- [ ]  delete monitor locally; should be deleted in the app
- [ ]  delete monitor in app; should be recreated

### DIFF
--- a/controllers/datadogmonitor/controller.go
+++ b/controllers/datadogmonitor/controller.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -36,6 +37,7 @@ import (
 const (
 	defaultRequeuePeriod    = 60 * time.Second
 	defaultErrRequeuePeriod = 5 * time.Second
+	defaultForceSyncPeriod  = 60 * time.Minute
 	maxTriggeredStateGroups = 10
 )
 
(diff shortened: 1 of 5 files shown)


controllers/datadogmonitor/controller.go
```diff
@@ -125,55 +127,66 @@ func (r *Reconciler) internalReconcile(ctx context.Context, req reconcile.Reques
...
+
+	// Check if we need to create the monitor, update the monitor definition, or update monitor state
 	if instance.Status.ID == 0 {
-		logger.V(1).Info("Monitor ID is not set; creating monitor in Datadog")
-		// If the monitor ID is 0, then it doesn't exist yet in Datadog. Create the monitor (only metric alerts)
+		shouldCreate = true
+	} else {
+		var m datadogapiclientv1.Monitor
+		if instanceSpecHash != statusSpecHash {
+			// Custom resource manifest has changed, need to update the API
```
fanny-jiang, line 139: It would be nice to add a log here indicating that a diff has been detected

levan-m (approved):

PR #793, opened 2023-05-19
[datdogmonitor] add restricted roles

celenechang:
### What does this PR do?

- Update datadog-api-client-go version
- Add RestrictedRoles monitor parameter

### Motivation

Better user experience

### Additional Notes

### Describe your test plan

Create a DatadogMonitor with the RestrictedRoles parameter, for instance:

```
kind: DatadogMonitor
metadata:
  name: datadog-monitor-test
spec:
  query: "avg(last_10m):avg:system.disk.in_use{*} by {host} > 0.9"
  restrictedRoles:
    - "<UUID>"
  type: "metric alert"
  name: "Test another monitor made from DatadogMonitor"
  message: "We are running out of disk space!"
  ```

### CODE REVIEW

File: controllers/datadogmonitor/controller_test.go [hunk 1]
```diff
@@ -542,7 +542,7 @@ func Test_convertStateToStatus(t *testing.T) {
 				}
 
 				m.State = &datadogapiclientv1.MonitorState{
-					Groups: &msg,
+					Groups: msg,
```
levan-m, line 545: Is this due to minor version change in `datadog-api-client-go`?

levan-m (approved):

PR #807, opened 2023-06-04
Adding Service Level Objective CRD and Controller

### LINKED ISSUE #684: Service Level Objectives CRD
I see in [this old ticket](https://github.com/DataDog/datadog-operator/issues/4) a request for Datadog operator handling custom resources (e.g. Monitor, Dashboard etc). I also see that [support for monitors has been added recently](https://github.com/DataDog/datadog-operator/tree/main/examples/datadogmonitor). Thanks for that feature.

I wanted to understand if there is a plan to also support a CRD for SLOs in the operator roadmap? SLOs are quite important for us and we would love to maintain SLO configuration close to rest of our infrastructure configuration (provisioned through kube manifests).

Any feedback here is highly appreciated.

jurabek:
### What does this PR do?

This pull request introduces the `DatadogSLO CRD` types and controller, addressing issue #684.

### Motivation
As each organization scales, they gonna rely on many `SLOs`, each of the SLO's managed by individual teams. This situation has led to complexity in maintaining these `SLOs` with us as well. The introduction of the `DatadogSLO CRD` types and controller into `datadog-operator` aims to simplify this process, enabling us to create and manage our `SLOs` and `Monitors` using our `GitOps` tools. This feature is crucial as we scale and manage a growing number of SLOs across various teams.

### Additional Notes
- The development of this feature was inspired by the `DatadogMonitor CRD` codebase, which has proven effective in similar use cases.
- Shared Finalizer: A shared finalizer has been added to enhance code reusability.
- Simplified Updating Status Conditions: The condition logic has been simplified and now uses Kubernetes utilities for better efficiency and code coverage.

### Describe your test plan

- `kind cluster create`
- `make buid`
- `make IMG=test/operator:test docker-build`
- `kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.8.0/cert-manager.yaml `
-  `kind load docker-image test/operator:test`
- `make IMG=test/operator:test deploy`
- `kubectl apply -f examples/datadogslo/metric-example.yaml`

```
kubectl get datadogslos -n system
NAME          ID                                 SYNC STATUS   AGE
example-slo   21e10346c75c5852a12214d1d89dd2c2   OK            2m53s
```

result in DD:
image


Sturgelose:
A question (though I can imagine that the delay of the merging of this PR might be related to this):

If we create an SLO based on a monitor or we want to create monitors on top of the SLO (burn rate and threshold crossed), this requires referring to a monitor_id.

How could this be possible in Helm or K8s?
This will require somehow referencing another k8s object and.... I cannot come with a proper solution to do so with the currently proposed PR.

Either, we embed the declaration of the monitor spec inside of the SLO CRD, or instead of referencing via ID, we can provide labels or a name, and the operator can figure out which monitor should link it to from locally created monitors in this very same cluster?

jurabek:
> Either, we embed the declaration of the monitor spec inside of the SLO CRD, or instead of referencing via ID, we can provide labels or a name, and the operator can figure out which monitor should link it to from locally created monitors in this very same cluster?

I think this should not be part of the operator, IMHO kubernetes operators are not dependency resolvers like Terraform does, I think you should solve your dependencies before it applied into operator.

I think only terraform will solve this kind of problems out of the box.

jurabek:
@Sturgelose 
we solved similar problem with our auto generated secrets using `Argo + Helm + Kubernetes ConfigMaps` the idea was simple, since we could benefit from Argo retries that when resource not exists it retries and we used Helm `_helper.tpl` to create cron jobs to sync values between two resources like:

```
# Helm chart for ResourceA
apiVersion: v1
kind: ResourceA
metadata:
  name: resourceA
 # uid: this field will be computed after applying
---
# Kubernetes job to create a ConfigMap with ResourceA's ID
apiVersion: batch/v1
kind: Job
metadata:
  name: job
spec:
  template:
    spec:
      containers:
      - name: job
        image: bitnami/kubectl
        command: ["/bin/sh", "-c"]
        args:
        - >
          ID=$(kubectl get ResourceA resourceA -o jsonpath='{.metadata.uid}');
          kubectl create configmap resourcea-id --from-literal=id=$ID
      restartPolicy: OnFailure

---
# Helm chart for ResourceB
apiVersion: v1
kind: ResourceB
metadata:
  name: resourceB
spec:
  externalResourceID: 
    valueFrom:
      configMapKeyRef:
        name: resourcea-id
        key: id


```

melchiormoulin:
As a datadog client, this would be a nice feature, can we help to make it merged ?

celenechang:
Hi @jurabek thank you for this contribution and apologies for the delay. Overall it looks good, do you mind if I make some edits on your branch?

jurabek:
@celenechang yeas feel free to do it 🙏

jurabek:
> Hi @jurabek thank you for this contribution and apologies for the delay. Overall it looks good, do you mind if I make some minor edits on your branch?

@celenechang yeas feel free to do it 🙏

rtrieu:
Created DOCS-6131 for docs review.

cswatt (commented):
doesn't look like this will need any documentation review 👍

celenechang:
Hi @jurabek , just dropping in to let you know this is actively being worked on. I hope to have another update soon.

zhengshizhao:
looks like the validation tests need to be updated to reflect the last change on validation. but overall LGTM

zhengshizhao (approved):
overall lgtm.

apis/datadoghq/v1alpha1/datadogslo_types.go
```diff
@@ -0,0 +1,180 @@
...
+	Timeframe DatadogSLOTimeFrame `json:"timeframe"`
+
+	// TargetThreshold is the target threshold such that when the service level indicator is above this threshold over the given timeframe, the objective is being met.
+	TargetThreshold resource.Quantity `json:"targetThreshold"`
+
+	// WarningThreshold is a optional warning threshold such that when the service level indicator is below this value for the given threshold, but above the target threshold, the objective appears in a "warning" state. This value must be greater than the target threshold.
+	WarningThreshold *resource.Quantity `json:"warningThreshold,omitempty"`
+
+	// ControllerOptions are the optional parameters in the DatadogSLO controller
+	ControllerOptions DatadogSLOControllerOptions `json:"controllerOptions,omitempty"`
```
CharlyF, line 56: given the options, shoudn't this be a ptr?

config/crd/patches/webhook_in_datadogslos.yaml
```diff
@@ -0,0 +1,17 @@
...
+  conversion:
+    strategy: Webhook
+    webhook:
+      clientConfig:
+        service:
+          namespace: system
+          name: webhook-service
+          path: /convert
+      conversionReviewVersions:
+      - v1
```
CharlyF, line 17: maybe we should only introduce this if we introduce a new apiversion?

controllers/datadogslo/controller.go
```diff
@@ -0,0 +1,308 @@
...
+		datadogAuth:   ddClient.Auth,
+		versionInfo:   versionInfo,
+		log:           log,
+		recorder:      recorder,
+	}
+}
+
+var _ reconcile.Reconciler = (*Reconciler)(nil)
+
+func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconcile.Result, error) {
```
CharlyF, line 69: why not passing the context? a todo is used instead in the `internalReconcile`. Could be useful for tracing or error/timeout handling.

controllers/datadogslo/controller.go
```diff
@@ -0,0 +1,308 @@
...
+	} else {
+		if instanceSpecHash != statusSpecHash {
+			shouldUpdate = true
+		} else if instance.Status.LastForceSyncTime == nil || (defaultForceSyncPeriod-now.Sub(instance.Status.LastForceSyncTime.Time)) <= 0 {
+			// Periodically force a sync with the API SLO to ensure parity
+			// Get SLO to make sure it exists before trying any updates. If it doesn't, set shouldCreate
+			_, err = r.get(instance)
+			if err != nil {
+				logger.Error(err, "error getting SLO", "SLO ID", instance.Status.ID)
+				if strings.Contains(err.Error(), "404 Not Found") {
```
CharlyF, line 132: should be a variable

controllers/datadogslo/controller.go
```diff
@@ -0,0 +1,308 @@
...
+
+var _ reconcile.Reconciler = (*Reconciler)(nil)
+
+func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconcile.Result, error) {
+	res, err := r.internalReconcile(req)
+	return res, err
+}
+
+func (r *Reconciler) internalReconcile(req reconcile.Request) (reconcile.Result, error) {
+	logger := r.log.WithValues("datadogslo", req.NamespacedName)
```
CharlyF, line 75: Seems like the controller has a logger already with a slightly different name https://github.com/DataDog/datadog-operator/pull/807/files#diff-32a274b974ae70f092b3f061d72858382354b8f710d468d80883475162e3920aR31

pkg/datadogclient/client.go
```diff
@@ -36,6 +36,43 @@ func InitDatadogMonitorClient(logger logr.Logger, creds config.Creds) (DatadogMo
...
+
+// InitDatadogSLOClient initializes the Datadog SLO API Client and establishes credentials.
+func InitDatadogSLOClient(logger logr.Logger, creds config.Creds) (DatadogSLOClient, error) {
+	if creds.APIKey == "" || creds.AppKey == "" {
+		return DatadogSLOClient{}, errors.New("error obtaining API key and/or app key")
+	}
+
+	configV1 := datadogapi.NewConfiguration()
+	apiClient := datadogapi.NewAPIClient(configV1)
+	client := datadogV1.NewServiceLevelObjectivesApi(apiClient)
```
CharlyF, line 65: What's the `auth` for out of curiosity?
Could we not use a generic client (for metrics, monitors and SLOs?)

fanny-jiang (commented):
I just had some small documentation suggestions. Still validating the controller behavior manually but looks good so far.

yury-komarov-contextlabs:
raising up PR
really important CRD

examples/datadogslo/metric-example.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: datadoghq.com/v1alpha1
+kind: DatadogSLO
+metadata:
+  name: example-slo
+  namespace: system 
+spec:
+  name: example-slo
+  description: "This is an example SLO from datadog-operator"
+  query:
+    denominator: "sum:requests.total{service:example,env:prod}.as_count()"
```
ddjamesfrullo, line 10: What if we use `numerator: "sum:universal.http.server.hits{service:example, env:prod, !http.status_class:5xx}.as_count()"` and `denominator: sum:universal.http.server.hits{service:example, env:prod}.as_count()"` instead of `requests.total`  and `request.success` as the default metrics? It's a bit more complicated, but if they have universal service monitoring, at least that stands a chance of working when they replace it with their service. If they don't have universal service monitoring, then at least we're advertising a Datadog product. Not a big deal to keep these defaults, though.

ddjamesfrullo:
I read the comments here about how to handle dependency. I think without dependency handling, monitor SLOs are not really useful to have here. At the same time, I'm happy with that because we want to move away from monitor SLOs. We're about to release a new type called time-slice SLOs that will allow you to create an uptime SLO directly from metrics. We can talk about how to add that here when the time is right. About creating SLO monitors (like burn rate and error budget monitors), it's an interesting idea to specify those monitors along with SLO details at creation time. Would it make sense, then, to hit both the SLO API and the monitor API in the implementation of the operator? Or were you thinking about having our SLO create API allow for monitor creation? I would prefer the latter.

examples/datadogslo/metric-example.yaml
celenechang: I think we could benefit from having another example or two - what do you think about adding your query suggestion in another file `metric-usm-example.yaml` ?

celenechang:
@ddjamesfrullo regarding the creation of SLO monitors, maybe we can discuss IRL future plans for the SLO API and if it makes sense to incorporate this behavior in the operator

controllers/datadogslo/controller.go
celenechang: yea I noticed that; I'm not entirely sure why we have this pattern of two loggers but it's present in the DatadogAgent controller as well. when testing I found this logger was a bit more informative to use since it also records the resource name and namespace

pkg/datadogclient/client.go
celenechang: we talked IRL, but changes in the new client version have resulted in separate clients for these APIs

examples/datadogslo/metric-example.yaml
celenechang: @ddjamesfrullo  I added another metric example using the query you suggested, as well as a monitor example.

fanny-jiang (approved):

celenechang:
@jurabek thanks again for opening this PR. It will be part of the 1.3.0 release which is planned to start in a few weeks.

PR #826, opened 2023-06-27
Set admission controller communication mode based on apm and dsd

khewonc:
### What does this PR do?

Automatically configures the admission controller inject config mode based on APM and DSD feature settings.
- if `features.admissionController.agentCommunicationMode` set, use that value
- if either APM or DSD uses the UDS socket, use `socket` mode
- otherwise, don't set `DD_ADMISSION_CONTROLLER_INJECT_CONFIG_MODE` since it will default to using `hostip` mode in the agent code

### Motivation

### Additional Notes

### Describe your test plan

Test that the `agentCommunicationMode` can be set manually, that `socket` mode is used when a UDS socket is used, and that the DCA uses `hostip` when a UDS socket isn't used. The following manifest snippet has examples of relevant fields that can be configured to test:

```yaml
  features:
    admissionController:
      enabled: true
      mutateUnlabelled: true
      agentCommunicationMode: service
    apm:
      enabled: true
      hostPortConfig:
        enabled: true
      unixDomainSocketConfig:
        enabled: false
    dogstatsd:
      hostPortConfig:
        enabled: true
      unixDomainSocketConfig:
        enabled: true
```

### DIFF
--- a/controllers/datadogagent/feature/admissioncontroller/feature.go
+++ b/controllers/datadogagent/feature/admissioncontroller/feature.go
@@ -64,9 +64,8 @@ func (f *admissionControllerFeature) Configure(dda *v2alpha1.DatadogAgent) (reqC
 			// use `socket` mode if either apm or dsd uses uds
 			apm := dda.Spec.Features.APM
 			dsd := dda.Spec.Features.Dogstatsd
-			if apm != nil && apiutils.BoolValue(apm.Enabled) && apiutils.BoolValue(apm.UnixDomainSocketConfig.Enabled) {
-				f.agentCommunicationMode = apicommon.AdmissionControllerSocketCommunicationMode
-			} else if dsd.UnixDomainSocketConfig != nil && apiutils.BoolValue(dsd.UnixDomainSocketConfig.Enabled) {
+			if apm != nil && apiutils.BoolValue(apm.Enabled) && apiutils.BoolValue(apm.UnixDomainSocketConfig.Enabled) ||
+				dsd.UnixDomainSocketConfig != nil && apiutils.BoolValue(dsd.UnixDomainSocketConfig.Enabled) {
 				f.agentCommunicationMode = apicommon.AdmissionControllerSocketCommunicationMode
 			}
 			// otherwise don't set to fall back to default agent setting `hostip`


controllers/datadogagent/feature/admissioncontroller/feature.go
```diff
@@ -56,8 +56,20 @@ func (f *admissionControllerFeature) Configure(dda *v2alpha1.DatadogAgent) (reqC
...
 		}
+		// agent communication mode set by user
 		if ac.AgentCommunicationMode != nil && *ac.AgentCommunicationMode != "" {
 			f.agentCommunicationMode = *ac.AgentCommunicationMode
+		} else {
+			// agent communication mode set automatically
+			// use `socket` mode if either apm or dsd uses uds
+			apm := dda.Spec.Features.APM
+			dsd := dda.Spec.Features.Dogstatsd
+			if apm != nil && apiutils.BoolValue(apm.Enabled) && apiutils.BoolValue(apm.UnixDomainSocketConfig.Enabled) {
```
celenechang, line 67: should we also check here if `apm.UnixDomainSocketConfig != nil` ?

controllers/datadogagent/feature/admissioncontroller/feature.go
```diff
@@ -56,8 +56,20 @@ func (f *admissionControllerFeature) Configure(dda *v2alpha1.DatadogAgent) (reqC
...
+				f.agentCommunicationMode = apicommon.AdmissionControllerSocketCommunicationMode
+			} else if dsd.UnixDomainSocketConfig != nil && apiutils.BoolValue(dsd.UnixDomainSocketConfig.Enabled) {
```
celenechang, line 69: could `dsd` be `nil` also?

celenechang: maybe instead of `if/else if` we could do 

`if (apm !=nil && ...) || (dsd != nil && ...)`  (but put them on separate lines for readability)

khewonc: DSD should also default to a non-nil value: https://github.com/DataDog/datadog-operator/blob/fe461fbaa4cfeb30857ca054753a50e618af74e0/apis/datadoghq/v2alpha1/datadogagent_default.go#L195-L197
I'll combine the `if` statements 👍

controllers/datadogagent/feature/admissioncontroller/feature.go
celenechang: i see, as long as apm is enabled. thanks!

PR #828, opened 2023-06-28
Add grpc port and http port to service when otlp is enabled

maxwellbogh:
Bug Fix

In the DatadogAgent config there is an option to enable otlp ingestion:
```
otlp:
    receiver:
    protocols:
        grpc:
            enabled: true
        http:
            enabled: true
```
This adds the following two containerPorts to the pod spec:
```
- containerPort: 4317
  hostPort: 4317
  name: otlpgrpcport
  protocol: TCP
- containerPort: 4318
  hostPort: 4318
  name: otlphttpport
  protocol: TCP
```
The problem is that these two ports are only added to the pod spec and not to the service object.

This PR adds the http and grpc ports to the service object if enabled.

### DIFF
--- a/controllers/datadogagent/feature/otlp/feature.go
+++ b/controllers/datadogagent/feature/otlp/feature.go
@@ -12,6 +12,8 @@ import (
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


fanny-jiang (approved):
Thanks for adding this!

fanny-jiang:
CI checks pass on my branch targeting main: https://github.com/DataDog/datadog-operator/commit/9d6fc5e1c5fce35a25871c5ee5ec60a6373083d9

PR #835, opened 2023-07-10
update feature factory to process feature if disabled but configured

celenechang:
### What does this PR do?

Add `IsConfigured` to FeatureFactory
Update naming of a few bool variables
Use `IsConfigured` when collating list of features to manage in each component reconciler

### Motivation

An issue arose regarding a feature that needs to be explicitly disabled when it's not enabled, but we didn't fully form the ability to do this with the FeatureFactory. This PR finishes what was started several months ago

### Additional Notes

### Describe your test plan

Deploy the operator and add a section to one of the features such as this, where an envvar is set when a feature is **disabled** (this example was tested in remote config):

```
       } else {
               reqComp = feature.RequiredComponents{
                       Agent: feature.RequiredComponent{
                               IsRequired: apiutils.NewBoolPointer(false),
                               Containers: []apicommonv1.AgentContainerName{
                                       apicommonv1.CoreAgentContainerName,
                               },
                       },
               }
        }
```

Apply a DatadogAgent manifest without the feature enabled, and ensure the expected configuration is set

### DIFF
--- a/controllers/datadogagent/controller_reconcile_agent.go
+++ b/controllers/datadogagent/controller_reconcile_agent.go
@@ -36,9 +36,9 @@ func (r *Reconciler) reconcileV2Agent(logger logr.Logger, requiredComponents fea
 
 	// requiredComponents needs to be taken into account in case a feature(s) changes and
 	// a requiredComponent becomes disabled, in addition to taking into account override.Disabled
-	disabled := false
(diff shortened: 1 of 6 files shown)


levan-m (approved):

clamoriniere (commented):
🚀  Nice work !!!

PR #841, opened 2023-07-13
Use default dogstatsd port for creating agent local service

fanny-jiang:
### What does this PR do?

* Remove dependency on `features.dogstatsd.hostPortConfig.enabled` to create the agent local service. Instead use the default dogstatsd port, 8125
* Remove checking for minimum supported local service version 1.21 if `forceEnableLocalService=true` before creating agent local service. Instead, `ShouldCreateAgentLocalService` returns `true` if Kubernetes version is 1.22+ or `forceEnableLocalService=true`

### Motivation

* Bug fix
* Improve parity with datadog helm chart. 

### Additional Notes

### Describe your test plan

* Deploy datadog-operator and minimal datadogagent manifest
* Check that the `dogstatsdport` is present on the agent service object: `kubectl get service datadog-agent -oyaml`

```

```

### DIFF
--- a/controllers/datadogagent/feature/dogstatsd/feature.go
+++ b/controllers/datadogagent/feature/dogstatsd/feature.go
@@ -140,20 +140,18 @@ func (f *dogstatsdFeature) ConfigureV1(dda *v1alpha1.DatadogAgent) (reqComp feat
 // ManageDependencies allows a feature to manage its dependencies.
 // Feature's dependencies should be added in the store.
 func (f *dogstatsdFeature) ManageDependencies(managers feature.ResourceManagers, components feature.RequiredComponents) error {
-	if f.hostPortEnabled {
-		// agent local service
-		if component.ShouldCreateAgentLocalService(managers.Store().GetVersionInfo(), f.forceEnableLocalService) {
-			apmPort := []corev1.ServicePort{
-				{
-					Protocol:   corev1.ProtocolUDP,
-					TargetPort: intstr.FromInt(int(f.hostPortHostPort)),
-					Port:       f.hostPortHostPort,
-					Name:       apicommon.DogstatsdHostPortName,
-				},
-			}
-			if err := managers.ServiceManager().AddService(f.localServiceName, f.owner.GetNamespace(), nil, apmPort, nil); err != nil {
-				return err
-			}
+	// agent local service
+	if component.ShouldCreateAgentLocalService(managers.Store().GetVersionInfo(), f.forceEnableLocalService) {
+		dsdPort := []corev1.ServicePort{
+			{
+				Protocol:   corev1.ProtocolUDP,
+				TargetPort: intstr.FromInt(apicommon.DefaultDogstatsdPort),
+				Port:       apicommon.DefaultDogstatsdPort,
+				Name:       apicommon.DefaultDogstatsdPortName,
+			},
+		}
+		if err := managers.ServiceManager().AddService(f.localServiceName, f.owner.GetNamespace(), nil, dsdPort, nil); err != nil {
+			return err
 		}
 	}
 
--- a/controllers/datadogagent/component/utils.go
+++ b/controllers/datadogagent/component/utils.go
@@ -26,7 +26,6 @@ import (
 )
 
 const (
-	localServiceMinimumVersion        = "1.21-0"
 	localServiceDefaultMinimumVersion = "1.22-0"
 )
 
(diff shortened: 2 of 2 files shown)


controllers/datadogagent/feature/dogstatsd/feature.go
```diff
@@ -140,20 +140,18 @@ func (f *dogstatsdFeature) ConfigureV1(dda *v1alpha1.DatadogAgent) (reqComp feat
...
-			if err := managers.ServiceManager().AddService(f.localServiceName, f.owner.GetNamespace(), nil, apmPort, nil); err != nil {
-				return err
-			}
+	// agent local service
+	if component.ShouldCreateAgentLocalService(managers.Store().GetVersionInfo(), f.forceEnableLocalService) {
+		dsdPort := []corev1.ServicePort{
+			{
+				Protocol:   corev1.ProtocolUDP,
+				TargetPort: intstr.FromInt(apicommon.DefaultDogstatsdPort),
+				Port:       apicommon.DefaultDogstatsdPort,
```
clamoriniere, line 149: think we need to add another `if` condition to override the `apicommon.DefaultDogstatsdPort` by the `f.hostPortHostPort`. 
the usecase that I see is when the agent pod is configured to run with `host network`, the container port is corresponding to the host port. 

to know if hostNetwork is enable we have this function https://github.com/DataDog/datadog-operator/blob/6beacdb5898bf470669c846526c8ebe1dcdeebea/apis/datadoghq/v2alpha1/utils.go#L73

and then it is used here https://github.com/DataDog/datadog-operator/blob/6beacdb5898bf470669c846526c8ebe1dcdeebea/controllers/datadogagent/feature/dogstatsd/feature.go#L197-L199

clamoriniere (commented):
look good, I added a comment about supporting properly hostnetwork configuration

apis/datadoghq/common/const.go
```diff
@@ -56,6 +56,8 @@ const (
 	DefaultDogstatsdPort = 8125
 	// DefaultDogstatsdPortName default dogstatsd port name
 	DefaultDogstatsdPortName = "dogstatsd"
```
fanny-jiang, line 58: The default dogstatsd port name `dogstatsd` is different from the `DogstatsdHostPortName`, which is `dogstatsdport` [(line 200)](https://github.com/DataDog/datadog-operator/pull/841/files#diff-9088a8abc4fac93cee6e4dd256cffc750221214fc8cf62040974e443b9f9853fR200). Should they be made the same? I.e., both be `dogstatsdport` like in the [helm chart](https://github.com/DataDog/helm-charts/blob/57a411047e49e9e31cad8532effb3d7e0e9b3ce3/charts/datadog/templates/agent-services.yaml#L84C13-L84C26).

controllers/datadogagent/feature/apm/feature.go
```diff
@@ -147,12 +150,19 @@ func (f *apmFeature) ConfigureV1(dda *v1alpha1.DatadogAgent) (reqComp feature.Re
 func (f *apmFeature) ManageDependencies(managers feature.ResourceManagers, components feature.RequiredComponents) error {
 	// agent local service
 	if component.ShouldCreateAgentLocalService(managers.Store().GetVersionInfo(), f.forceEnableLocalService) {
+		if f.hostPortEnabled {
```
fanny-jiang, line 153: Should we check if hostNetwork is being used?

apis/datadoghq/common/const.go
celenechang: i'd be in favor of this

controllers/datadogagent/feature/apm/feature.go
clamoriniere: yes if hostPortEnabled and hostNetwork are enabled the container's Pod `Port` and so `TargetPort` must be equal to the `hostPort`.
So you need also to have the `ServicePort.TargetPort` equal to `hostPort`

controllers/datadogagent/feature/dogstatsd/feature.go
```diff
@@ -140,20 +140,22 @@ func (f *dogstatsdFeature) ConfigureV1(dda *v1alpha1.DatadogAgent) (reqComp feat
...
+			Protocol:   corev1.ProtocolUDP,
+			TargetPort: intstr.FromInt(int(apicommon.DefaultDogstatsdPort)),
+			Port:       apicommon.DefaultDogstatsdPort,
+			Name:       apicommon.DefaultDogstatsdPortName,
+		}
+		if f.useHostNetwork && f.hostPortEnabled {
+			dsdPort.TargetPort = intstr.FromInt(int(f.hostPortHostPort))
+			dsdPort.Port = f.hostPortHostPort
+			dsdPort.Name = apicommon.DogstatsdHostPortName
+		}
```
clamoriniere, line 155: I think it still need to make the diff, between `if f.hostPortEnabled ` and `if f.useHostNetwork && f.hostPortEnabled`

```suggestion
		if f.hostPortEnabled {
		    dsdPort.Port = f.hostPortHostPort
			dsdPort.Name = apicommon.DogstatsdHostPortName
		    if f.useHostNetwork {
			   dsdPort.TargetPort = intstr.FromInt(int(f.hostPortHostPort))
			}
		}
```

fanny-jiang: Thanks, I assume that difference should be made for APM too?

clamoriniere (approved):
👍 
thank you for the reactivity on the changes after code reviews 🙇

PR #843, opened 2023-07-14
Force pretty print CRDs for consistent formatting

levan-m:
### What does this PR do?

It looks `yq` in `patch-crds.sh` formatted some CRDs, while others were kept as generated by `controller-gen`. This caused `description` fields to have line breaks and lists not indent. This change will apply same formatting to all crds which should make them easier to review in the repo or when copying to other repos.  

### Motivation

Consistent formatting for PRs.

### Additional Notes

### Describe your test plan


celenechang (approved):
thanks - i'd be OK with leaving the script name as is since the pretty print is just formatting, but up to you

levan-m:
I was 50/50 so reverted script and Make target name.

PR #865, opened 2023-07-25
format agent version check

celenechang:
### What does this PR do?

Adds a format function in pkg/utils/version to support image tags that use dashes in lieu of periods.

### Motivation

### Additional Notes

### Describe your test plan

Test a DatadogAgent CR with the node agent image override set to something like 7-46-0-jmx . Ensure that the KSMFeature RBACs and collectors are present as expected.

### DIFF
--- a/pkg/utils/version_test.go
+++ b/pkg/utils/version_test.go
@@ -11,7 +11,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestCompareVersion(t *testing.T) {
+func TestIsAboveMinVersion(t *testing.T) {
 	testCases := []struct {
 		version    string
 		minVersion string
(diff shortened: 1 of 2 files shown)


levan-m (approved):

pkg/utils/version_test.go
```diff
@@ -70,3 +110,74 @@ func TestCompareVersion(t *testing.T) {
...
+			version:  "7-46-0-beta",
+			expected: "7.46.0-beta",
+		},
+		{
+			version:  "customImage",
+			expected: "",
+		},
+		{
+			version:  "customImage",
+			expected: "",
```
khewonc, line 166: Nit: duplicate test

khewonc (approved):

PR #872, opened 2023-07-31
Improve internal image tagging

clamoriniere:
### What does this PR do?

Add a `commit sha` in the internal image tag, if the build is triggered manually

### Motivation

simply internal testing. Adding the commit sha in the image tag allow to push several image from the same branch without overriding the previous version. It can be usefull if the image was already replicated in some registry. Or to compare two versions

### Additional Notes

### Describe your test plan

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -216,8 +216,11 @@ trigger_internal_operator_image:
     - if: $CI_COMMIT_TAG
       variables:
         IMAGE_RELEASE_PROD: "true"
+        IMAGE_TAG: ${CI_COMMIT_REF_SLUG}
     - if: $PUSH_IMAGES_TO_STAGING == 'true'
       when: manual
+      variables:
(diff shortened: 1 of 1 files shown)


celenechang (approved):

.gitlab-ci.yml
```diff
@@ -214,38 +213,44 @@ trigger_internal_operator_image:
...
+    IMAGE_VERSION: tmpl-v1
+    IMAGE_NAME: $PROJECTNAME_CHECK
+    TMPL_SRC_IMAGE: v${CI_PIPELINE_ID}-${CI_COMMIT_SHORT_SHA}
+    RELEASE_TAG: ${CI_COMMIT_REF_SLUG}
+    BUILD_TAG: ${CI_COMMIT_REF_SLUG}
+    RELEASE_STAGING: "true"
+    RELEASE_PROD: "true"
+
+
+trigger_internal_staging_only_operator_image:
```
celenechang, line 249: could we call this something like `trigger_custom_operator_image_staging` (and similarly for the operator check)? to distinguish its use more from the above job `trigger_internal_operator_image`

PR #873, opened 2023-07-31
[datadogmonitor] reduce number of reconcile requeues

celenechang:
### What does this PR do?

Reduce the high number of requeues of a datadogmonitor instance to lighten load of operator. The primary change is to only update a status condition if the status condition status has changed, or the message has changed.
Remove `result.Requeue: true` if `result.RequeueAfter` is set (this is implied)

### Motivation

### Additional Notes

### Describe your test plan

The most effective way to test this change is to examine a cluster that is reconciling many DatadogMonitor objects. The test cluster we used had ~90. Check that the reconcile count per monitor is not more than a couple per minute.

### DIFF
--- a/pkg/controller/utils/condition/datadogmonitor.go
+++ b/pkg/controller/utils/condition/datadogmonitor.go
@@ -45,9 +45,12 @@ func SetDatadogMonitorCondition(condition *datadoghqv1alpha1.DatadogMonitorCondi
 	if condition.Status != conditionStatus {
 		condition.LastTransitionTime = now
 		condition.Status = conditionStatus
+		condition.LastUpdateTime = now
+		condition.Message = desc
+	} else if condition.Message != desc {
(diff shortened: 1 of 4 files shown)


fanny-jiang (approved):

pkg/controller/utils/condition/datadogmonitor.go
```diff
@@ -45,9 +45,12 @@ func SetDatadogMonitorCondition(condition *datadoghqv1alpha1.DatadogMonitorCondi
 	if condition.Status != conditionStatus {
 		condition.LastTransitionTime = now
 		condition.Status = conditionStatus
+		condition.LastUpdateTime = now
+		condition.Message = desc
+	} else if condition.Message != desc {
```
celenechang, line 50: if the message has changed but the status has not, no need to update the other fields

khewonc (approved):

PR #883, opened 2023-08-03
upgrade(remote-config): Create cluster agent volume when Remote Config is enabled

BaptisteFoy:
### What does this PR do?
Remote Config needs access to `/opt/datadog-agent/run` in order to create its local database where configurations & their metadata are stored. By default, the cluster agent has the `readOnlyRootFilesystem` parameter set to true for security reasons. To circumvent this, we create an EmptyDir volume on the required path so that the agent can create the Remote Config database.

Current workaround is to create the volume at agent creation or disable the `readOnlyRootFilesystem` parameter. This doesn't impact many users currently as no product using Remote Config on the cluster agent is GA.

### Motivation
Fixing bugs

### Additional Notes
N/A

### Minimum Agent Versions
No

### Describe your test plan
Create a cluster agent and check in the logs it is able to start Remote Config properly
Error: `Failed to start remote-configuration: unable to create remote-config service: failed to create rc db dir`
Success: `

### Checklist

### DIFF
--- a/controllers/datadogagent/feature/remoteconfig/feature.go
+++ b/controllers/datadogagent/feature/remoteconfig/feature.go
@@ -119,7 +119,7 @@ func (f *rcFeature) ManageClusterAgent(managers feature.PodTemplateManagers) err
 
 	if f.enabled {
 		// Volume to create the Remote Config Database
-		// Mandatory as the cluster agent root FS is by default read only
+		// Mandatory as the cluster agent root FS is read only by default 
 		managers.Volume().AddVolume(rcVolume)
 		managers.VolumeMount().AddVolumeMount(rcVolumeMount)
 	}
--- a/controllers/datadogagent/feature/remoteconfig/feature.go
+++ b/controllers/datadogagent/feature/remoteconfig/feature.go
@@ -21,7 +21,7 @@ import (
 )
 
 const (
-	rcDBVolumeName = "remote-config"
+	rcDBVolumeName = "datadogrun"
 	rcDBVolumePath = "/opt/datadog-agent/run/"
 )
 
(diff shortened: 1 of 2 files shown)


controllers/datadogagent/feature/remoteconfig/feature_test.go
```diff
@@ -52,38 +52,86 @@ func Test_rcFeature_Configure(t *testing.T) {
...
 }
+
+func checkEqual(t testing.TB, description string, expected interface{}, actual interface{}) {
+	assert.True(
+		t,
+		apiutils.IsEqualStruct(expected, actual),
+		"%s\ndiff = %s",
+		description,
+		cmp.Diff(expected, actual),
+	)
```
julesmcrt, line 136: **❓ Question:** ‏Why don't you use `assert.Equal()`?

controllers/datadogagent/feature/remoteconfig/feature.go
```diff
@@ -93,6 +117,13 @@ func (f *rcFeature) ManageClusterAgent(managers feature.PodTemplateManagers) err
 	}
 	managers.EnvVar().AddEnvVar(enabledEnvVar)
 
+	if f.enabled {
+		// Volume to create the Remote Config Database
+		// Mandatory as the cluster agent root FS is by default read only
```
julesmcrt, line 122: **🥜 Nitpick:** ‏‏

```suggestion
		// Mandatory as the cluster agent root FS is read only by default 
```

coignetp (approved):

controllers/datadogagent/feature/remoteconfig/feature_test.go
BaptisteFoy: There is `IsEqualStruct` defined in the operator which is more powerful as it'll check equality at the attribute level. This seems to be standard in the operator.

mellon85 (approved):

celenechang (approved):

PR #918, opened 2023-09-12
Levan m/cc feature segfault

levan-m:
### What does this PR do?

Recent change led to a bug which causes segfault when `features.clusterChecks.enabled` is set to `false`. This is due to the fact that `Configure` function returns a non-empty `RequiredComponents` later leading call to `ManageDependencies`, which should happen since feature isn't enabled.

PR improves the test suite so this kind of issues can be caught. Previously suite made direct calls to `feature.Configure` and decided whether to exit early without calling `ManageDependencies`. With the update suite calls `feature.BuildFeatures` and verifies output for errors and lets all `Managed*` functions run on the output. This approach is closer to the prod logic.

### Motivation

segviolation with a basic feature config. 

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist


controllers/datadogagent/feature/clusterchecks/feature.go
```diff
@@ -57,22 +57,23 @@ func (f *clusterChecksFeature) ID() feature.IDType {
...
-			f.logger.Error(err, "couldn't generate hash for cluster checks config")
-		} else {
-			f.logger.V(2).Info("created cluster checks", "hash", hash)
-		}
-		f.customConfigAnnotationValue = hash
-		f.customConfigAnnotationKey = object.GetChecksumAnnotationKey(feature.ClusterChecksIDType)
-	}
 
 	if clusterChecksEnabled {
+		f.useClusterCheckRunners = clusterChecksEnabled && apiutils.BoolValue(dda.Spec.Features.ClusterChecks.UseClusterChecksRunners)
```
celenechang, line 64:
```suggestion
		f.useClusterCheckRunners = apiutils.BoolValue(dda.Spec.Features.ClusterChecks.UseClusterChecksRunners)
```

in fact, maybe we can get rid of the `clusterChecksEnabled` variable and just do `if apiutils.BoolValue(dda.Spec.Features.ClusterChecks.Enabled)` at the top of the block, since we don't need this var elsewhere?

levan-m: I overlooked the redundancy, thanks for catching it, fixed now. Had to force push after rebasing main.

controllers/datadogagent/feature/clusterchecks/feature.go
```diff
@@ -57,22 +57,21 @@ func (f *clusterChecksFeature) ID() feature.IDType {
...
-		hash, err := comparison.GenerateMD5ForSpec(dda.Spec.Features.ClusterChecks)
-		if err != nil {
-			f.logger.Error(err, "couldn't generate hash for cluster checks config")
-		} else {
-			f.logger.V(2).Info("created cluster checks", "hash", hash)
+func (f *clusterChecksFeature) Configure(dda *v2alpha1.DatadogAgent) (reqComp feature.RequiredComponents) {
+	if apiutils.BoolValue(dda.Spec.Features.ClusterChecks.Enabled) {
+		f.useClusterCheckRunners = apiutils.BoolValue(dda.Spec.Features.ClusterChecks.UseClusterChecksRunners)
+
+		{
```
celenechang, line 64: i think this set of brackets is not needed? although curiously i see it was there before

levan-m: Yes, it's not really needed. Not sure what original intent was but it improves readability and narrows variable scope. I moved it to a method.

celenechang (approved):
thanks for this

PR #936, opened 2023-10-02
[CECO-639] Dogstatsd tag cardinality feature

tbavelier:
### What does this PR do?

* Adds a new feature `dogstatsd.tagCardinality` to configure the cardinality of Dogstatsd custom metrics when origin detection is enabled. It defaults to `low` in the Agent : https://github.com/DataDog/datadog-agent/blob/65c790dbe099096f73a9e148736810035b8be5d5/pkg/config/config.go#L985.
    * If the user input value is not in the enum `{low, orchestrator, high}`, the Agent code ignores it and defaults to `low` per https://github.com/DataDog/datadog-agent/blob/5976398e0ce92c506c8e6b7bdfcd98a30935ab19/pkg/tagger/collectors/utils.go#L28C6-L28C28

### Motivation

* https://datadoghq.atlassian.net/browse/CECO-639

### Additional Notes

* Since `v2alpha1` has been GA for a few months, this enhancement is only added in the new API and not to `v1alpha1`

### Minimum Agent Versions

No

### Describe your test plan

1. install the custom operator image without installing the CRDs : `helm install my-datadog-operator datadog/datadog-operator --set  installCRDs=false`
2. Apply the new CRD from this PR : `kubectl create -f config/crd/bases/v1/datadoghq.com_datadogagents.yaml`
3. Test the following scenarios and ensure the results are adequate by describing the Agent pod and reviewing the env var of the `agent` container with `kubectl describe pod <NODE AGENT POD> | grep "DD_DOGSTATSD"`:
    1. ```yaml
       kind: DatadogAgent
       metadata:
         name: datadog
       spec:
         global:
           [...]
         features:
           dogstatsd:
             originDetectionEnabled: true
       ```
       **Scenario : origin detection and no specific tag cardinality**
       Assert the presence of `DD_DOGSTATSD_ORIGIN_DETECTION: true` and absence of `DD_DOGSTATSD_TAG_CARDINALITY` 
    2. ```yaml
       kind: DatadogAgent
       metadata:
         name: datadog
       spec:
         global:
           [...]
         features:
           dogstatsd:
             originDetectionEnabled: true
             tagCardinality: "illegal_value"
       ```
       **Scenario : origin detection and not allowed specific tag cardinality**
       Assert the presence of `DD_DOGSTATSD_ORIGIN_DETECTION: true` and presence of `DD_DOGSTATSD_TAG_CARDINALITY: illegal_value`.
       Ensure the Agent logs a warning mentioning it is reverting to low : Image 2023-10-06 at 9 29 07 AM

    3. ```yaml
       kind: DatadogAgent
       metadata:
         name: datadog
       spec:
         global:
           [...]
         features:
           dogstatsd:
             originDetectionEnabled: true
             tagCardinality: "high"
       ```
       **Scenario : origin detection and allowed specific tag cardinality**
       Assert the presence of `DD_DOGSTATSD_ORIGIN_DETECTION: true` and `DD_DOGSTATSD_TAG_CARDINALITY: high`


### Checklist

### DIFF
--- a/apis/datadoghq/v2alpha1/datadogagent_default.go
+++ b/apis/datadoghq/v2alpha1/datadogagent_default.go
@@ -210,7 +210,9 @@ func defaultFeaturesConfig(ddaSpec *DatadogAgentSpec) {
 
 	apiutils.DefaultBooleanIfUnset(&ddaSpec.Features.Dogstatsd.OriginDetectionEnabled, defaultDogstatsdOriginDetectionEnabled)
 
-	apiutils.DefaultStringIfUnset(&ddaSpec.Features.Dogstatsd.TagCardinality, defaultDogstatsdTagCardinality)
+	if *ddaSpec.Features.Dogstatsd.OriginDetectionEnabled {
+		apiutils.DefaultStringIfUnset(&ddaSpec.Features.Dogstatsd.TagCardinality, defaultDogstatsdTagCardinality)
+	}
 
 	if ddaSpec.Features.Dogstatsd.HostPortConfig == nil {
 		ddaSpec.Features.Dogstatsd.HostPortConfig = &HostPortConfig{
--- a/apis/datadoghq/v2alpha1/datadogagent_default.go
+++ b/apis/datadoghq/v2alpha1/datadogagent_default.go
@@ -51,7 +51,6 @@ const (
 	// defaultUSMEnabled bool = false
 
 	defaultDogstatsdOriginDetectionEnabled bool   = false
-	defaultDogstatsdTagCardinality         string = "low"
 	defaultDogstatsdHostPortEnabled        bool   = false
 	defaultDogstatsdPort                   int32  = 8125
 	defaultDogstatsdSocketEnabled          bool   = true
@@ -209,10 +208,6 @@ func defaultFeaturesConfig(ddaSpec *DatadogAgentSpec) {
 	}
 
 	apiutils.DefaultBooleanIfUnset(&ddaSpec.Features.Dogstatsd.OriginDetectionEnabled, defaultDogstatsdOriginDetectionEnabled)
-
-	if *ddaSpec.Features.Dogstatsd.OriginDetectionEnabled {
-		apiutils.DefaultStringIfUnset(&ddaSpec.Features.Dogstatsd.TagCardinality, defaultDogstatsdTagCardinality)
-	}
 
 	if ddaSpec.Features.Dogstatsd.HostPortConfig == nil {
 		ddaSpec.Features.Dogstatsd.HostPortConfig = &HostPortConfig{
--- a/apis/datadoghq/v2alpha1/datadogagent_default.go
+++ b/apis/datadoghq/v2alpha1/datadogagent_default.go
@@ -51,6 +51,7 @@ const (
 	// defaultUSMEnabled bool = false
 
 	defaultDogstatsdOriginDetectionEnabled bool   = false
+	defaultDogstatsdTagCardinality         string = "low"
 	defaultDogstatsdHostPortEnabled        bool   = false
 	defaultDogstatsdPort                   int32  = 8125
 	defaultDogstatsdSocketEnabled          bool   = true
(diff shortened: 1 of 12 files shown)


apis/datadoghq/v2alpha1/datadogagent_default.go
```diff
@@ -209,6 +210,10 @@ func defaultFeaturesConfig(ddaSpec *DatadogAgentSpec) {
 
 	apiutils.DefaultBooleanIfUnset(&ddaSpec.Features.Dogstatsd.OriginDetectionEnabled, defaultDogstatsdOriginDetectionEnabled)
 
+	if *ddaSpec.Features.Dogstatsd.OriginDetectionEnabled {
+		apiutils.DefaultStringIfUnset(&ddaSpec.Features.Dogstatsd.TagCardinality, defaultDogstatsdTagCardinality)
+	}
+
```
celenechang, line 216: do you mind removing this so that the default setting will fallback to the agent?

apis/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -354,6 +354,12 @@ type DogstatsdFeatureConfig struct {
 	// +optional
 	OriginDetectionEnabled *bool `json:"originDetectionEnabled,omitempty"`
 
+	// TagCardinality configures tag cardinality for the metrics collected using origin detection (`low`, `orchestrator` or `high`).
+	// See also: https://docs.datadoghq.com/getting_started/tagging/assigning_tags/?tab=containerizedenvironments#environment-variables
+	// Cardinality default: low
+	// +optional
+	TagCardinality *string `json:"tagCardinality,omitempty"`
```
celenechang, line 361: in lieu of the function below to check for validity, we could add a new struct here `TagCardinalityName` (for example), and then three constants of this type - similar to what we do for `AgentConfigFileName`

maycmlee (approved):

controllers/datadogagent/feature/dogstatsd/feature.go
```diff
@@ -47,6 +47,7 @@ type dogstatsdFeature struct {
 
 	useHostNetwork         bool
 	originDetectionEnabled bool
+	tagCardinality         string
```
celenechang, line 50: suggestion: use type `*TagCardinalityName` here, and convert to string when setting the envvar below

controllers/datadogagent/feature/dogstatsd/feature.go
```diff
@@ -239,6 +243,15 @@ func (f *dogstatsdFeature) ManageNodeAgent(managers feature.PodTemplateManagers)
 		if f.udsEnabled {
 			managers.PodTemplateSpec().Spec.HostPID = true
 		}
+		// Tag cardinality is only configured if origin detection is enabled.
+		// Only three values (`low`, `orchestrator` and `high`) can be configured, others are ignored.
+		switch v2alpha1.TagCardinalityName(f.tagCardinality) {
+		case v2alpha1.LowTagCardinality, v2alpha1.OrchestratorTagCardinality, v2alpha1.HighTagCardinality:
```
celenechang, line 249: since we already use a defined type for `TagCardinalityName`, i don't think switch/case is needed anymore and could just do `if f.tagCardinality != nil`

apis/datadoghq/v2alpha1/datadogagent_types.go
tbavelier: Addressed in comment below

controllers/datadogagent/feature/dogstatsd/feature.go
tbavelier: Due to the niche feature and the fact there is already an Agent validation, we defer to the latter instead of adding an additional validation in the Agent. This simplifies the code by removing the alias type and the switch case. Addressed in https://github.com/DataDog/datadog-operator/pull/936/commits/d5e41d371eba46b837bd531beb48d064cb78eaca

celenechang (approved):
looks great, thanks for the teamwork @tbavelier

PR #950, opened 2023-10-16
[kubelet] Add kubelet CA volume mount to security Agent container

tbavelier:
### What does this PR do?

* Mounts kubelet CA certificates in `security-agent` container when provided by the user in `spec.global.kubelet`

### Motivation

* Self-testing, my `security-agent` container would CLBO since it wasn't able to get a hostname from `kubelet` despite providing the certificates to ensure TLS connection

### Minimum Agent Versions

~~Are there minimum versions of the Datadog Agent and/or Cluster Agent required?~~
* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

* Start a basic `kind` cluster (kubelet CA certificate is stored in `/var/lib/kubelet/pki/kubelet.crt`)
* Use a CRD that enables the security agent such as below, while using a kubelet CA host configuration using the node name as the kubelet host (since `kind` does not set the node IP as SAN) : 
```
spec:
  global:
    kubelet:
      hostCAPath: "/var/lib/kubelet/pki/kubelet.crt"
      host:
        fieldRef:
          fieldPath: spec.nodeName
  features:
    cspm:
      enabled: true
```
* Ensure the security Agent have the certificate mounted in `/var/run/host-kubelet-ca.crt` and is able to access the kubelet by reviewing its logs

### Checklist

### DIFF
--- a/controllers/datadogagent/override/global.go
+++ b/controllers/datadogagent/override/global.go
@@ -205,6 +205,8 @@ func ApplyGlobalSettings(logger logr.Logger, manager feature.PodTemplateManagers
 						apicommonv1.CoreAgentContainerName,
 						apicommonv1.ProcessAgentContainerName,
 						apicommonv1.TraceAgentContainerName,
+						apicommonv1.SystemProbeContainerName,
+						apicommonv1.SecurityAgentContainerName,
 					},
 				)
 				manager.Volume().AddVolume(&kubeletVol)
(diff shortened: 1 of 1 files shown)


tbavelier:
The `security-agent` and `system-probe` should retrieve the hostname "remotely" from the core Agent instead of directly accessing the kubelet, thus they do not need/require the kubelet CA certificates.

PR #972, opened 2023-11-08
fix(rbac): enable PodSecurityPolicy rules only when supported

inobu:
fix https://github.com/DataDog/datadog-operator/issues/975

### What does this PR do?

This PR updates the RBAC policy rule generation logic to conditionally include support for the `PodSecurityPolicy` resource based on a support flag. This change is in response to the deprecation and removal of the `api: policy/v1beta1` PodSecurityPolicy resource in Kubernetes 1.25.

### Motivation

Kubernetes 1.25 has removed the `api: policy/v1beta1` PodSecurityPolicy resource, which necessitates changes in our RBAC policy rule generation to ensure our application remains compatible with the latest Kubernetes API changes. This PR introduces a flag that, when set, includes the necessary rules for `PodSecurityPolicy` only if the cluster supports it, thus maintaining backward compatibility while respecting the latest Kubernetes version's API.

### Additional Notes

While this change primarily targets compatibility with Kubernetes 1.25, it's implemented in a manner that ensures backward compatibility with earlier versions of Kubernetes that still support `PodSecurityPolicy`. As a result, this update should not impact any existing deployments on older Kubernetes versions.

### Minimum Agent Versions

To accommodate the changes in Kubernetes API, the minimum versions of the Datadog Agent and the Cluster Agent are specified as follows:

* Agent: vX.Y.Z (Please specify the minimum version that includes the updated RBAC logic)
* Cluster Agent: vX.Y.Z (Please specify the minimum version that includes the updated RBAC logic)

### Describe your test plan

Testing this PR should include the following scenarios:

1. Deploying on a Kubernetes cluster version 1.25 and verifying that no `PodSecurityPolicy` rules are generated when the support flag is disabled.
2. Deploying on a Kubernetes cluster version < 1.25 with the support flag enabled and verifying that `PodSecurityPolicy` rules are correctly generated and applied.
3. Ensuring that enabling the support flag on a Kubernetes 1.25 cluster doesn't cause any issues, given that the resource is no longer supported in this version.

### Checklist

### DIFF
--- a/controllers/datadogagent/feature/cspm/feature.go
+++ b/controllers/datadogagent/feature/cspm/feature.go
@@ -190,7 +190,9 @@ func (f *cspmFeature) ManageDependencies(managers feature.ResourceManagers, comp
 	// Manage RBAC
 	rbacName := getRBACResourceName(f.owner)
 
-	return managers.RBACManager().AddClusterPolicyRules(f.owner.GetNamespace(), rbacName, f.serviceAccountName, getRBACPolicyRules())
+	pi := managers.Store().GetPlatformInfo()
+
+	return managers.RBACManager().AddClusterPolicyRules(f.owner.GetNamespace(), rbacName, f.serviceAccountName, getRBACPolicyRules(pi.IsResourceSupported("PodSecurityPolicy")))
 }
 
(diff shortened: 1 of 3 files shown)


inobu:
How do I add a label ?
Looks like I don't have permission

PR #973, opened 2023-11-08
Profile manifest validation

khewonc:
### What does this PR do?

Adds a function to validate that a DatadogAgentProfile manifest contains the required fields

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: n/a
* Cluster Agent: n/a

### Describe your test plan

### Checklist

### DIFF
--- a/apis/datadoghq/v1alpha1/datadogagentprofile_validation.go
+++ b/apis/datadoghq/v1alpha1/datadogagentprofile_validation.go
@@ -9,8 +9,8 @@ import (
 	"fmt"
 )
 
-// IsValidDatadogAgentProfile is used to check if a DatadogAgentProfileSpec is valid
-func IsValidDatadogAgentProfile(spec *DatadogAgentProfileSpec) error {
+// ValidateDatadogAgentProfileSpec is used to check if a DatadogAgentProfileSpec is valid
+func ValidateDatadogAgentProfileSpec(spec *DatadogAgentProfileSpec) error {
 	// check that profileAffinity contains a set of requirements
 	if spec.ProfileAffinity == nil {
 		return fmt.Errorf("profileAffinity must be defined")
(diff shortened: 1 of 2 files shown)


apis/datadoghq/v1alpha1/datadogagentprofile_validation.go
```diff
@@ -0,0 +1,45 @@
...
+		return fmt.Errorf("config override must be defined")
+	}
+	if spec.Config.Override[NodeAgentComponentName] == nil {
+		return fmt.Errorf("node agent override must be defined")
+	}
+	if spec.Config.Override[NodeAgentComponentName].Containers == nil {
+		return fmt.Errorf("node agent container must be defined")
+	}
+	for name, container := range spec.Config.Override[NodeAgentComponentName].Containers {
+		if container.Resources == nil {
```
davidor, line 39: I don't think this is right because it means that the profile needs an override for each of the containers (core agent, trace agent, etc.).

I think that we should only require one. In the future, when we add support for other things, like enabling/disabling features, not even one resources override should be required, because it should be possible to have profiles that enable/disable features without specifying any resource limits/requests.

apis/datadoghq/v1alpha1/datadogagentprofile_validation.go
```diff
@@ -0,0 +1,45 @@
...
+// Copyright 2016-2020 Datadog, Inc.
+
+package v1alpha1
+
+import (
+	"fmt"
+)
+
+// IsValidDatadogAgentProfile is used to check if a DatadogAgentProfileSpec is valid
+func IsValidDatadogAgentProfile(spec *DatadogAgentProfileSpec) error {
```
davidor, line 13: Nit: I think the name is a bit confusing. If it starts with `Is`, I'd expect the function to return a boolean instead of an error.
I think we should rename it to `ValidateDatadogAgentProfileSpec` or similar.

davidor (approved):

apis/datadoghq/v1alpha1/datadogagentprofile_validation.go
```diff
@@ -0,0 +1,49 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2020 Datadog, Inc.
```
CharlyF, line 4:
```suggestion
// Copyright 2016-present Datadog, Inc.
```

PR #836, opened 2023-07-12
Allow enabling SBOM collection for host and container images

lebauce:
### What does this PR do?

Add "sbom" feature to collection SBOM for hosts and container images

### Motivation

What inspired you to submit this pull request?

### Additional Notes

Anything else we should know when reviewing?

### Describe your test plan

Write there any instructions and details you may have to test your PR.

PR #983, opened 2023-11-16
Enable by default container-image collection

clamoriniere:
### What does this PR do?

This commit set the envvar DD_CONTAINER_IMAGE_ENABLED=true by default
in the NodeAgent's "agent" container.

This PR is needed for the other PR https://github.com/DataDog/datadog-operator/pull/836 to work with agent 7.49.0

### Motivation

We want to enabled container-image collection by default. it will
be the case from agent 7.50.0. But in Operator 1.3.0, the Agent is
defaulted to 7.49.0 since 7.50.0 is not yet released.

### Additional Notes

### Minimum Agent Versions

* Agent: v7.48.0
* Cluster Agent: N/A

### Describe your test plan

Deploy the agent with the operator. The container images should appears in the new beta [container image page](https://app.datadoghq.com/containers/images)

Another way to validate that the PR works, is to exec in a running NodeAgent pod and run the `agent config` command to check if the `container_image.enabled` option is set to true.

```
$ kubectl exec -it <node-name-pod-name> -- agent config | grep -A2 container_image
container_image:
  enabled: true
```

### Checklist

### DIFF
--- a/apis/datadoghq/common/envvar.go
+++ b/apis/datadoghq/common/envvar.go
@@ -105,6 +105,7 @@ const (
 	DDRuntimeSecurityConfigNetworkEnabled             = "DD_RUNTIME_SECURITY_CONFIG_NETWORK_ENABLED"
 	DDRuntimeSecurityConfigActivityDumpEnabled        = "DD_RUNTIME_SECURITY_CONFIG_ACTIVITY_DUMP_ENABLED"
 	DDRuntimeSecurityConfigRemoteConfigurationEnabled = "DD_RUNTIME_SECURITY_CONFIG_REMOTE_CONFIGURATION_ENABLED"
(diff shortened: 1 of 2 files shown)


L3n41c (commented):
Is it OK to have this not configurable ?

celenechang:
similar to lenaic's comment, is the plan to eventually add a feature for container image collection?

clamoriniere:
Hi @celenechang anf @L3n41c 

I was surprise that https://github.com/DataDog/datadog-operator/pull/836 didn't contains the option for container-image collection. like it was done in the helm-chart.

I think we want to enable this feature by default, like we generate container metrics... so maybe no need to add a feature for it. 

If a customer want to disable it, they can use the `spec.override.nodeagent.agent.envs` to set to false.

L3n41c (approved):

PR #999, opened 2023-11-23
[profiles] Handle conflicts between profiles

davidor:
### What does this PR do?

This PR addresses the issue where multiple profiles conflict when they are applicable to the same node. 

When there's a conflict, only the DaemonSet of one of the profiles should be created. The oldest profile takes precedence. If the creation timestamp is the same, then the one alphabetically first takes precedence.

I've created this PR as draft because it includes commits from #992 so that one needs to be merged first.
Only the last commit is new.

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist


davidor:
Rebased to fix conflicts.

pkg/agentprofile/agent_profile.go
```diff
@@ -29,16 +32,52 @@ const (
...
-	return append(res, defaultProfile(res))
+		for _, node := range nodes {
+			matchesNode, err := profileMatchesNode(&profile, &node)
+			if err != nil {
+				return nil, err
+			}
+
+			if matchesNode {
+				if nodesWithProfilesApplied[node.Name] {
+					// Conflict. This profile should not be applied.
```
fanny-jiang, line 60: It would be nice if we could surface profile conflicts in the operator logs so that users know that a conflicting profile was found and not applied. Would this be the right place to log a message?

fanny-jiang (approved):
Would be nice to add logging, but not a blocker for merging the PR 👍🏼

pkg/agentprofile/agent_profile.go
davidor: I think that for users it would be better show this kind of information in the status of the profile resource.
We don't have anything yet for the status, so I prefer to leave it for later in a separate PR.

PR #1004, opened 2023-11-30
[override/podtemplatespec] Merge affinities

davidor:
### What does this PR do?

When there's an affinity override defined in the DDA it simply replaces the one defined by default. This PR changes that so that both affinities are merged instead.

We need this for the profile features because in order to make the feature work we define node affinities and anti pod affinities. We need to combine those with the ones defined in the override of the DDA instead of just replacing.

### Checklist


davidor:
Rebased.

khewonc (approved):

PR #1012, opened 2023-12-08
[APMON-406] Set Datadog namespace env variable on Cluster Agent

liliyadd:
### What does this PR do?

Sets an environment variable DD_KUBE_RESOURCES_NAMESPACE on Cluster Agent, indicating the Kubernetes namespace where Datadog is running at. This env variable is already set by Helm Charts and is used in Cluster Agent.

### Motivation

To move APM Instrumentation namespaces exclusion logic to Cluster Agent, Cluster Agent needs to know the namespace where it's running.

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Updated Cluster Agent unit-tests in the Operator

### Checklist

### DIFF
--- a/controllers/datadogagent/clusteragent.go
+++ b/controllers/datadogagent/clusteragent.go
@@ -529,6 +529,10 @@ func getEnvVarsForClusterAgent(logger logr.Logger, dda *datadoghqv1alpha1.Datado
 			Value: utils.GetDatadogLeaderElectionResourceName(dda),
 		},
 		{
+			Name:  apicommon.DDKubeResourcesNamespace,
+			Value: utils.GetDatadogAgentResourceNamespace(dda),
+		},
+		{
 			Name:  apicommon.DDComplianceConfigEnabled,
 			Value: strconv.FormatBool(complianceEnabled),
 		},
--- a/controllers/datadogagent/clusteragent_test.go
+++ b/controllers/datadogagent/clusteragent_test.go
@@ -201,6 +201,10 @@ func clusterAgentDefaultEnvVars() []corev1.EnvVar {
 			Name:  "DD_CLUSTER_AGENT_TOKEN_NAME",
 			Value: fmt.Sprintf("%stoken", testDdaName),
 		},
+		{
+			Name:  "DD_KUBE_RESOURCES_NAMESPACE",
+			Value: testDdaNamespace,
+		},
 	}
 }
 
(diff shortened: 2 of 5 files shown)


controllers/datadogagent/clusteragent.go
```diff
@@ -528,6 +528,10 @@ func getEnvVarsForClusterAgent(logger logr.Logger, dda *datadoghqv1alpha1.Datado
 			Name:  apicommon.DDLeaderLeaseName,
 			Value: utils.GetDatadogLeaderElectionResourceName(dda),
 		},
+		{
+			Name:  apicommon.DDKubeResourcesNamespace,
+			Value: utils.GetDatadogAgentResourceNamespace(dda),
+		},
```
khewonc, line 534: Hi, this file configures the cluster agent for the old controller of the operator that we're trying to move away from in favor of the newer version that was GA'ed in v1.0.0. Could you instead make the change for the newer operator version? Adding the env var under `defaultEnvVars` in [this file](https://github.com/DataDog/datadog-operator/blob/7bd890e16e83176f5cd085e39e624d040172fd45/controllers/datadogagent/component/clusteragent/default.go#L121) instead should make the equivalent change.

khewonc (changes requested):

Julia-elsammak:
Hi Team. I will be working on implementing changes for the newer operator version: 
> Adding the env var under `defaultEnvVars` (https://github.com/DataDog/datadog-operator/blob/7bd890e16e83176f5cd085e39e624d040172fd45/controllers/datadogagent/component/clusteragent/default.go#L121)

controllers/datadogagent/clusteragent_test.go
```diff
@@ -201,6 +201,10 @@ func clusterAgentDefaultEnvVars() []corev1.EnvVar {
 			Name:  "DD_CLUSTER_AGENT_TOKEN_NAME",
 			Value: fmt.Sprintf("%stoken", testDdaName),
 		},
+		{
+			Name:  "DD_KUBE_RESOURCES_NAMESPACE",
+			Value: testDdaNamespace,
+		},
```
celenechang, line 207: this can be removed as well

liliyadd:
@khewonc can you please re-review this PR now that @Julia-elsammak moved setting the new env variable into a new controller?

khewonc (approved):

PR #1021, opened 2023-12-19
[CECO-755][introspection] Limit provider values using allowlist

khewonc:
### What does this PR do?

Adds an allowlist for provider values. Currently, we only check for a valid label key so any label value will become a provider even if the operator code doesn't fully support the provider. For example, `cloud.google.com/gke-os-distribution:foo` would create a provider `foo` whether or not we had the necessary changes in agent configuration to support provider `foo`. With this PR, a provider must be allowlisted or it will automatically fall under the `default` provider.

Current behavior when adding an ubuntu nodepool:
```
$ kubectl get ds
NAME                       DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR   AGE
datadog-agent-gcp-cos      2         2         2       2            2           <none>          5m52s
datadog-agent-gcp-ubuntu   2         2         2       2            2           <none>          2m6s
```
New behavior when adding an ubuntu nodepool:
```
$ kubectl get ds
NAME                    DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR   AGE
datadog-agent-default   2         2         0       2            0           <none>          35s
datadog-agent-gcp-cos   2         2         2       2            2           <none>          5m22s
```

### Motivation

Reduce confusion on which providers may be supported in the operator

### Additional Notes

### Minimum Agent Versions

* Agent: n/a
* Cluster Agent: n/a

### Describe your test plan

### Checklist

### DIFF
--- a/pkg/kubernetes/provider.go
+++ b/pkg/kubernetes/provider.go
@@ -35,6 +35,12 @@ const (
 	GCPProviderLabel = "cloud.google.com/gke-os-distribution"
 )
 
+// ProviderValue allowlist
+var providerValueAllowlist = map[string]struct{}{
+	GCPCosContainerdProviderValue: {},
+	GCPCosProviderValue:           {},
+}
+
 // NewProviderStore generates an empty ProviderStore instance
 func NewProviderStore(log logr.Logger) ProviderStore {
 	return ProviderStore{
@@ -48,7 +54,9 @@ func determineProvider(labels map[string]string) string {
 	if len(labels) > 0 {
 		// GCP
 		if val, ok := labels[GCPProviderLabel]; ok {
-			return generateProviderName(GCPCloudProvider, val)
+			if provider := generateProviderName(GCPCloudProvider, val); provider != "" {
+				return provider
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/kubernetes/provider.go
```diff
@@ -116,10 +124,23 @@ func (p *ProviderStore) GenerateProviderNodeAffinity(provider string) []corev1.N
...
 }
 
-// generateProviderName creates a provider name from the cloud provider and provider value
-// this should not be used to create a resource name as it may contain underscores
+// generateProviderName creates a provider name from the cloud provider and
+// provider value. NOTE: this should not be used to create a resource name as
+// it may contain underscores
 func generateProviderName(cloudProvider, providerValue string) string {
-	return cloudProvider + "-" + providerValue
+	if isProviderValueAllowed(providerValue) {
```
celenechang, line 131: could we either rename `generateProviderName` , or move this check outside the function? as it stands, it doesn't seem intuitive for validation to be occurring here as well

celenechang (approved):

PR #1025, opened 2023-12-26
Add node informer and profiles store

fanny-jiang:
### What does this PR do?

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist

### DIFF
--- a/controllers/datadogagent/controller_test.go
+++ b/controllers/datadogagent/controller_test.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"github.com/DataDog/datadog-operator/pkg/agentprofile"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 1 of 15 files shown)


pkg/kubernetes/profiles.go
fanny-jiang, line 1: Maybe we can rename this to `nodeStore.go` since it is pretty generic and unspecific to datadogagent profiles.

pkg/kubernetes/nodestore_test.go
```diff
@@ -0,0 +1,427 @@
...
+var (
+	nodeUID1 = generateRandomNodeUUID()
+	nodeUID2 = generateRandomNodeUUID()
+)
+
+func Test_SetOrUpdateNode(t *testing.T) {
+	tests := []struct {
+		name          string
+		node          v1.Node
+		existingNodes *NodeStore
```
davidor, line 22: Is there any reason why `existingNodes` and `wantNodes` are a `NodeStore`? I think that we could simplify this by using `[]metav1.ObjectMeta` instead.

pkg/kubernetes/nodestore_test.go
```diff
@@ -0,0 +1,294 @@
...
+			gotNodes := nodeStore.GetNodes()
+			assert.EqualValues(t, tt.wantNodes, gotNodes)
+		})
+	}
+}
+
+func Test_UnsetNodes(t *testing.T) {
+	tests := []struct {
+		name          string
+		node          v1.Node
```
davidor, line 216: `UnsetNode` receives a name, not a `v1.Node` so the tests could be simplified a bit if we changed this to be a string instead.

controllers/datadogagent/controller_test.go
```diff
@@ -2883,6 +2885,140 @@ func TestReconcileDatadogAgent_Reconcile(t *testing.T) {
...
+
+			gotNodes := &corev1.NodeList{}
+			err = client.List(context.TODO(), gotNodes)
+			require.NoError(t, err, "Node with matching profile label not found")
+
+			for _, node := range gotNodes.Items {
+				if val, ok := tt.expectProfileLabel[node.Name]; ok {
+					if val {
+						assert.Equal(t, node.Labels[agentprofile.ProfileLabelKey], "true")
+						assert.Contains(t, node.Labels, agentprofile.ProfileLabelKey)
```
davidor, line 3012: Nit: This one is redundant. If the assert above is true this one has to be too.

davidor (approved):
:+1:

I just added some comments about a couple of things that could be simplified in the tests.

PR #1027, opened 2023-12-28
Decrease log verbosity level

khewonc:
### What does this PR do?

Decreases the log verbosity level for some logs

### Motivation

Debugging purposes. When using semantically named logging levels, the most verbose we can get with zapr is `debug` which is equivalent to `V(1)` in logr. Moving some logging in the operator from `V(2)` to `V(1)` will allow more verbose operator logs when using `loglevel=debug` without having to create new numbered levels as described in https://pkg.go.dev/github.com/go-logr/zapr#readme-increasing-verbosity.

### Additional Notes

### Minimum Agent Versions

* Agent: n/a
* Cluster Agent: n/a

### Describe your test plan

Spin up the operator using the helm chart with `--set logLevel="debug"`. After applying up a DatadogAgent, the operator should print out `DEBUG` log lines such as `{"level":"DEBUG","ts":"2023-12-28T22:03:14Z","logger":"controllers.DatadogAgent","msg":"dependencies.store Add object to update","datadogagent":"default/datadog","obj.namespace":"default","obj.name":"datadog-cluster-agent","obj.kind":"services"}` that was changed in this PR. When the operator is spun up without `--set logLevel="debug"`, the operator logs shouldn't show any `DEBUG` log lines.

### Checklist

### DIFF
--- a/controllers/datadogagent/controller.go
+++ b/controllers/datadogagent/controller.go
@@ -192,7 +192,7 @@ func (r *Reconciler) reconcileInstance(ctx context.Context, logger logr.Logger, 
 	// Now create/update dependencies
 	errs = append(errs, depsStore.Apply(ctx, r.client)...)
 	if len(errs) > 0 {
-		logger.V(2).Info("Dependencies apply error", "errs", errs)
+		logger.V(1).Info("Dependencies apply error", "errs", errs)
 		return result, errors.NewAggregate(errs)
 	}
 	// -----------------------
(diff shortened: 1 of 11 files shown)


celenechang:
@khewonc thanks - while I agree in principle, since we use DEBUG logging on our staging account I wonder if this will significantly increase logs volume

celenechang:
Closing for now, will look into a different approach

PR #1034, opened 2024-01-04
[Onboarding telemetry] Set install id, install time and install type env variables

liliyadd:
### What does this PR do?

When installing Datadog Agent, set INSTALL ID, INSTALL TIME and INSTALL TYPE env variables on Cluster Agent and Trace Agent. This information will be used to map first traces sent by the application to a specific Agent installation to build time-to-value KPI. https://docs.google.com/document/d/14vsrCbnAKnXmJAkacX9I6jKPGKmxsq0PKUb3dfiZpWE/edit?pli=1#heading=h.8d3o7vtyu1y1

Same logic for Helm is in https://github.com/DataDog/helm-charts/pull/1263/files#diff-a80e1dad44b204bcd75eed397bfa4a89f944c4804baebd6f988392094f058dc2

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

- Unit tests.
- Dev image of the Operator to install Agent in staging cluster.

### Checklist


controllers/datadogagent/component/agent/default.go
```diff
@@ -307,6 +307,25 @@ func envVarsForCoreAgent(dda metav1.Object) []corev1.EnvVar {
 	return append(envs, commonEnvVars(dda)...)
 }
 
+func envVarsForTraceAgent(dda metav1.Object) []corev1.EnvVar {
```
liliyadd, line 310: Is this the right place to add new env variables for Trace Agent? Looks like this codepath is not covered by agent_test.go tests.

controllers/datadogagent/component/clusteragent/default.go
```diff
@@ -144,6 +146,18 @@ func defaultEnvVars(dda metav1.Object) []corev1.EnvVar {
...
+			Value: component.AgentInstallId,
+		},
+		{
+			Name:  apicommon.DDInstrumentationInstallTime,
+			Value: component.AgentInstallTime,
+		},
+		{
+			Name:  apicommon.DDInstrumentationInstallType,
+			Value: component.DefaultAgentInstallType,
+		},
```
CharlyF, line 160: Are these env vars not specific to the APM use case?

controllers/datadogagent/component/clusteragent/default.go
```diff
@@ -17,6 +17,8 @@ import (
 	apicommon "github.com/DataDog/datadog-operator/apis/datadoghq/common"
 	apicommonv1 "github.com/DataDog/datadog-operator/apis/datadoghq/common/v1"
 	apiutils "github.com/DataDog/datadog-operator/apis/utils"
+
+	//componentutils "github.com/DataDog/datadog-operator/pkg/controllers/datadogagent/component"
```
CharlyF, line 21:
```suggestion
```

controllers/datadogagent/component/utils.go
```diff
@@ -27,6 +30,15 @@ import (
 
 const (
 	localServiceDefaultMinimumVersion = "1.22-0"
+
+	DefaultAgentInstallType = "k8s_manual"
```
CharlyF, line 34: how is this used? shouldn't this be `datadog-operator`?

CharlyF (changes requested):

controllers/datadogagent/component/clusteragent/default.go
liliyadd: Yes, they power APM Onboarding KPIs. Should we add _APM_ prefix to the env variable names?

controllers/datadogagent/component/utils.go
liliyadd: It's used for communicating to the Trace Agent the injection type that the client libraries were injected with. k8s_manual is the default value and will be overwritten by Cluster Agent if either local library injection or Single Step Instrumentation occur.

We only need this env variable until Agent 7.51.0 version is released, since we already merged a change to Trace Agent to handle properly the case when install type is not set.

controllers/datadogagent/component/clusteragent/default.go
CharlyF: no need if the instrumentation metadata naming has a consensus across product (wherever that could be used). However, the pattern we use is to only add feature specific configs when the config is enabled.  In this case, APM and the cluster agent are not yet tied (will be done in SSI), so while it'd be best to add more logic, this should be OK since we want APM to enabled.

CharlyF (approved):

PR #1066, opened 2024-02-01
Correct a malformed example command in installation.md

dgoffredo:
One of the `kubectl` commands in this installation guide passed a non-path parameter to `kubectl apply -f`. I'm not sure what the `agent_spec=` prefix is supposed to do, but `kubectl` rejects it on my cluster.


alai97 (approved):
Looks good for docs.

celenechang (approved):

PR #1070, opened 2024-02-06
[profiles] Replace custom node cache with k8s client one

davidor:
### What does this PR do?

Replaces the custom node cache with the k8s client one.

This should reduce memory usage because we were relying on the k8s client cache for some calls anyway.

### Checklist


controllers/datadogagent/controller_reconcile_agent.go
```diff
@@ -271,18 +271,42 @@ func (r *Reconciler) labelNodesWithProfiles(ctx context.Context, profilesByNode
...
+		// the node, it should be added.
 		if !isDefaultProfile && !profileLabelExists {
-			node.Labels[agentprofile.ProfileLabelKey] = "true"
-			if err := r.client.Update(ctx, node); err != nil {
-				return err
+			newLabels = make(map[string]string, len(node.Labels)+1)
+			for label, value := range node.Labels {
+				newLabels[label] = value
 			}
+			newLabels[agentprofile.ProfileLabelKey] = "true"
```
khewonc, line 294: What are your thoughts on having the label value be `<profile-name>` or `<profile-namespace>-<profile-name>` instead of `true`?

khewonc (approved):

controllers/datadogagent/controller_reconcile_agent.go
davidor: Sounds good 👍 
Let's open a separate PR for that.

PR #1083, opened 2024-02-12
Add Datadog Agent Profiles

davidor:
### What does this PR do?

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist


controllers/datadogagent/controller_reconcile_v2.go
```diff
@@ -140,20 +143,31 @@ func (r *Reconciler) reconcileInstanceV2(ctx context.Context, logger logr.Logger
...
+		for provider := range providersList {
+			result, err = r.reconcileV2Agent(logger, requiredComponents, features, instance, resourceManagers, newStatus, provider, &profile)
+			if utils.ShouldReturn(result, err) {
+				return r.updateStatusIfNeededV2(logger, instance, newStatus, result, err)
+			}
+		}
+	}
+
+	if err = r.handleProfiles(ctx, profiles, profilesByNode, instance.Namespace); err != nil {
+		return reconcile.Result{}, err
```
levan-m, line 170: I'm a bit confused by error handling here as some errors are collected into `errs`, some lead to return from function.
In this case, why are we returning here (but not on line 156)?
What happens if some profile/provider combinations succeed in line 162, shouldn't we doe Create/Update/Cleanup dependencies?

controllers/datadogagent/controller_reconcile_v2.go
```diff
@@ -140,20 +143,31 @@ func (r *Reconciler) reconcileInstanceV2(ctx context.Context, logger logr.Logger
...
 		}
+		for provider := range providersList {
+			result, err = r.reconcileV2Agent(logger, requiredComponents, features, instance, resourceManagers, newStatus, provider, &profile)
+			if utils.ShouldReturn(result, err) {
+				return r.updateStatusIfNeededV2(logger, instance, newStatus, result, err)
+			}
+		}
+	}
+
+	if err = r.handleProfiles(ctx, profiles, profilesByNode, instance.Namespace); err != nil {
```
levan-m, line 169: This function does node labeling, after DSs are already create. What happens if DS creation is successful but labeling fails (or keeps failing)? Do we reconcile such nodes?

davidor: The error handling logic follows the same pattern as in other places of the code.

All the steps in `handleProfiles` should be successful. If there's an error in any of the steps, we return an error in the Reconcile function so that the request is retried later.

If it keeps failing, we would have the same problem as in any other part of the code that creates, updates or deletes any Kubernetes objects. There isn't much we can do. Although the only case where I think the node labeling could be failing permanently (and not other operations) is if RBACs are not properly set, but that should be taken care of by our installation charts.

controllers/datadogagent/controller_reconcile_v2.go
davidor: I don't think it makes a bit difference in the end because any error here means that the request needs to be requeued and retried. I think it's a bit clearer to ensure that all the agent daemonsets are created as expected and unnecessary ones deleted before trying to manage their dependencies.

khewonc: When adding introspection, I was also thinking we shouldn't block the rest of the reconcile even if there was an error with `handleProviders` since an error there would come from an issue with deleting old DaemonSets. Then, when merging main into the profiles branch, I tried to make minimal changes so the different approaches to error handling were kept. We can make the behavior for both consistent in another PR

levan-m (approved):

PR #1118, opened 2024-03-12
Add DatadogPodAutoscaler resource

vboulineau:
### What does this PR do?

Add `DatadogPodAutoscaler` CRD.

### Motivation

Product.

### Minimum Agent Versions

* Cluster Agent: v7.56

### Checklist


apis/datadoghq/v1alpha1/datadogautoscaling_types.go
```diff
@@ -0,0 +1,282 @@
...
+//     update:
+//       mode: Auto|DryRun
+//   recommender:
+//     name: containerMetrics
+//     containerMetrics:
+//       cpuUtilizationTarget: 60
+//   constraints:
+//     minReplicas: 1
+//     maxReplicas: 10
+//     containers:
```
xlucas, line 30: Aren't those missing from the CRD?

apis/datadoghq/v1alpha1/datadogautoscaling_types.go
```diff
@@ -0,0 +1,282 @@
...
+//       - name: "*"
+//         enabled: true
```
xlucas, line 32: I assume the purpose of this field is to factor in/out certain containers from recommendations? Does it apply to both vertical + horizontal dimensions?

apis/datadoghq/v1alpha1/datadogautoscaling_types.go
```diff
@@ -0,0 +1,282 @@
...
+//         requests:
```
xlucas, line 33: Should we have a parent key to indicate to what resource type this applies to? I guess it's memory here.

apis/datadoghq/v1alpha1/datadogautoscaling_types.go
```diff
@@ -0,0 +1,282 @@
...
+//   targetRef:
+//     apiVersion: apps/v1
+//     kind: Deployment
+//     name: test
+//   owner: local
+//   policy:
+//     update:
+//       mode: Auto|DryRun
+//   recommender:
+//     name: containerMetrics
```
xlucas, line 24: Maybe it sounds a bit generic, should we have something that is more self-describing about what's the recommender logic is? Seems we're headed towards multidimensional with watermarks.

config/crd/bases/v1/datadoghq.com_datadogautoscalings.yaml
```diff
@@ -0,0 +1,240 @@
...
+                    - updateTime
+                  type: object
+                vertical:
+                  description: Vertical is the status of the vertical scaling, if activated.
+                  properties:
+                    desiredResources:
+                      description: DesiredResources is the desired resources for containers
+                      items:
+                        properties:
+                          limits:
```
xlucas, line 186: same comment as above for the type of resource this applies to

apis/datadoghq/v1alpha1/datadogautoscaling_types.go
vboulineau: Yes that's something we need to clearly define, this is more of an example.

config/crd/bases/v1/datadoghq.com_datadogautoscalings.yaml
```diff
@@ -0,0 +1,240 @@
...
+                          description: CPUUtilizationTarget is the target CPU utilization for the containers.
+                          format: int32
+                          maximum: 100
+                          minimum: 0
+                          type: integer
+                      type: object
+                    name:
+                      description: Name is the name of the recommender to use.
+                      enum:
+                        - ContainerMetrics
```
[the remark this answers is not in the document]
vboulineau: Yes, we have two options here:
Option1: we put the schema with everything in the CRD and only allow a fixed set of values.

Option2: we allow arbitrary string and we use `allow-unknown-field` on a `setting` field.

config/crd/bases/v1/datadoghq.com_datadogautoscalings.yaml
vboulineau: YAML is generated so discarding this comment and keeping one in code.

config/crd/bases/v1/datadoghq.com_datadogautoscalings.yaml
```diff
@@ -0,0 +1,240 @@
...
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.6.1
+  creationTimestamp: null
+  name: datadogautoscalings.datadoghq.com
+spec:
+  group: datadoghq.com
+  names:
+    kind: DatadogAutoscaling
+    listKind: DatadogAutoscalingList
+    plural: datadogautoscalings
```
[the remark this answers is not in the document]
vboulineau: Makes sense to me `DatadogPodAutoscaler` or `DatadogAutoscaler`?

apis/datadoghq/v1alpha1/datadogautoscaling_types.go
xlucas: Yep that's what got me confused because I was thinking of an "inverted" HPA `ContainerResource` metric type for a moment.

config/crd/bases/v1/datadoghq.com_datadogautoscalings.yaml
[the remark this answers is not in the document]
xlucas: `DatadogPodAutoscaler` sounds good to me, just wondering if we are going to interact with `/scale` & `/status` subresource at some point to be able to scale arbitrary targets like the HPA. In hindsight maybe the second option will be more flexible 🤔

apis/datadoghq/v1alpha1/datadogautoscaling_types.go
vboulineau: It uses the include/exclude syntax from VPA/MPA. Should we consider that Vertical/Horizontal share the same settings about included/excluded containers? I don't think it's necessary the case, but I'm not sure where to put the equivalent selected containers for HPA, as if we think about the current `DatadogMetrics` with arbitrary query, it does not require any information.

apis/datadoghq/v1alpha1/datadogpodautoscaler_types.go
```diff
@@ -0,0 +1,405 @@
...
+
+	// DatadogPodAutoscalerDrynRunMode will skip the update of the target resource.
+	DatadogPodAutoscalerDrynRunMode DatadogPodAutoscalerUpdateMode = "DryRun"
+)
+
+// DatadogPodAutoscalerUpdatePolicy defines the policy to update target resource.
+type DatadogPodAutoscalerUpdatePolicy struct {
+	// Mode defines the mode of the update policy.
+	Mode DatadogPodAutoscalerUpdateMode `json:"mode,omitempty"`
+}
```
fisherevans, line 109: I think `policy.dryRun: true/false` will be confused with `policy.update.mode: DryRun/Auto`. I think we should align the `update` policy with the `upscale` and `downscale` policies, and make it `policy.update.strategy: disabled/auto`.

vboulineau: Implemented

apis/datadoghq/v1alpha1/datadogpodautoscaler_types.go
```diff
@@ -227,32 +240,81 @@ type DatadogPodAutoscalerScalingRule struct {
 	Match *DatadogPodAutoscalerScalingRuleMatch `json:"match,omitempty"`
 }
 
-// DatadogPodAutoscalerUpdateMode defines the mode of the update policy.
-// +kubebuilder:validation:Enum:=ContainerMetrics
-type DatadogPodAutoscalerRecommenderName string
+// DatadogPodAutoscalerTargetType defines the type of the target.
+// +kubebuilder:validation:Enum:=Resource;ContainerResource
```
mfpierre, line 244: is it explicit enough or should we explicitely say `PodResource` and `ContainerResource`?

vboulineau: I have the same hesitation, changed back and forth, one is closer to upstream (`Resource`) and the other one more explicit `PodResource`).

celenechang (approved):
approved, but why is the skip qa label added? IMO we should QA it for the release

vboulineau:
> approved, but why is the skip qa label added? IMO we should QA it for the release

We do have a dedicated QA for Autoscaling, no worries on that and there's no code in the operator to actually QA, everything is in DCA

celenechang:
@vboulineau OK thanks. right, and we have the sister PR to this one we can QA.

PR #1131, opened 2024-03-27
[PROCS-3710] Add ability to run  Processes features on Core Agent

daniel-taf:
### What does this PR do?

This PR adds the ability to run all Processes features on the core agent. This can be done through two ways:
- Using the commandline flag: `runProcessChecksOnCoreAgent`
- Overriding environment variables through `spec.override`. This takes precedence over the commandline flag.

Currently both the environment variables and the commandline flag are `false` by default as this as a beta feature that will only be used internally to begin with. The default will eventually be `true` and so there will need to be further changes to operator for removing the flag.

### Motivation

PROCS-3710

### Minimum Agent Versions

* Agent: v7.53.0
* Cluster Agent: N/A

### Describe your test plan

Each feature was tested in both configurations (core agent and process agent) on a local kind cluster.
Live Processes Example:
**Process Agent**
```
  features:
    liveProcessCollection:
      enabled: true
```
Results in:
```
❯ kubectl get pod datadog-agent-xlnnl -o json | jq '.spec.containers[].name'
"agent"
"trace-agent"
"process-agent"
❯ kubectl logs datadog-agent-xlnnl --container process-agent | grep process-agent
2024-03-28 18:25:29 UTC | PROCESS | INFO | (pkg/process/runner/runner.go:286 in Run) | Starting process-agent with enabled checks=[process rtprocess]
```
**Core Agent via flag**
```
#agent.yaml
  features:
    liveProcessCollection:
      enabled: true
```
```
# Add new flag to container command args
spec:
      containers:
      - command:
        - /manager
        args:
        - --enable-leader-election
        - --pprof
        - --webhookEnabled
        - --runProcessChecksOnCoreAgent
```
Results in:
```
❯ kubectl get pod datadog-agent-4l5bs -o json | jq '.spec.containers[].name'
"agent"
"trace-agent"
❯ kubectl logs datadog-agent-4l5bs | grep process-agent
2024-03-28 19:10:57 UTC | CORE | INFO | (pkg/process/runner/runner.go:286 in Run) | Starting process-agent with enabled checks=[process rtprocess]
```
**Core Agent via env var**
```
#agent.yaml
...
spec:
  override:
    nodeAgent:
      env:
          value: "true"
...
  features:
    liveProcessCollection:
      enabled: true
```
Results in:
```
❯ kubectl get pod datadog-agent-97n27 -o json | jq '.spec.containers[].name'
"agent"
"trace-agent"
❯ kubectl logs datadog-agent-97n27 | grep process-agent
Defaulted container "agent" out of: agent, trace-agent, init-volume (init), init-config (init)
2024-03-28 19:17:20 UTC | CORE | INFO | (pkg/process/runner/runner.go:286 in Run) | Starting process-agent with enabled checks=[process rtprocess]
```

**Core Agent without min version**
```
#agent.yaml
...
spec:
  override:
    nodeAgent:
      nodeAgent:
      image:
        name: agent
        tag: 7.52.0
      env:
          value: "true"
...
  features:
    liveProcessCollection:
      enabled: true
```
Results in:
```
❯ kubectl get pod datadog-agent-qxg7k -o json | jq '.spec.containers[].name'
"agent"
"trace-agent"
"process-agent"
❯ kubectl logs datadog-agent-qxg7k --container process-agent | grep checks
2024-03-28 19:39:59 UTC | PROCESS | INFO | (pkg/process/runner/runner.go:275 in Run) | Starting process-agent with enabled checks=[process rtprocess]
```

### Checklist

### DIFF
--- a/controllers/datadogagent/feature/livecontainer/feature.go
+++ b/controllers/datadogagent/feature/livecontainer/feature.go
@@ -6,18 +6,14 @@
 package livecontainer
 
 import (
-	"strconv"
-
 	corev1 "k8s.io/api/core/v1"
 
 	"github.com/DataDog/datadog-operator/apis/datadoghq/v1alpha1"
 	"github.com/DataDog/datadog-operator/apis/datadoghq/v2alpha1"
 	apiutils "github.com/DataDog/datadog-operator/apis/utils"
-	"github.com/DataDog/datadog-operator/pkg/utils"
 
 	apicommon "github.com/DataDog/datadog-operator/apis/datadoghq/common"
 	apicommonv1 "github.com/DataDog/datadog-operator/apis/datadoghq/common/v1"
-	"github.com/DataDog/datadog-operator/controllers/datadogagent/component"
 	"github.com/DataDog/datadog-operator/controllers/datadogagent/feature"
 	"github.com/DataDog/datadog-operator/controllers/datadogagent/feature/liveprocess"
 	"github.com/DataDog/datadog-operator/controllers/datadogagent/object/volume"
@@ -49,24 +45,6 @@ func (f *liveContainerFeature) ID() feature.IDType {
 	return feature.LiveContainerIDType
 }
 
-func (f *liveContainerFeature) overrideRunInCoreAgent(dda *v2alpha1.DatadogAgent) {
-	if nodeAgent, ok := dda.Spec.Override[v2alpha1.NodeAgentComponentName]; ok {
-		// Agent version must >= 7.53.0 to run feature in core agent
-		if nodeAgent.Image != nil && !utils.IsAboveMinVersion(component.GetAgentVersionFromImage(*nodeAgent.Image), liveprocess.RunInCoreAgentMinVersion) {
-			f.runInCoreAgent = false
-		} else {
-			for _, env := range nodeAgent.Env {
-				if env.Name == apicommon.DDProcessConfigRunInCoreAgent {
-					val, err := strconv.ParseBool(env.Value)
-					if err == nil {
-						f.runInCoreAgent = val
-					}
-				}
-			}
-		}
-	}
-}
-
 // Configure is used to configure the feature from a v2alpha1.DatadogAgent instance.
 func (f *liveContainerFeature) Configure(dda *v2alpha1.DatadogAgent) (reqComp feature.RequiredComponents) {
 	if dda.Spec.Features.LiveContainerCollection != nil && apiutils.BoolValue(dda.Spec.Features.LiveContainerCollection.Enabled) {
@@ -74,7 +52,7 @@ func (f *liveContainerFeature) Configure(dda *v2alpha1.DatadogAgent) (reqComp fe
 			apicommonv1.CoreAgentContainerName,
 		}
 
-		f.overrideRunInCoreAgent(dda)
+		f.runInCoreAgent = liveprocess.OverrideRunInCoreAgent(dda, f.runInCoreAgent)
 
 		if !f.runInCoreAgent {
 			reqContainers = append(reqContainers, apicommonv1.ProcessAgentContainerName)
--- a/controllers/datadogagent/feature/livecontainer/feature.go
+++ b/controllers/datadogagent/feature/livecontainer/feature.go
@@ -6,15 +6,20 @@
 package livecontainer
 
 import (
+	"strconv"
+
 	corev1 "k8s.io/api/core/v1"
 
 	"github.com/DataDog/datadog-operator/apis/datadoghq/v1alpha1"
 	"github.com/DataDog/datadog-operator/apis/datadoghq/v2alpha1"
 	apiutils "github.com/DataDog/datadog-operator/apis/utils"
+	"github.com/DataDog/datadog-operator/pkg/utils"
 
 	apicommon "github.com/DataDog/datadog-operator/apis/datadoghq/common"
 	apicommonv1 "github.com/DataDog/datadog-operator/apis/datadoghq/common/v1"
+	"github.com/DataDog/datadog-operator/controllers/datadogagent/component"
 	"github.com/DataDog/datadog-operator/controllers/datadogagent/feature"
+	"github.com/DataDog/datadog-operator/controllers/datadogagent/feature/liveprocess"
 	"github.com/DataDog/datadog-operator/controllers/datadogagent/object/volume"
 )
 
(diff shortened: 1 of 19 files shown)


controllers/datadogagent/feature/test/testsuite.go
```diff
@@ -98,16 +99,17 @@ func runTest(t *testing.T, tt FeatureTest, buildFunc feature.BuildFunc) {
 	var gotConfigure feature.RequiredComponents
 	var dda metav1.Object
 	var isV2 bool
+	featureOptions := &feature.Options{}
+	if tt.FeatureOptions != nil {
+		featureOptions = tt.FeatureOptions
+	}
+	featureOptions.Logger = logger
```
daniel-taf, line 106: Added this to be able to unit test behavior with certain feature options. Tried to maintain the same logger behavior. Let me know if not advisable for any reason 🙇

controllers/datadogagent/feature/orchestratorexplorer/feature.go
```diff
@@ -66,7 +66,7 @@ type orchestratorExplorerFeature struct {
 	processAgentRequired bool
 }
 
-const NoProcessAgentMinVersion = "7.51.0"
+const NoProcessAgentMinVersion = "7.51.0-0"
```
daniel-taf, line 69: Had to be done in order for this check to work properly with prereleases. Otherwise, `7.53.0-rc.1` >= `7.51.0` would evaluate as false. Also done this way [here](https://github.com/DataDog/datadog-operator/blob/7db14b9f28692d91a33d8a483169e11119c71171/controllers/datadogagent/feature/kubernetesstatecore/feature.go#L71).

controllers/datadogagent/feature/livecontainer/feature.go
```diff
@@ -28,26 +33,57 @@ func init() {
 func buildLiveContainerFeature(options *feature.Options) feature.Feature {
 	liveContainerFeat := &liveContainerFeature{}
 
+	if options != nil {
+		liveContainerFeat.runInCoreAgent = options.RunProcessChecksOnCoreAgent
```
robertjli, line 37: Do we need to add a version check here as well?

controllers/datadogagent/feature/liveprocess/feature.go
```diff
@@ -115,6 +168,16 @@ func (f *liveProcessFeature) manageNodeAgent(agentContainerName apicommonv1.Agen
...
 
+	// cgroups volume mount
+	cgroupsVol, cgroupsVolMount := volume.GetVolumes(apicommon.CgroupsVolumeName, apicommon.CgroupsHostPath, apicommon.CgroupsMountPath, true)
+	managers.VolumeMount().AddVolumeMountToContainer(&cgroupsVolMount, agentContainerName)
+	managers.Volume().AddVolume(&cgroupsVol)
+
+	// procdir volume mount
+	procdirVol, procdirVolMount := volume.GetVolumes(apicommon.ProcdirVolumeName, apicommon.ProcdirHostPath, apicommon.ProcdirMountPath, true)
+	managers.VolumeMount().AddVolumeMountToContainer(&procdirVolMount, agentContainerName)
+	managers.Volume().AddVolume(&procdirVol)
```
robertjli, line 179: Should we gate these mounts behind the `runInCoreAgent` enablement?

controllers/datadogagent/feature/livecontainer/feature.go
daniel-taf: I was working under the assumption that the option would only be used internally, however, I think I will move the current version check I have into the `Configure` function after line 77.  That way, we will need less changes down the line.

daniel-taf: No, we do not. The version check is only necessary when a version is specified in `spec.override`. Otherwise, latest is used.

controllers/datadogagent/feature/liveprocess/feature.go
daniel-taf: I added these here because they are required for the feature to work on any container when live containers is disabled explicitly (its usually enabled by default). It's a rare occurrence, but this allows the feature to be independent.

controllers/datadogagent/feature/liveprocess/feature.go
```diff
@@ -97,14 +134,30 @@ func (f *liveProcessFeature) ManageClusterAgent(managers feature.PodTemplateMana
...
 	return nil
 }
 
 // ManageNodeAgent allows a feature to configure the Node Agent's corev1.PodTemplateSpec
 // It should do nothing if the feature doesn't need to configure it.
 func (f *liveProcessFeature) ManageNodeAgent(managers feature.PodTemplateManagers, provider string) error {
-	f.manageNodeAgent(apicommonv1.ProcessAgentContainerName, managers, provider)
+	containerName := apicommonv1.CoreAgentContainerName
+	if !f.runInCoreAgent {
+		containerName = apicommonv1.ProcessAgentContainerName
```
levan-m, line 151: Nit - by moving line 158 here, could one move rest of the code in `manageNodeAgent` from both functions?


```suggestion
		containerName = apicommonv1.ProcessAgentContainerName
		managers.EnvVar().AddEnvVarToContainer(apicommonv1.CoreAgentContainerName, runInCoreAgentEnvVar)
```


```go
	runInCoreAgentEnvVar := &corev1.EnvVar{
		Name:  apicommon.DDProcessConfigRunInCoreAgent,
		Value: apiutils.BoolToString(&f.runInCoreAgent),
	}
	managers.EnvVar().AddEnvVarToContainer(agentContainerName, runInCoreAgentEnvVar)
```

controllers/datadogagent/feature/test/testsuite.go
levan-m: Thanks for calling out, looks fine to me!

controllers/datadogagent/feature/livecontainer/feature.go
daniel-taf: In this case, it seems the version check would need to be a little bit different than the one I have. We would need to check the value of this variable instead of the override.

daniel-taf: I updated the override function to handle this case.

controllers/datadogagent/feature/liveprocess/feature.go
```diff
@@ -28,19 +32,44 @@ func init() {
...
 }
 
+const RunInCoreAgentMinVersion = "7.53.0-0"
+
 // ID returns the ID of the Feature
 func (f *liveProcessFeature) ID() feature.IDType {
 	return feature.LiveProcessIDType
 }
 
+func (f *liveProcessFeature) overrideRunInCoreAgent(dda *v2alpha1.DatadogAgent) {
```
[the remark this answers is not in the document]
daniel-taf: I had it this way to be able to edit the `runInCoreAgent` member variable for each feature, however I reworked the function to not do that.

controllers/datadogagent/feature/liveprocess/feature.go
levan-m: Sorry for confusion! This is a nit and let you decide what's best but idea was to have common logic `ManageClusterAgent` and `ManageSingleContainerNodeAgent` in `manageNodeAgent` and only leave custom one in specific function to make distinction a bit more clear. Below is one way to do it 

```go
// ManageSingleContainerNodeAgent allows a feature to configure the Agent container for the Node Agent's corev1.PodTemplateSpec
// if SingleContainerStrategy is enabled and can be used with the configured feature set.
func (f *liveProcessFeature) ManageSingleContainerNodeAgent(managers feature.PodTemplateManagers, provider string) error {
	f.manageNodeAgent(apicommonv1.UnprivilegedSingleAgentContainerName, managers, provider)
	return nil
}

func (f *liveProcessFeature) ManageNodeAgent(managers feature.PodTemplateManagers, provider string) error {
	runInCoreAgentEnvVar := &corev1.EnvVar{
		Name:  apicommon.DDProcessConfigRunInCoreAgent,
		Value: apiutils.BoolToString(&f.runInCoreAgent),
	}
	if !f.runInCoreAgent {
		managers.EnvVar().AddEnvVarToContainer(apicommonv1.CoreAgentContainerName, runInCoreAgentEnvVar)
		f.manageNodeAgent(apicommonv1.ProcessAgentContainerName, managers, provider)
	} else {
		managers.EnvVar().AddEnvVarToContainer(apicommonv1.ProcessAgentContainerName, runInCoreAgentEnvVar)
		f.manageNodeAgent(apicommonv1.CoreAgentContainerName, managers, provider)
	}
	return nil
}

func (f *liveProcessFeature) manageNodeAgent(agentContainerName apicommonv1.AgentContainerName, managers feature.PodTemplateManagers, provider string) error {
	runInCoreAgentEnvVar := &corev1.EnvVar{
		Name:  apicommon.DDProcessConfigRunInCoreAgent,
		Value: apiutils.BoolToString(&f.runInCoreAgent),
	}
	managers.EnvVar().AddEnvVarToContainer(agentContainerName, runInCoreAgentEnvVar)

```

controllers/datadogagent/feature/livecontainer/feature.go
levan-m: Thanks, change looks good. We check both and override taking precedence which is how overrides work.

Do you think test is warranted for default tag value (might be tricky to write such test)?

levan-m (approved):

levan-m:
Change looks good to me; However please get approval from @celenechang as well since the setup is based on her proposal.

celenechang:
Made a few small changes on the last commit, merging for 1.6

PR #1138, opened 2024-04-11
[profiles][CECO-1085] Add priority class name to dap

khewonc:
### What does this PR do?

Add `priorityClassName` as a configurable field in DAP. This can override the `priorityClassName` in the node agent

### Motivation

Blocker for internal usage

### Additional Notes

### Minimum Agent Versions

* Agent: n/a
* Cluster Agent: n/a

### Describe your test plan

1. Enable DAP in the operator
2. Create DDA and DAP without `priorityClassName` to ensure it still works
3. Add `priorityClassName` to the DAP and check that the agent pod uses it (`kubectl describe pod xxx`)
4. Test that changing the `priorityClassName` in the DAP updates the agent pod
5. Remove the `priorityClassName` in the DAP and check that it's removed from the agent pod

### Checklist


apis/datadoghq/v1alpha1/datadogagentprofile_types.go
```diff
@@ -34,7 +34,12 @@ type Config struct {
 }
 
 type Override struct {
+	// Configure the basic configurations for an Agent container
+	// Valid Agent container names are: `agent`
 	Containers map[commonv1.AgentContainerName]*Container `json:"containers,omitempty"`
+
+	// If specified, indicates the pod's priority
```
celenechang, line 41: do you think we need more info in the docstring similar to [here](https://github.com/DataDog/datadog-operator/blob/main/apis/datadoghq/v2alpha1/datadogagent_types.go#L1022)?

pkg/agentprofile/agent_profile.go
```diff
@@ -111,12 +111,18 @@ func ComponentOverrideFromProfile(profile *datadoghqv1alpha1.DatadogAgentProfile
...
 
-	return v2alpha1.DatadogAgentComponentOverride{
+	profileComponentOverride := v2alpha1.DatadogAgentComponentOverride{
 		Name:       &overrideDSName,
 		Affinity:   affinityOverride(profile),
 		Containers: containersOverride(profile),
 		Labels:     labelsOverride(profile),
 	}
+
+	if priorityClassName := priorityClassNameOverride(profile); priorityClassName != nil {
```
levan-m, line 121: nit - this function is called `ComponentOverrideFromProfile` while the `priorityClassNameOverride` specifically checks `NodeAgentComponentName` which is confusing. Since we know that for foreseeable future DAP will work only for Agent component maybe we could drop/replace `Component` from the names.

levan-m (approved):

PR #1179, opened 2024-05-13
feat(origin detection): add OriginDetectionUnified setting

wdhif:
### What does this PR do?

Add a new Global setting for the Origin Detection Unified setting of the Agent. This setting is a global one as it control Origin Detection behaviour for multiple Operator features (APM and DogStatsD).

### Motivation

This is needed to support newest Agent features in the Operator.

### Minimum Agent Versions

* Agent: v7.54.0

### Describe your test plan

With the following manifest:
```
kind: DatadogAgent
metadata:
  name: datadog
spec:
  override:
    nodeAgent:
      image:
        tag: 7.54.0-rc.1-jmx
        pullPolicy: Always
      containers:
        agent:
          env:
          - name: "DD_TELEMETRY_ENABLED"
            value: "true"
    clusterAgent:
      image:
        tag: 7.54.0-rc.1
        pullPolicy: Always
    clusterChecksRunner:
      image:
        tag: 7.54.0-rc.1-jmx
        pullPolicy: Always
  global:
    clusterName: "wassim-minikube"
    originDetectionUnified:
      enabled: true
    kubelet:
      tlsVerify: false
    credentials:
      apiSecret:
        secretName: datadog-secret
        keyName: api-key
      appSecret:
        secretName: datadog-secret
        keyName: app-key
  features:
    orchestratorExplorer:
      enabled: true
    liveProcessCollection:
      enabled: true
    admissionController:
      enabled: true
    externalMetricsServer:
      enabled: true
```

You should see the following:
```
➜  originDetectionUnified k exec -it datadog-agent-8xc84 -- agent config | grep origin_detection_unified
Defaulted container "agent" out of: agent, trace-agent, process-agent, init-volume (init), init-config (init)
origin_detection_unified: true
```

With the following manifest:
```
kind: DatadogAgent
metadata:
  name: datadog
spec:
  override:
    nodeAgent:
      image:
        tag: 7.54.0-rc.1-jmx
        pullPolicy: Always
      containers:
        agent:
          env:
          - name: "DD_TELEMETRY_ENABLED"
            value: "true"
    clusterAgent:
      image:
        tag: 7.54.0-rc.1
        pullPolicy: Always
    clusterChecksRunner:
      image:
        tag: 7.54.0-rc.1-jmx
        pullPolicy: Always
  global:
    clusterName: "wassim-minikube"
    kubelet:
      tlsVerify: false
    credentials:
      apiSecret:
        secretName: datadog-secret
        keyName: api-key
      appSecret:
        secretName: datadog-secret
        keyName: app-key
  features:
    orchestratorExplorer:
      enabled: true
    liveProcessCollection:
      enabled: true
    admissionController:
      enabled: true
    externalMetricsServer:
      enabled: true
```

You should see the following:
```
➜  originDetectionUnified k exec -it datadog-agent-bv5rv -- agent config | grep origin_detection_unified
Defaulted container "agent" out of: agent, trace-agent, process-agent, init-volume (init), init-config (init)
origin_detection_unified: false
```

### Checklist


zhuminyi (approved):

apis/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -797,6 +797,10 @@ type GlobalConfig struct {
 	// +listType=set
 	Tags []string `json:"tags,omitempty"`
 
+	// OriginDetectionUnified enables Unified mechanism for Origin Detection
+	// +optional
+	OriginDetectionUnified *bool `json:"originDetectionUnified,omitempty"`
```
clamoriniere, line 802: We usually put in the name the suffix `enabled` like `OriginDetectionUnifiedEnabled` or we add a sub option `enabled` like
```yaml
global:
  originDetectionUnified:
    enabled: true
```

the second solution is when we think that we might add later other option linked to "originDetectionUnified".
If we are not sure the best is to use the second option.

apis/datadoghq/v2alpha1/datadogagent_default.go
```diff
@@ -142,6 +144,10 @@ func defaultGlobalConfig(ddaSpec *DatadogAgentSpec) {
 		ddaSpec.Global.LogLevel = apiutils.NewStringPointer(defaultLogLevel)
 	}
 
+	if ddaSpec.Global.OriginDetectionUnified == nil {
+		ddaSpec.Global.OriginDetectionUnified = apiutils.NewBoolPointer(defaultOriginDetectionUnified)
+	}
```
clamoriniere, line 149: by adding a default value here. it means that the DatadogAgent resource will be patch to add the info into the spec. 

For this specific option, I'm not sure it is the best solution since the default value is set to "false" . so it means that if later we set the default to `true`. it will not be taken into consideration. 

So usually with "default" an option when it is not an option on which we want to change the default value.

instead in the DSD feature logic you should check is the option is nil then it is equal to the default value.

apis/datadoghq/v2alpha1/datadogagent_types.go
wdhif: Done 👍 I did it using the second option just in case.

apis/datadoghq/v2alpha1/datadogagent_default.go
wdhif: Done, although it might be possible to optimize the code.

controllers/testutils/agent.go
```diff
@@ -344,9 +344,12 @@ func NewDatadogAgentWithGlobalConfigSettings(namespace string, name string) v2al
...
 		},
-		Registry:              apiutils.NewStringPointer("my-custom-registry"),
-		LogLevel:              apiutils.NewStringPointer("INFO"),
-		Tags:                  []string{"tagA:valA", "tagB:valB"},
+		Registry: apiutils.NewStringPointer("my-custom-registry"),
+		LogLevel: apiutils.NewStringPointer("INFO"),
+		Tags:     []string{"tagA:valA", "tagB:valB"},
+		OriginDetectionUnified: &v2alpha1.OriginDetectionUnified{
+			Enabled: apiutils.NewBoolPointer(false),
+		},
```
clamoriniere, line 352: Do we need this change? Or should we change a test to validate that the envvar was added properly to the `agent` container? where the dogstatsd server runs?

controllers/datadogagent/override/global.go
```diff
@@ -137,6 +137,13 @@ func applyGlobalSettings(logger logr.Logger, manager feature.PodTemplateManagers
 		}
 	}
 
+	if config.OriginDetectionUnified != nil && config.OriginDetectionUnified.Enabled != nil {
```
clamoriniere, line 140: Could you update the corresponding test in this file https://github.com/DataDog/datadog-operator/blob/4838b3cae44ea0329ba036b9f25a8fe4a6cf7f2e/controllers/datadogagent/override/global_test.go#L34

clamoriniere (commented):
the change looks good, however it can be good to add at least a unit-test

docs/configuration.v2alpha1.md
```diff
@@ -194,6 +194,7 @@ spec:
 | global.networkPolicy.dnsSelectorEndpoints | DNSSelectorEndpoints defines the cilium selector of the DNS server entity. |
 | global.networkPolicy.flavor | Flavor defines Which network policy to use. |
 | global.nodeLabelsAsTags | Provide a mapping of Kubernetes Node Labels to Datadog Tags. <KUBERNETES_NODE_LABEL>: <DATADOG_TAG_KEY> |
+| global.originDetectionUnified.enabled | Enabled enables Unified mechanism for Origin Detection. Default: false |
```
maycmlee, line 197:
```suggestion
| global.originDetectionUnified.enabled | Enabled enables unified mechanism for origin Detection. Default: false |
```

wdhif: Fixed 😄

controllers/datadogagent/override/global.go
wdhif: I've added a case 👍

controllers/testutils/agent.go
wdhif: I have just removed it 👍

docs/configuration.v2alpha1.md
maycmlee: Ah yes, thanks!

maycmlee (approved):

zhuminyi (approved):

apis/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -719,6 +719,14 @@ type Endpoint struct {
 	Credentials *DatadogCredentials `json:"credentials,omitempty"`
 }
 
+// OriginDetectionUnified define the origin detection unified mechanism behavior.
```
celenechang, line 722:
```suggestion
// OriginDetectionUnified defines the origin detection unified mechanism behavior.
```

apis/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -797,6 +805,10 @@ type GlobalConfig struct {
 	// +listType=set
 	Tags []string `json:"tags,omitempty"`
 
+	// OriginDetectionUnified define the origin detection unified mechanism behavior.
```
celenechang, line 808:
```suggestion
	// OriginDetectionUnified defines the origin detection unified mechanism behavior.
```

PR #1187, opened 2024-05-16
Fix agentConfig struct

mftoure:
### What does this PR do?
This PR adds the right `ID` field to the agentConfigData Struct

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist

### DIFF
--- a/pkg/remoteconfig/updater.go
+++ b/pkg/remoteconfig/updater.go
@@ -57,7 +57,8 @@ type RcServiceConfiguration struct {
 
 // DatadogAgentRemoteConfig contains the struct used to update DatadogAgent object from RemoteConfig
 type DatadogAgentRemoteConfig struct {
-	ID            string                       `json:"name"`
+	ID            string                       `json:""`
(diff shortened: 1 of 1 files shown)


pkg/remoteconfig/updater.go
```diff
@@ -57,7 +57,8 @@ type RcServiceConfiguration struct {
 
 // DatadogAgentRemoteConfig contains the struct used to update DatadogAgent object from RemoteConfig
 type DatadogAgentRemoteConfig struct {
-	ID            string                       `json:"name"`
+	ID            string                       `json:""`
```
celenechang, line 60:
```suggestion
	ID            string                       `json:"id"`
```

mftoure: The id is actually only present inside the metadata of the config, not in the config it self. That's why it will be populated later

PR #1192, opened 2024-05-20
[CECO-1183] Override install info tool version

khewonc:
### What does this PR do?

Add an env var that can override the node agent's install info tool version

### Motivation

CECO-1183

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

* In a cluster, check that the `tool_version` in the created install info configmap defaults to `datadog-operator`
* Set the env var `DD_TOOL_VERSION` with a non-empty string value in the operator pod
* After the new operator becomes leader, check that the install info config map `tool_version` value changed
* Restart the agent pod so it reads from the updated configmap
* Check that the tool version changed to whatever the value of `DD_TOOL_VERSION` is (`install_method_tool_version` when running `agent status`)

### Checklist

### DIFF
--- a/controllers/datadogagent/feature/enabledefault/feature_test.go
+++ b/controllers/datadogagent/feature/enabledefault/feature_test.go
@@ -1,0 +1,60 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+package enabledefault
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


controllers/datadogagent/feature/enabledefault/feature_test.go
```diff
@@ -0,0 +1,60 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+package enabledefault
```
levan-m, line 6: Thanks for adding a test!

levan-m (approved):

PR #1195, opened 2024-05-22
[CONTP-136] Add Language Detection Config to APM SSI

adel121:
### What does this PR do?

This Pull Request adds the capability to configure language detection for APM SSI instrumentation and sets default configurations to allow APM SSI users to benefit from language detection by default without the need to do any configurations.

### Motivation

Expose language detection feature for APM SSI through the datadog operator.

### Additional Notes

- Enabling language detection should require enabling APM SSI. This is indicated in the description of the new config option.

### Minimum Agent Versions

* Agent: v7.52.0
* Cluster Agent: v7.52.0

### Important Note

In case the user deploys the agent with the process check running in the core agent, language detection will not work because Live Process Collection is not enabled by default. 

In this case, the user should also enable live process collection on the core agent in order for it to work.

The user should have something like this:

```
kind: DatadogAgent
metadata:
  name: datadog
  namespace: system
spec:
  global:
    credentials:
      apiSecret:
        secretName: datadog-secret
        keyName: api-key
    kubelet:
      tlsVerify: false
  override:
    nodeAgent:
      env:
          value: "true"
  features:
    apm:
      enabled: true
      instrumentation:
        enabled: true
    liveProcessCollection:
      enabled: true
```

### Describe your test plan

The most straight forward way is to test deploying the agent using the operator in the following cases:
- Case 1:
   - APM enabled
   - SSI enabled
   - Expectation: language detection is enabled
- Case 2:
   - APM enabled
   - SSI disabled
   - Expectation: language detection is disabled
- Case 3:
   - APM disabled
   - Expectation: language detection is disabled
- Case 4:
   - APM enabled
   - SSI enabled
   - Language Detection disabled
   - Expectation: language detection is disabled

To whether language detection is enabled or not:
- Verify that no clusterrole is created for the cluster agent to allow listing, watching and patching deployments (usually it is called `apm-cluster-agent`)
- Verify that `language_detection.enabled` option is set to `false` in the core agent, the cluster agent and process agent. This can be checked by running `agent config` on each of these three containers. 

You should also ensure that you are able to deploy `v1alpha1` without any issues.

**Sample Testing for Case 1:**

Use the following template for datadog agent
```
kind: DatadogAgent
metadata:
  name: datadog
  namespace: system
spec:
  global:
    credentials:
      apiSecret:
        secretName: datadog-secret
        keyName: api-key
    kubelet:
      tlsVerify: false
  features:
    apm:
      enabled: true
      instrumentation:
        enabled: true
```

Asserting the clusterrole was created:
```
adelhajhassan@COMP-Y0K2MF67D1 datadog-operator % kubectl get clusterrole | grep apm
datadog-apm-cluster-agent                                              2024-05-23T10:31:30Z
```

Asserting the proper config for language detection:
```
adelhajhassan@COMP-Y0K2MF67D1 datadog-operator % kubectl exec datadog-agent-r2jnp -c agent -- agent config | grep -A1 language_detection
  language_detection:
    cleanup:
--
language_detection:
  enabled: true



adelhajhassan@COMP-Y0K2MF67D1 datadog-operator % kubectl exec datadog-agent-r2jnp -c process-agent -- agent config | grep -A1 language_detection
  language_detection:
    cleanup:
--
language_detection:
  enabled: true


adelhajhassan@COMP-Y0K2MF67D1 datadog-operator % kubectl exec datadog-cluster-agent-55db5d7f9d-cvfs4   -- agent config | grep -A1 language_detection
  language_detection:
    cleanup:
--
language_detection:
  enabled: true
```

*Testing that everything works*

Deploy a dummy deployment:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: python-process-deployment
  labels:
    app: python-process-app
spec:
  replicas: 3
  selector:
    matchLabels:
      app: python-process-app
  template:
    metadata:
      labels:
        app: python-process-app
    spec:
      containers:
        - name: python-process-container
          image: python:3.7 # Replace with your Python image
          command: ["python3", "-u", "your-python-script.py"]
          volumeMounts:
            - name: python-script-volume
              mountPath: /your-python-script.py
              subPath: your-python-script.py
      volumes:
        - name: python-script-volume
          configMap:
            name: python-script-configmap
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: python-script-configmap
data:
  your-python-script.py: |
    while True:
      pass
```

Wait for some time, and then check that the language annotations were added on the deployment:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    deployment.kubernetes.io/revision: "1"
    internal.dd.datadoghq.com/python-process-container.detected_langs: python
```

### Checklist


apis/datadoghq/common/envvar.go
```diff
@@ -83,6 +85,8 @@ const (
 	DDKubeResourcesNamespace                          = "DD_KUBE_RESOURCES_NAMESPACE"
 	DDKubeStateMetricsCoreConfigMap                   = "DD_KUBE_STATE_METRICS_CORE_CONFIGMAP_NAME"
 	DDKubeStateMetricsCoreEnabled                     = "DD_KUBE_STATE_METRICS_CORE_ENABLED"
+	DDLanguageDetectionEnabled                        = "DD_LANGUAGE_DETECTION_ENABLED"
+	DDLanguageDetectionReportingEnabled               = "DD_LANGUAGE_DETECTION_REPORTING_ENABLED"
```
adel121, line 89: I added these two Env Vars and sorted the existing list.

controllers/datadogagent/feature/apm/const.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+const (
+	apmRBACPrefix = "apm"
+)
+
+// getRBACResourceName return the RBAC resources name
+func getRBACResourceName(owner metav1.Object) string {
+	return fmt.Sprintf("%s-%s-%s", owner.GetName(), apmRBACPrefix, "cluster-agent")
+}
```
adel121, line 21: Not sure if it is advised to create a new clusterrole specific to APM, or if we should reuse the existing shared cluster agent clusterrole.

docs/configuration.v2alpha1.md
```diff
@@ -46,6 +46,7 @@ spec:
 | features.apm.instrumentation.disabledNamespaces | DisabledNamespaces disables injecting the Datadog APM libraries into pods in specific namespaces. |
 | features.apm.instrumentation.enabled | Enabled enables injecting the Datadog APM libraries into all pods in the cluster. Default: false |
 | features.apm.instrumentation.enabledNamespaces | EnabledNamespaces enables injecting the Datadog APM libraries into pods in specific namespaces. |
+| features.apm.instrumentation.languageDetection.enabled | Enabled enables language detection to automatically detect languages of user workloads (beta). Requires SingleStepInstrumentation.Enabled to be true Default: true |
```
jhgilbert, line 49:
```suggestion
| features.apm.instrumentation.languageDetection.enabled | Enabled enables language detection to automatically detect languages of user workloads (beta). Requires SingleStepInstrumentation.Enabled to be true. Default: true |
```

jhgilbert (approved):
Approved with one suggestion, thanks!

apis/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -151,6 +151,20 @@ type SingleStepInstrumentation struct {
 	// ex: "java": "v1.18.0"
 	// +optional
 	LibVersions map[string]string `json:"libVersions,omitempty"`
+
+	// Language detection currently only detects languages and adds them as annotations on deployments, but doesn't use these languages for injecting libraries to applicative pods.
+	// It requires Agent 7.52+ and Cluster Agent 7.52+
```
khewonc, line 156:
```suggestion
	// (Requires Agent 7.52.0+ and Cluster Agent 7.52.0+)
```
To match the existing descriptions that have `Requires Agent xxx`

controllers/datadogagent/feature/apm/const.go
khewonc: I believe we use both methods so either is fine by me

khewonc: I believe we use both methods so either is fine by me

controllers/datadogagent/feature/apm/feature.go
```diff
@@ -363,6 +392,19 @@ func (f *apmFeature) manageNodeAgent(agentContainerName apicommonv1.AgentContain
...
 	managers.Port().AddPortToContainer(agentContainerName, apmPort)
 
+	// APM SSI Language Detection
+	if f.shouldEnableLanguageDetection() {
+		managers.EnvVar().AddEnvVar(&corev1.EnvVar{
+			Name:  apicommon.DDLanguageDetectionEnabled,
+			Value: "true",
+		})
+
+		managers.EnvVar().AddEnvVarToContainer(apicommonv1.ProcessAgentContainerName, &corev1.EnvVar{
```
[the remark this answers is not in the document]
adel121: > Is the process agent required for this feature?

Yes process agent is required for this feature.
The process agent runs the process check which detects running processes and their respective languages. 
The core agent then gets these collected processes via the remote workloadmeta collector, and leverage these data to provide the rest of the feature execution.
I think I should update the list of required components to include the process agent.

> Also out of curiosity, would this be affected by the DD_PROCESS_CONFIG_RUN_IN_CORE_AGENT_ENABLED=true setting?

Could you please clarify your question, I don't get it.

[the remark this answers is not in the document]
khewonc: Since `DD_PROCESS_CONFIG_RUN_IN_CORE_AGENT_ENABLED=true` would make features like live processes, live containers, and process discovery run in the core agent container instead of the process agent container, I was wondering if `DD_PROCESS_CONFIG_PROCESS_COLLECTION_ENABLED` would also need to be moved to the core agent based on `DD_PROCESS_CONFIG_RUN_IN_CORE_AGENT_ENABLED`. Live processes can [add the env var to the core agent](https://github.com/DataDog/datadog-operator/blob/7667a52a94b75e2a30af4411deaa445819a1bc4e/controllers/datadogagent/feature/liveprocess/feature.go#L134), so if this feature is also enabled, I think we could have `DD_PROCESS_CONFIG_PROCESS_COLLECTION_ENABLED=true` in both core agent and process agent containers. Would that cause an issue?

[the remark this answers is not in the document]
adel121: I don't think having `DD_PROCESS_CONFIG_COLLECTION_ENABLED=true` in both the core agent and the process agent can cause an issue, as long as process collection is still done correctly and process data are being (finally) populated in the workloadmeta store of the core agent (regardless of whether the process check was executed in the process agent or in the core agent).

controllers/datadogagent/feature/apm/feature.go
```diff
@@ -125,6 +133,11 @@ func (f *apmFeature) Configure(dda *v2alpha1.DatadogAgent) (reqComp feature.Requ
 				},
 			},
 		}
+
+		if f.shouldEnableLanguageDetection() {
+			reqComp.Agent.Containers = append(reqComp.Agent.Containers, apicommonv1.ProcessAgentContainerName)
```
AliDatadog, line 138: I think we can use the feature without the process-agent

controllers/datadogagent/feature/apm/rbac.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/DataDog/datadog-operator/pkg/kubernetes/rbac"
+	rbacv1 "k8s.io/api/rbac/v1"
+)
+
+func getLanguageDetectionRBACPolicyRules() []rbacv1.PolicyRule {
+	rbacRules := []rbacv1.PolicyRule{
+		{
+			APIGroups: []string{rbac.AppsAPIGroup},
+			Resources: []string{
+				// Currently only deployments are supported
```
AliDatadog, line 18: Note: if we want to avoid modifying the operator everytime we add a supported resource, it's better to add them all

controllers/datadogagent/feature/apm/feature.go
adel121: Refer to [this](https://github.com/DataDog/datadog-operator/pull/1195#discussion_r1613558536)

I think no change is needed.

controllers/datadogagent/feature/apm/rbac.go
adel121: I think it would be better to stick to the minimum privilege principle, we should not add a permission that is not currently used. 

To add, I assume adding support for new workload types will be done in the same release, so it would be needed to update the operator only once. 

And finally, it is better to keep the operator coherent with helm. 

WDYT?

controllers/datadogagent/feature/apm/feature.go
AliDatadog: Alright but here we mark the container as required. If it becomes possible to collect processes without the process-agent, IMO we should have an option to do it without requiring the component.

controllers/datadogagent/feature/apm/rbac.go
AliDatadog: No strong opinion here. IMO we'll add the support very soon and it can be confusing for some users that might upgrade the agent and not the operator

controllers/datadogagent/feature/apm/feature.go
adel121: Makes sense @AliDatadog 
I updated the PR to distinguish between the two cases.

robertjli: 👋 With `DD_PROCESS_CONFIG_RUN_IN_CORE_AGENT_ENABLED` set to true, it is indeed possible to run the process check in the core agent _without the process agent_. So in this case the container should not be required, in fact it may result in CLB because the process agent will just quit immediately.

controllers/datadogagent/feature/apm/feature.go
```diff
@@ -363,6 +402,26 @@ func (f *apmFeature) manageNodeAgent(agentContainerName apicommonv1.AgentContain
...
+
+		managers.EnvVar().AddEnvVar(&corev1.EnvVar{
+			Name:  apicommon.DDLanguageDetectionEnabled,
+			Value: "true",
+		})
+
+		managers.EnvVar().AddEnvVarToContainer(processCheckContainerName, &corev1.EnvVar{
+			Name:  apicommon.DDProcessCollectionEnabled,
+			Value: "true",
+		})
```
xornivore, line 422: We need to remove this

xornivore (dismissed):

controllers/datadogagent/feature/apm/feature.go
daniel-taf: It seems like this still needs updating. The process agent container should not be required if the checks are on the core agent. Unless you mean to do:
```go
if !f.processCheckRunsInCoreAgent && f.shouldEnableLanguageDetection() {}
```

Also, if you do want to run language detection on the core agent solely, there should be a check for the process check being enabled.

jhgilbert (approved):

controllers/datadogagent/feature/apm/feature.go
adel121: > Unless you mean to do 

Indeed, this is what I intended to do. Will update it.

> Also, if you do want to run language detection on the core agent solely, there should be a check for the process check being enabled.

Not sure about this. My understanding is that at the end we should be able to run language detection in the core agent regardless of whether it is coming from the process check (which requires enabling live process collection) or from the process collector (which should not ideally require live process collection to be enabled).

daniel-taf: Yes, I agree that should happen in the end. But, I'm not sure what agent version will be default here and I think the Operator can lag behind the latest release. Thus, there's a risk that the process agent won't be required and live processes isn't enabled resulting in language detection not working when expected. A possible alternative here is to have a version check for when we expect for it to work on the core agent (7.57.0 seems most likely).

daniel-taf (approved):

controllers/datadogagent/feature/apm/feature.go
daniel-taf: Also, is SSI enabled by default? Seems like no, but is there plans for that to be the case in the near future?

adel121: > Also, is SSI enabled by default? Seems like no, but is there plans for that to be the case in the near future?

I think the answer is no. 

But maybe #apm-onboarding can have a more decisive answer.

daniel-taf: We discussed off Github and decided that we would require the process agent for now and update this when equivalent support on the core agent is available.

daniel-taf (approved):

controllers/datadogagent/feature/apm/feature.go
```diff
@@ -363,6 +397,16 @@ func (f *apmFeature) manageNodeAgent(agentContainerName apicommonv1.AgentContain
 	}
 	managers.Port().AddPortToContainer(agentContainerName, apmPort)
 
+	// APM SSI Language Detection
+	if f.shouldEnableLanguageDetection() {
+
+		managers.EnvVar().AddEnvVar(&corev1.EnvVar{
```
khewonc, line 403: Does this env var need to be added to all containers? Can we limit it to only the containers that require it be set?

controllers/datadogagent/feature/apm/feature.go
```diff
@@ -125,6 +133,11 @@ func (f *apmFeature) Configure(dda *v2alpha1.DatadogAgent) (reqComp feature.Requ
 				},
 			},
 		}
+
+		if f.shouldEnableLanguageDetection() {
```
khewonc, line 137: Should `f.shouldEnableLanguageDetection()` be called after the `if apm.SingleStepInstrumentation != nil` if block below? I think `f.singleStepInstrumentation` will always be `nil` on this line

controllers/datadogagent/feature/apm/feature.go
khewonc: Does this depend on the process agent running any other features? For example, what would happen if all the process agent-based features were disabled, i.e. process discovery disabled, live containers disabled, etc. and language detection was enabled? Would the process agent crash?

apis/datadoghq/common/envvar.go
```diff
@@ -98,6 +98,8 @@ const (
 	DDKubeResourcesNamespace                             = "DD_KUBE_RESOURCES_NAMESPACE"
 	DDKubeStateMetricsCoreConfigMap                      = "DD_KUBE_STATE_METRICS_CORE_CONFIGMAP_NAME"
 	DDKubeStateMetricsCoreEnabled                        = "DD_KUBE_STATE_METRICS_CORE_ENABLED"
+	DDLanguageDetectionEnabled                           = "DD_LANGUAGE_DETECTION_ENABLED"
+	DDLanguageDetectionReportingEnabled                  = "DD_LANGUAGE_DETECTION_REPORTING_ENABLED"
```
khewonc, line 102: I don't believe `DDLanguageDetectionReportingEnabled` is used anywhere. Do we need this env var

khewonc (approved):
Approved with a nit about an unused env var

PR #1201, opened 2024-05-27
Add uncompressed layers support using either mounts or overlayfs

lebauce:
### What does this PR do?

Add features.sbom.containerImage.uncompressedLayersSupport.

### Motivation

GKE or EKS container runtimes are configured to discard uncompressed layers which
break container image scanning.
This PR adds a configuration setting to allow mounting the container image before scanning it,
or use the experimental overlayfs direct scan.

### Additional Notes

### Minimum Agent Versions

* Agent: v7.55.0
* Cluster Agent: vX.Y.Z

### Describe your test plan

Set up the agent CRD with:

```
spec:
  ...
  features:
    sbom:
      enabled: true
      containerImage:
        enabled: true
        uncompressedLayersSupport: true
        overlayFSDirectScan: true
```

On a K8s using containerd and the option `discard_uncompressed_layers=true` (which is the default) such as eks, gke or minikube, you should be able to see container images on the Datadog app.

### Checklist


docs/configuration.v2alpha1.md
```diff
@@ -134,6 +134,8 @@ spec:
 | features.remoteConfiguration.enabled | Enable this option to activate Remote Configuration. Default: true |
 | features.sbom.containerImage.analyzers | Analyzers to use for SBOM collection. |
 | features.sbom.containerImage.enabled | Enable this option to activate SBOM collection. Default: false |
+| features.sbom.containerImage.overlayFSDirectScan | Enable this option to enable experimental overlayfs direct scan. Default: false |
```
kayayarai, line 137:
```suggestion
| features.sbom.containerImage.overlayFSDirectScan | Enable this option to enable experimental overlayFS direct scan. Default: false |
```

kayayarai (approved):
Capitalization correction and 👍

controllers/datadogagent/feature/sbom/feature.go
```diff
@@ -161,6 +165,33 @@ func (f *sbomFeature) ManageNodeAgent(managers feature.PodTemplateManagers, prov
 			Value: strings.Join(f.containerImageAnalyzers, " "),
 		})
 	}
+	if f.containerImageUncompressedLayersSupport {
+		if f.containerImageOverlayFSDirectScan {
+			managers.EnvVar().AddEnvVarToContainer(apicommonv1.CoreAgentContainerName, &corev1.EnvVar{
```
levan-m, line 170: Is it intentional that the new env vars are added to only node agent container, while existing ones (`DDSBOMEnabled`, `DDSBOMContainerImageAnalyzers`, etc.) are added to all containers?

levan-m: Test doesn't make the difference clear either. This verifies that previous env vars are added to all containers, there is no test for the new env vars. Could you please add one so the intention is clear?
https://github.com/DataDog/datadog-operator/pull/1201/files#diff-62de4539307e0ac7e109b94bb59b9ad8faff84eae1258be8b76a8c7d7212b523R89-R90

levan-m (changes requested):
Left a comment to clarify env var setting and add a corresponding test. Please add test instructions in the PR description as well for QA.
Otherwise looks good to me and can be merged once above is addressed.

controllers/datadogagent/feature/sbom/feature.go
lebauce: Yes, it is intentional. This option currently only makes sense for the core agent.

lebauce: Added a test for it

levan-m: Thanks!

levan-m (approved):

PR #1317, opened 2024-07-25
[profiles] Reduce patch frequency

khewonc:
### What does this PR do?

Nodes were being patched every reconcile when there was no change in labels. This only patches nodes when there is a label change

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Test in staging:
* Check the workqueue graphs to see if metrics stay consistent
* Check that there aren't an enormous number of patch requests in the k8s requests by method/target object graphs

### Checklist


controllers/datadogagent/controller_reconcile_agent_test.go
```diff
@@ -1534,3 +1536,432 @@ func Test_removeStaleStatus(t *testing.T) {
 		})
 	}
 }
+
+func Test_labelNodesWithProfiles(t *testing.T) {
```
levan-m, line 1540: Thanks for adding comprehensive test!

levan-m (approved):

controllers/datadogagent/controller_reconcile_agent.go
```diff
@@ -283,30 +283,45 @@ func (r *Reconciler) labelNodesWithProfiles(ctx context.Context, profilesByNode
...
+		// Remove old profile label key if it is present
+		if _, oldProfileLabelExists := node.Labels[agentprofile.OldProfileLabelKey]; oldProfileLabelExists {
+			labelsToRemove[agentprofile.OldProfileLabelKey] = true
 		}
 
-		// If the profile is not the default one and the label does not exist in
-		// the node, it should be added.
-		if !isDefaultProfile && !profileLabelExists {
-			newLabels[agentprofile.ProfileLabelKey] = profileNamespacedName.Name
+		if len(labelsToRemove) > 0 || len(labelsToAddOrChange) > 0 {
```
celenechang, line 310: is there a benefit to having this `||` instead of two separate conditionals?

celenechang (approved):
one small comment, otherwise looks good!

controllers/datadogagent/controller_reconcile_agent.go
khewonc: The first for loop for `labelsToRemove` goes through all the node labels but the second one for `labelsToAddOrChange` doesn't. If we separated them, we'd need to loop through `node.Labels` twice

celenechang: ah i see, thanks!

PR #1340, opened 2024-08-02
[Orch] Add CRD RC handler

JLineaweaver:
### What does this PR do?

Adds remote config support for custom resources in the operator. This new code will merge the user defined configuration with the configuration stored in remote config. Operator users with remote config enabled should no longer require a custom operator configuration to send custom resources.

### Motivation

### Additional Notes

### Minimum Agent Versions

* Cluster Agent: v7.42

### Describe your test plan

- RC is only available on datad0g.com.  Update operator to use staging api key and app key. Update `DD_SITE` to `datad0g.com`
- Make sure operator has list and watch permission on * / *
- Update DatadogAgent to include
```
    orchestratorExplorer:
      enabled: true
    remoteConfiguration:
      enabled: true
```
- Turn on/off custom resources indexing in UI
- You should be able to see this section in `status` of DatadogAgent
```
  remoteConfigConfiguration:
    features:
      orchestratorExplorer:
        customResources:
        - datadoghq.com/v2alpha1/datadogagents
        - mutations.gatekeeper.sh/v1/assign
        - kueue.x-k8s.io/v1beta1/admissionchecks
        - test.com/v1/josh
        ........
```

### Checklist


pkg/remoteconfig/orchestrator_k8s_crd.go
```diff
@@ -0,0 +1,126 @@
...
+
+	dda, err := r.getDatadogAgentInstance(ctx)
+	if err != nil {
+		r.logger.Error(err, "Failed to get updatable agents")
+		return
+	}
+
+	if err := r.crdUpdateInstanceStatus(dda, mergedUpdate); err != nil {
+		r.logger.Error(err, "Failed to update status")
+		applyStatus(configIDs[len(configIDs)-1], state.ApplyStatus{State: state.ApplyStateError, Error: err.Error()})
```
timothyalexandersoftware, line 45: This looks like it will only update the ApplyState of the last config that was sent. So if three configs get sent, the first two would be "Unacknowledged" and the last one would be "Error". Not sure if this is intended or not. 

For more info on this feature, you can the ApplyState of configs in the Clients page of the [RC Admin app](https://remote-config-admin.us1.prod.dog/?orgId=2&product=AGENT_CONFIG). 
image

pkg/remoteconfig/updater.go
```diff
@@ -59,11 +59,12 @@ type RcServiceConfiguration struct {
...
-	Name          string                       `json:"name,omitempty"`
-	CoreAgent     *CoreAgentFeaturesConfig     `json:"config,omitempty"`
-	SystemProbe   *SystemProbeFeaturesConfig   `json:"system_probe,omitempty"`
-	SecurityAgent *SecurityAgentFeaturesConfig `json:"security_agent,omitempty"`
+	ID            string                        `json:"id,omitempty"`
+	Name          string                        `json:"name,omitempty"`
+	CoreAgent     *CoreAgentFeaturesConfig      `json:"config,omitempty"`
+	SystemProbe   *SystemProbeFeaturesConfig    `json:"system_probe,omitempty"`
+	SecurityAgent *SecurityAgentFeaturesConfig  `json:"security_agent,omitempty"`
+	CRDs          *CustomResourceDefinitionURLs `json:"crds,omitempty"`
```
timothyalexandersoftware, line 67: Take this with a grain of salt since I'm not too familiar with the operator, but this looks to me like structurally, configs are broken out by agent binary. Should the CRD configs be wrapped inside a ClusterAgentFeatureConfig type?

JLineaweaver: Yeah that's a great call. I think this is a good idea for future work

pkg/remoteconfig/orchestrator_k8s_crd.go
JLineaweaver: Hmm I have to double check this. I was just following the existing pattern here.  Above it does apply status to each config. I need to do some testing to see

pkg/remoteconfig/updater.go
JLineaweaver: [51d2cf9](https://github.com/DataDog/datadog-operator/pull/1340/commits/51d2cf97370b4a5278df24135b119e4816e96415)

pkg/remoteconfig/updater.go
```diff
@@ -137,7 +146,7 @@ func (r *RemoteConfigUpdater) Start(apiKey string, site string, clusterName stri
 		"",
 		r.serviceConf.baseRawURL,
 		r.serviceConf.hostname,
-		[]string{fmt.Sprintf("cluster_name:%s", r.serviceConf.clusterName)},
+		r.getTags,
```
JLineaweaver, line 149: This is a required breaking change

pkg/remoteconfig/updater.go
```diff
@@ -155,7 +162,7 @@ func (r *RemoteConfigUpdater) Start(apiKey string, site string, clusterName stri
 	rcClient, err := client.NewClient(
 		rcService,
 		client.WithAgent("datadog-operator", version.Version),
-		client.WithProducts(state.ProductAgentConfig),
+		client.WithProducts(state.ProductAgentConfig, crdRcProduct),
```
lebauce, line 165: Could it be a single product ?
In both case, this is to manipulate agent config

JLineaweaver: These are completely different products in remote config configuring different parts of the agent. This here is referring to the RC product

lebauce: What I'm afraid of is that we have 2 RC products to configure the agent. What happens if you enable CSM VM through AgentConfig RC product but set `features.sbom.containerImage.enabled` to true in the CRD RC product ?

levan-m: I'm also confused by this as well. `ProductOrchestratorK8sCRDs` doesn't even imply whole Orchestrator Explorer feature, so are we going to have `ProductOrchestrator` later? 
With this approach we may end up with 30+ products. What are the implication of such granularity? 
cc @timothyalexandersoftware

timothyalexandersoftware: I think I'd need to understand a bit more about the existing product(s) using `ProductAgentConfig` and how they are configuring the agent, as compared to what Josh is introducing. If these are truly just different configurations for the agent that are getting configured via two different products, there is indeed an opportunity to consolidate. 

This would add a bit of complexity for managing the release of different features though -- e.g. if both features use the same product, it would be non-trivial to have a killswitch to turn off the K8s CRDs product/feature without affecting the security product and vice versa.

levan-m: Checked with @timothyalexandersoftware and for now this approach is acceptable but we may have to revisit it later and consolidate these products into one.

internal/controller/datadogagent/feature/orchestratorexplorer/feature.go
```diff
@@ -92,9 +96,12 @@ func (f *orchestratorExplorerFeature) Configure(dda *v2alpha1.DatadogAgent) (req
 			Containers: reqContainers,
 		}
 
-		if orchestratorExplorer.Conf != nil {
+		if orchestratorExplorer.Conf != nil || len(orchestratorExplorer.CustomResources) > 0 {
 			f.customConfig = orchestratorExplorer.Conf
-			hash, err := comparison.GenerateMD5ForSpec(f.customConfig)
+
+			// Used to force restart of DCA
+			// use entire orchestratorExplorer to handle custom config and CRDs
```
celenechang, line 103: even if `scrubContainers` is changed, dca will restart; is that ok?

go.sum
```diff
@@ -65,6 +65,8 @@ github.com/DataDog/datadog-agent/pkg/proto v0.55.0-rc.10 h1:ERkVmUoDPttyVKSCJM1f
 github.com/DataDog/datadog-agent/pkg/proto v0.55.0-rc.10/go.mod h1:gHkSUTn6H6UEZQHY3XWBIGNjfI3Tdi0IxlrxIFBWDwU=
 github.com/DataDog/datadog-agent/pkg/remoteconfig/state v0.55.0-rc.10 h1:nwJ2JWfjCmf6tpJD1RYHh4JV5HO2Njg6smxOGi8MyOE=
 github.com/DataDog/datadog-agent/pkg/remoteconfig/state v0.55.0-rc.10/go.mod h1:3yFk56PJ57yS1GqI9HAsS4PSlAeGCC9RQA7jxKzYj6g=
+github.com/DataDog/datadog-agent/pkg/remoteconfig/state v0.59.0-rc.1 h1:CIXKFvUsp5zgE+egvx+QrRTw8r54FMPZVu+z25UNqWs=
```
celenechang, line 68: are there any changes between rc1 and rc4?

internal/controller/datadogagent/feature/orchestratorexplorer/feature.go
```diff
@@ -125,6 +133,24 @@ func (f *orchestratorExplorerFeature) Configure(dda *v2alpha1.DatadogAgent) (req
 	return reqComp
 }
 
+func (f *orchestratorExplorerFeature) mergeConfigs(ddaSpec *v2alpha1.DatadogAgentSpec, ddaStatus *v2alpha1.DatadogAgentStatus) {
+	if ddaStatus.RemoteConfigConfiguration == nil || ddaStatus.RemoteConfigConfiguration.Features == nil || ddaStatus.RemoteConfigConfiguration.Features.OrchestratorExplorer == nil {
```
celenechang, line 137: could also add `|| ddaStatus.RemoteConfigConfiguration.Features.OrchestratorExplorer.CustomResources == nil`

internal/controller/datadogagent/feature/orchestratorexplorer/feature_test.go
```diff
@@ -142,7 +142,17 @@ func orchestratorExplorerClusterAgentWantFuncV2() *test.ComponentTest {
 			customConfig := v2alpha1.CustomConfig{
 				ConfigData: apiutils.NewStringPointer(customConfDataV2),
 			}
-			hash, err := comparison.GenerateMD5ForSpec(&customConfig)
+			truePointer := true
```
celenechang, line 145: technically not a pointer..

pkg/remoteconfig/updater.go
```diff
@@ -219,6 +228,55 @@ func getEndpoint(prefix, site string) string {
 	return prefix + defaultSite
 }
 
+// getAndUpdateDatadogAgent is used to prevent race conditions when update the DDA's status
```
celenechang, line 231:
```suggestion
// getAndUpdateDatadogAgent is used to prevent race conditions when updating the DDA's status
```

pkg/remoteconfig/orchestrator_k8s_crd.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+	}
+
+	if len(crds) == 0 {
+		r.logger.Info("No CRDs received")
+		return DatadogAgentRemoteConfig{}, nil
+	}
+
+	// Merge configs
+	var finalConfig DatadogAgentRemoteConfig
```
celenechang, line 70: i wonder if it would be clearer to just return `ClusterAgentFeaturesConfig` instead of the entire `DatadogAgentRemoteConfig`

internal/controller/datadogagent/feature/orchestratorexplorer/feature_test.go
celenechang, line 145:
```suggestion
			trueValue := true
```

we do it like this elsewhere

pkg/remoteconfig/updater.go
```diff
@@ -63,6 +65,7 @@ type DatadogAgentRemoteConfig struct {
 	ID            string                       `json:"id,omitempty"`
 	Name          string                       `json:"name,omitempty"`
 	CoreAgent     *CoreAgentFeaturesConfig     `json:"config,omitempty"`
+	ClusterAgent  *ClusterAgentFeaturesConfig  `json:"cluster_agent,omitempty"`
```
levan-m, line 68: This structure is unmarshalled from the RawConfig here and 
https://github.com/DataDog/datadog-operator/blob/03611c4e0472e9f993aa07d30b9fe051207a2fc9/pkg/remoteconfig/updater.go#L330-L331

And this config won't contain ClusterAgentFeatureConfig so curious why is this added here?

pkg/remoteconfig/orchestrator_k8s_crd.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// Cleanup CRD duplicates and add to final config
+	crds = removeDuplicateStr(crds)
+	if finalConfig.ClusterAgent == nil {
+		finalConfig.ClusterAgent = &ClusterAgentFeaturesConfig{}
+	}
+	if finalConfig.ClusterAgent.CRDs == nil {
+		finalConfig.ClusterAgent.CRDs = &CustomResourceDefinitionURLs{}
+	}
+	finalConfig.ClusterAgent.CRDs.Crds = &crds
```
levan-m, line 80: Current implementation for security products unmarshals `configuration_order`  and uses to merge multiple configs:
https://github.com/DataDog/datadog-operator/blob/03611c4e0472e9f993aa07d30b9fe051207a2fc9/pkg/remoteconfig/updater.go#L323-L324

do we need similar logic for CRDs too?

pkg/remoteconfig/orchestrator_k8s_crd.go
kangyili: I think it follows the default pattern, which always returns the entire configuration. 
https://github.com/DataDog/datadog-operator/blob/03611c4e0472e9f993aa07d30b9fe051207a2fc9/pkg/remoteconfig/updater.go#L317

Ultimately, we need to adhere to the function `getAndUpdateDatadogAgentWithRetry`, which takes the complete configuration and an update function as its input.
https://github.com/DataDog/datadog-operator/blob/03611c4e0472e9f993aa07d30b9fe051207a2fc9/pkg/remoteconfig/updater.go#L253

internal/controller/datadogagent/feature/orchestratorexplorer/feature_test.go
kangyili: https://github.com/DataDog/datadog-operator/pull/1340/commits/c3fcffee51686d555c957949fda69fa35dfa48c2

internal/controller/datadogagent/feature/orchestratorexplorer/feature.go
kangyili: https://github.com/DataDog/datadog-operator/pull/1340/commits/c3fcffee51686d555c957949fda69fa35dfa48c2

go.sum
kangyili: There is only go mode change. The latest 2 commits: https://github.com/DataDog/datadog-agent/commits/pkg/remoteconfig/state/v0.59.0-rc.4/pkg/remoteconfig/state

pkg/remoteconfig/updater.go
levan-m: Sorry if my question wasn't clear. I'm confused why the `ClusterAgentFeaturesConfig` is added here.

This structure
https://github.com/DataDog/datadog-operator/blob/03611c4e0472e9f993aa07d30b9fe051207a2fc9/pkg/remoteconfig/updater.go#L64-L71

 is for RC product `ProductAgentConfig` supporting CoreAgent, SystemProbe and SecurityAgent features. It's subscribed here

https://github.com/DataDog/datadog-operator/blob/03611c4e0472e9f993aa07d30b9fe051207a2fc9/pkg/remoteconfig/updater.go#L181

While 

https://github.com/DataDog/datadog-operator/blob/03611c4e0472e9f993aa07d30b9fe051207a2fc9/pkg/remoteconfig/updater.go#L183

subscribes a different `ProductOrchestratorK8sCRDs`. So I'm curious why it's put under `DatadogAgentRemoteConfig` structure instead of having it's own structure?

levan-m (approved):

PR #1366, opened 2024-08-21
[PROCS-4294] Add ability to run Language Detection in Core Agent

daniel-taf:
### What does this PR do?

In Agent 7.57.0, the ability to run language detection in the core agent was introduced using the `DD_PROCESS_CONFIG_RUN_IN_CORE_AGENT_ENABLED` config: https://github.com/DataDog/datadog-agent/pull/27559. This PR updates the Operator to support that feature by not requiring the Process Agent container when the config is enabled.

### Motivation

PROCS-4294

### Additional Notes

* Live Process Collection is not needed for language detection.
* I updated the version check for using the `RUN_IN_CORE_AGENT` config from requiring 7.53 -> 7.57. Reasoning [here](https://github.com/DataDog/datadog-operator/pull/1366#discussion_r1727595438).

### Minimum Agent Versions

* Agent: 7.57.0
* Cluster Agent: N/A

### Describe your test plan

Unit tests were added and manual testing was performed on a kind cluster. A custom build of the operator was deployed with the following config:
```
kind: DatadogAgent
metadata:
  name: datadog
spec:
  override:
    nodeAgent:
      image:
        name: agent
        tag: 7.57.0-rc.3
        pullPolicy: IfNotPresent
      env:
          value: "true"
  global:
    clusterName: dtaf-qa
    registry: docker.io/datadog
    kubelet:
      # this fixes an issue in kind where the agent isn't able to start due to a hostname issue
      tlsVerify: false 
    credentials:
      apiSecret:
        secretName: datadog-secret
        keyName: api-key
  features:
    apm:
      enabled: true
      instrumentation:
        enabled: true
```
Verified the process agent container did not start:
```
❯ kubectl get pod datadog-agent-b5dgw -o json | jq '.spec.containers[].name'
"agent"
"trace-agent"
```
Verified the relevant WLM collector for language started:
```
❯ kubectl logs datadog-agent-b5dgw -c agent | grep local-process-collector
2024-08-22 17:33:38 UTC | CORE | INFO | (comp/core/workloadmeta/impl/store.go:554 in startCandidates) | workloadmeta collector "local-process-collector" started successfully
```
Verified language data was present in `workload-list`:
```
❯ agent workload-list
...
=== Entity process sources(merged):[local_process_collector] id: 23573 ===
----------- Entity ID -----------
PID: 23573
Namespace PID: 4318
Container ID:
Creation time: 2024-08-22 17:40:15 +0000 UTC
Language: python
===
```

### Checklist


controllers/datadogagent/feature/utils/utils.go
```diff
@@ -19,6 +19,7 @@ import (
 // Process Checks utils
 
 const RunInCoreAgentMinVersion = "7.53.0-0"
+const LanguageDetectionRunInCoreAgentMinVersion = "7.57.0-0"
```
daniel-taf, line 22: I'm considering collapsing these into one version we check and continuously update until we GA. This would greatly simplify the code.

controllers/datadogagent/feature/utils/utils.go
```diff
@@ -18,7 +18,7 @@ import (
 
 // Process Checks utils
 
-const RunInCoreAgentMinVersion = "7.53.0-0"
+const RunInCoreAgentMinVersion = "7.57.0-0"
```
daniel-taf, line 21: I just updated this version to the latest that supports language detection in the core agent and I think we should continue to update this until we GA (this is not a publicly documented feature yet). The code would be simplified as a result and I think there would be less headache in the future as we would need to worry about less versions. Curious if there any reasons not to do this.

daniel-taf: Adding for further context: this version check impacts the following: Process Collection, Process Discovery, Container Collection, and now Language Detection/APM SSI. The config itself controls whether these features run on the process agent or the core agent.

robertjli: > The only concern I can think of is negatively impacting our internal deployments using this config, but 7.57.0 should already be rolled out by then.

Will this impact beta customers for the run-in-core-agent feature which will be on 7.56?

daniel-taf: I dont think so. They would be on Operator 1.8.

daniel-taf: And, once operator 1.9 is released, I imagine 7.57 should already be released.

robertjli (approved):

controllers/datadogagent/feature/utils/utils.go
levan-m: What happens to the customers who have Agent version pinned to pre-7.57 version and upgrade to the new version of Operator?

daniel-taf: If the `RunInCoreAgent` feature is enabled then their agent deployment would revert to the prior setup. I think thats ok because this is not a documented feature for the Operator yet and where it is documented, its descirbed as experimental. Otherwise, nothing should happen.

levan-m (approved):

PR #1373, opened 2024-08-26
[profiles] Add update strategy overrides to profiles

khewonc:
### What does this PR do?

Adds update strategy overrides to profiles

### Motivation

https://datadoghq.atlassian.net/browse/CECO-1014

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

* Deploy operator with [profiles enabled](https://github.com/DataDog/datadog-operator/blob/main/docs/datadog_agent_profiles.md#enabling-datadogagentprofiles)
* Deploy DDA
* Create a profile with a custom update strategy that is different from what is set in the DDA
* Check the profile daemonset to see if the update strategy was overridden

### Checklist

### DIFF
--- a/apis/datadoghq/v1alpha1/datadogagentprofile_types.go
+++ b/apis/datadoghq/v1alpha1/datadogagentprofile_types.go
@@ -7,6 +7,7 @@ package v1alpha1
 
 import (
 	commonv1 "github.com/DataDog/datadog-operator/apis/datadoghq/common/v1"
+	"github.com/DataDog/datadog-operator/apis/datadoghq/v2alpha1"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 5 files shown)


apis/datadoghq/v1alpha1/datadogagentprofile_types.go
```diff
@@ -48,6 +49,10 @@ type Override struct {
 	// +optional
 	PriorityClassName *string `json:"priorityClassName,omitempty"`
 
+	// The deployment strategy to use to replace existing pods with new ones.
+	// +optional
+	UpdateStrategy *v2alpha1.UpdateStrategy `json:"updateStrategy,omitempty"`
```
khewonc, line 54: Should we use v2alpha1 directly or move `UpdateStrategy` to [common](https://github.com/DataDog/datadog-operator/blob/main/apis/datadoghq/common/v1/types.go)?

levan-m: Agree, thanks for calling out would have overlooked. I'll create a PR.

PR #1399, opened 2024-09-10
[cleanup] reorganize constant files

celenechang:
### What does this PR do?

Move some constant definitions
Provide descriptions of what constant files are used for
A bit of housekeeping, like fixing grammar on descriptions and alphabetizing chunks of constant definitions

### Motivation

Cleanup

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

This change is noop. Can run `make build` and `make test` to confirm

### Checklist


pkg/kubernetes/rbac/const.go
```diff
@@ -5,106 +5,108 @@
...
 
-	// Rbac resource kinds
+	// RBAC resource kinds (singular)
 
-	ClusterRoleKind    = "ClusterRole"
-	RoleKind           = "Role"
-	ServiceAccountKind = "ServiceAccount"
+	ClusterRoleBindingKind = "ClusterRoleBinding"
+	ClusterRoleKind        = "ClusterRole"
+	RoleKind               = "Role"
```
tbavelier, line 110: Should we consider also adding within the singular `RoleBindingKind = "RoleBinding"` as it was present previously in `internal/controller/datadogagent/const.go`, even though it's not used yet ?

tbavelier (approved):
Looking great! One minor comment but LGTM

pkg/kubernetes/rbac/const.go
celenechang: thanks for the review - since we're not using it i'll leave it out, but if ever do need it in the future it'll be clear where it goes

PR #1407, opened 2024-09-11
[cleanup] move uncommon items from api/common/v1

celenechang:
### What does this PR do?

Move items out of common/v1 that are no longer needed there (i.e. they are specific to the v2alpha1 API)
Move a few items in common/const.go to prevent circular imports. The rest will be reviewed/moved in a future PR
Consolidate remaining common/types into a single file
Consolidate `CustomConfig` definitions into a single definition in the v2alpha1 API (and remove conversion function)

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Change should be noop. 
Make sure that the image builds and runs without issue.
Make sure that the Dogstatsd feature works with a CustomConfig.

### Checklist


davidor (approved):
:+1: for files owned by container-platform

api/datadoghq/common/types.go
tbavelier, line 1: `UpdateStrategy` and `RollingUpdate` are only used for profiles/within the Agent controller: should they have their place within this file ? Not sure where they should fit necessarily tho

tbavelier, line 1: `UpdateStrategy` and `RollingUpdate` are only used for profiles/within the Agent controller: should they have their place within this file ? Not sure where they should fit necessarily tho but maybe `api/datadoghq/v2alpha1/datadogagent_types.go` (as `DaemonSetStatus` and `DeploymentStatus` moved from older file) ?

tbavelier (approved):
LGTM, left a minor comment that we can discuss offline
Bonus: `.configData` still works 👍 
```shell
╰─❯ k get dd datadog -ojson | jq .spec.features.dogstatsd
{
  "mapperProfiles": {
    "configData": "- name: 'abc'\n  prefix: 'abc'\n  mappings:\n      - match: 'abc.*'\n        match_type: 'regex'\n        name: 'abc'"
  }
}
╭─ ~                                                                                                                                                                   ⎈ kind-local-k8s/system 15:12:16
╰─❯ k exec -it datadog-agent-tv867 -- bash
Defaulted container "agent" out of: agent, process-agent, init-volume (init), init-config (init)
root@datadog-agent-tv867:/# agent config | grep mapper_profiles -A 10
dogstatsd_mapper_profiles:
- mappings:
  - match: abc.*
    match_type: regex
    name: abc
    tags: {}
  name: abc
  prefix: abc
```

api/datadoghq/common/types.go
celenechang: thanks for the review @tbavelier ! it's certainly a bit gray where they should live, but there is a code refactor for profiles planned so perhaps after that the best place for them will be clearer

PR #1414, opened 2024-09-12
[profiles] Fall back to EDS options if needed

khewonc:
### What does this PR do?

If we don't set a max unavailable in the DAP or the DDA, try to use the [edsMaxPodUnavailable](https://github.com/DataDog/datadog-operator/blob/2a7b83cae17fb1cb6304832f9dde548c9e4fd75e/cmd/main.go#L166) value before falling back to the kubernetes default.

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

* Spin up an operator pod with [DAP enabled](https://github.com/DataDog/datadog-operator/blob/main/docs/datadog_agent_profiles.md#enabling-datadogagentprofiles) and [edsMaxPodUnavailable](https://github.com/DataDog/datadog-operator/blob/2a7b83cae17fb1cb6304832f9dde548c9e4fd75e/cmd/main.go#L166) set to different values
* Create a profile with a max unavailable set to a different value from the DDA and EDS option
```yaml
kind: DatadogAgentProfile
metadata:
  name: dap-test
spec:
  profileAffinity:
    profileNodeAffinity:
      [...]
  config:
    override:
      nodeAgent:
        updateStrategy:
          type: "RollingUpdate"
          rollingUpdate:
            maxUnavailable: 5%
```
* The profile's `CreateStrategy.MaxUnavailable` should be using the calculated value from the profile, i.e. if there are 40 agent pods, 5% of that = `2`
* Remove the profile's `updateStrategy`. The `CreateStrategy.MaxUnavailable` should use the DDA's max unavailable value
* Delete the DDA's `updateStrategy`. The `CreateStrategy.MaxUnavailable` should use the `edsMaxPodUnavailable` value

### Checklist

### DIFF
--- a/internal/controller/datadogagent/component/agent/new.go
+++ b/internal/controller/datadogagent/component/agent/new.go
@@ -8,12 +8,12 @@ package agent
 import (
 	"time"
 
+	apiutils "github.com/DataDog/datadog-operator/api/utils"
 	"github.com/DataDog/datadog-operator/internal/controller/datadogagent/common"
 
 	edsv1alpha1 "github.com/DataDog/extendeddaemonset/api/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 // NewDaemonset use to generate the skeleton of a new daemonset based on few information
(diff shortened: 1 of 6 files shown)


internal/controller/datadogagent/component/agent/new.go
```diff
@@ -117,8 +117,3 @@ func defaultEDSSpec(options *ExtendedDaemonsetOptions) edsv1alpha1.ExtendedDaemo
 	}
 	return spec
 }
-
-func newIntOrStringPointer(str string) *intstr.IntOrString {
```
levan-m, line 121: thanks for refactor

levan-m (approved):

Issue #896, opened 2023-08-18
Loading environment variables using envFrom 


Ekman:
This is potentially a feature request.

Is it possible to use the `envFrom`, or something similar that produces the same results?

```yaml
override:
  nodeAgent:
    envFrom:
      - secretRef:
          name: my-secret-secret
```

khewonc:
Hi @Ekman, we don't currently have an `envFrom` field, but I'll add `envFrom` as a feature request to our backlog. Would using `env.valueFrom` work for your use case for now? https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#define-container-environment-variables-using-secret-data

Ekman:
Thanks for the suggestion! Unfortunately, `env.valueFrom` does not cut it for us as we have too many secrets to load.

khewonc:
I see, thanks for the additional detail. We'll reach out on this thread with any updates on the issue.

jredel:
Hi, Any news on this issue ?

PR #1429, opened 2024-09-23
Load environment variables in override with `envFrom`

swang392:
### What does this PR do?

Add ability for environment variables to be loaded in the override map of datadog-agent.yaml using `envFrom`. 
Example:
Screenshot 2024-09-23 at 4 52 29 PM

### Motivation

#896 

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

* make a kind cluster with Operator
* Make and load some test configMaps and secrets into kind cluster. To double check the priority between env and envFrom (env should take priority), add a field `DD_CLUSTER_NAME: <not your cluster name>`
* Add fields into `datadog-agent.yaml`: spec.override.nodeAgent.EnvFrom, spec.override.clusterAgent.envFrom, and/or spec.override.clusterChecksRunner.envFrom with configMaps or secrets
* add Agent to kind cluster
* run `kubectl describe pod` for agent, cluster checks runner, and cluster agent and verify that the configs or secrets that you specified are there. They should be in a separate section right above the environment variables. 
* run `kubectl exec -it <datadog agent pod> bash` and then `env`. You should be able to see the specific variables you added with their values.
* For agent that you modified DD_CLUSTER_NAME for, when you run the exec command, the `DD_CLUSTER_NAME` field should not be overridden to what was put in the configMap/secret.

### Checklist


fanny-jiang (approved):
🎉

fanny-jiang (approved):

PR #1448, opened 2024-10-03
Fix behavior when overriding names of DCA and CCR components

mrdoggopat:
### What does this PR do?
Fix behavior when overriding names of DCA and CCR components. Previously, the old DCA and CCR components do not get cleaned up when changing the override names. Only the node agent component gets cleaned up.

### Motivation
CECO-1136

### Describe your test plan
Use the following `DatadogAgent` configurations to change the default name of the components:
```
spec:
  override:
    nodeAgent:
      name: foo-agent
    clusterAgent:
      name: foo-dca-agent
    clusterChecksRunner:
      name: foo-ccr-agent
```

After the pods are successfully running, change the config to this:
```
spec:
  override:
    nodeAgent:
      name: foo-agent-test
    clusterAgent:
      name: foo-dca-agent-test
    clusterChecksRunner:
      name: foo-ccr-agent-test
```
Run `kubectl get pods` and you should see that the old pod names for the DCA and CCR are cleaned up and terminated and the new ones should be running with the following names for the node agent, DCA, and CCR respectively:
- foo-agent-test
- foo-dca-agent-test
- foo-ccr-agent-test

Can also remove the above configurations and check if the clean up is happening as expected and the new ones should default to the original names. For example, if you named your DatadogAgent as `datadog`, it should just be:
- datadog-agent
- datadog-cluster-agent
- datadog-cluster-checks-runner

### Checklist


internal/controller/datadogagent/controller_reconcile_ccr.go
```diff
@@ -119,3 +121,40 @@ func deleteStatusWithClusterChecksRunner(newStatus *datadoghqv2alpha1.DatadogAge
...
 }
+
+// cleanupOldCCRDeployments deletes CCR deployments when a CCR Deployment's name is changed using clusterChecksRunner name override
+func (r *Reconciler) cleanupOldCCRDeployments(ctx context.Context, logger logr.Logger, dda *datadoghqv2alpha1.DatadogAgent, newStatus *datadoghqv2alpha1.DatadogAgentStatus) error {
+	matchLabels := client.MatchingLabels{
+		apicommon.AgentDeploymentComponentLabelKey: datadoghqv2alpha1.DefaultClusterChecksRunnerResourceSuffix,
+		kubernetes.AppKubernetesManageByLabelKey:   "datadog-operator",
+	}
+	deploymentName := getDeploymentNameFromCCR(dda)
+	validDeploymentNames := map[string]struct{}{
```
khewonc, line 132: Since there will only be one CCR (and DCA) deployment in a cluster, I think it'll be simpler if we don't create a map for this. We can directly compare the valid deployment name against the `deployment.Name` in line 140

internal/controller/datadogagent/controller_reconcile_ccr_test.go
```diff
@@ -0,0 +1,278 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+	for _, tt := range testCases {
+		t.Run(tt.name, func(t *testing.T) {
+			fakeClient := fake.NewClientBuilder().WithScheme(sch).WithObjects(tt.existingAgents...).Build()
+			logger := logf.Log.WithName("Test_cleanupOldDCADeployments")
```
khewonc, line 245:
```suggestion
			logger := logf.Log.WithName("Test_cleanupOldCCRDeployments")
```

internal/controller/datadogagent/controller_reconcile_dca.go
```diff
@@ -132,3 +134,40 @@ func (r *Reconciler) cleanupV2ClusterAgent(logger logr.Logger, dda *datadoghqv2a
...
 }
+
+// cleanupOldDCADeployments deletes DCA deployments when a DCA Deployment's name is changed using clusterAgent name override
+func (r *Reconciler) cleanupOldDCADeployments(ctx context.Context, logger logr.Logger, dda *datadoghqv2alpha1.DatadogAgent, resourcesManager feature.ResourceManagers, newStatus *datadoghqv2alpha1.DatadogAgentStatus) error {
+	matchLabels := client.MatchingLabels{
+		apicommon.AgentDeploymentComponentLabelKey: datadoghqv2alpha1.DefaultClusterAgentResourceSuffix,
+		kubernetes.AppKubernetesManageByLabelKey:   "datadog-operator",
+	}
+	deploymentName := getDeploymentNameFromDCA(dda)
+	validDeploymentNames := map[string]struct{}{
```
khewonc, line 145: Same as the CCR. We only have one DCA deployment per cluster so we don't need a map for this

internal/controller/datadogagent/controller_reconcile_v2.go
```diff
@@ -200,6 +200,14 @@ func (r *Reconciler) reconcileInstanceV2(ctx context.Context, logger logr.Logger
 		errs = append(errs, err)
 		logger.Error(err, "Error cleaning up old DaemonSets")
 	}
+	if err = r.cleanupOldDCADeployments(ctx, logger, instance, resourceManagers, newStatus); err != nil {
```
khewonc, line 203: I think we can move all 3 of the `cleanup` methods to have one of the these orders so it's easier to read:
* Have each `cleanup` right after each component's `reconcile`
```
reconcileV2ClusterAgent
cleanupOldDCADeployments

reconcileV2Agent
cleanupExtraneousDaemonSets

reconcileV2ClusterChecksRunner
cleanupOldCCRDeployments
```
* Have all the components' `reconcile`, then have all the components' `cleanup`
```
reconcileV2ClusterAgent
reconcileV2Agent
reconcileV2ClusterChecksRunner

cleanupOldDCADeployments
cleanupExtraneousDaemonSets
cleanupOldCCRDeployments
```

khewonc (approved):

Issue #864, opened 2023-07-25
Pod Disruption Budget isn't created for cluster agent


ylee-figma:
**Output of the info page (if this is a bug)**
```
❯ k get datadogagents.datadoghq.com
NAME      AGENT             CLUSTER-AGENT     CLUSTER-CHECKS-RUNNER   AGE
datadog   Running (4/4/4)   Running (2/2/2)                           76d

❯ k get pdb -A | rg cluster

```

**Describe what happened:**
If the replica count is above 1, then PDB should be created according to this logic: https://github.com/DataDog/datadog-operator/blob/2b77d724bd06c256106fe45c54f9a90ddaa2fb8b/controllers/datadogagent/poddisruptionbudget.go#L40

However, maybe because the apiVersion changed on the agent, it's not creating PDBs.

**Describe what you expected:**
PDB for cluster agent to be created

**Steps to reproduce the issue:**
Scale up replica of cluster agent to 2.

**Additional environment details (Operating System, Cloud provider, etc):**

levan-m:
Hello,

Thanks for submitting the issue! You are right `0.8.x` managed PDB for both Cluster Agent and Cluster Checks Runners, this isn't yet supported in `1.x.x`, fact highlighted [here](https://github.com/DataDog/datadog-operator#migrating-from-08x-to-100). We haven't prioritized this for the upcoming `1.1.0`, we may be able to do it for the next one.

jalaziz:
Coming from the helm chart installation method, I noticed this was missing too.

tbavelier:
Completed in https://github.com/DataDog/datadog-operator/pull/1454, will be part of 1.11.x release

PR #1454, opened 2024-10-09
create Pod Disruption Budget for DCA and CCR deployments

swang392:
### What does this PR do?

Create Pod Disruption budgets for cluster agent and cluster checks runners.

### Motivation

#864 We had PDB in version 0.8.x but it was not included in later versions of the operator.

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

- create a kind cluster
- in `datadog-agent.yaml`:
    - enable PDB for DCA and CLC -- override.ClusterAgent.createPodDisruptionBudget = true
    - add replicas:  override.ClusterAgent.replicas = 3
- deploy operator and agent to your kind cluster
- run `kubectl get poddisruptionbudgets` and you should see the DCA and CLC PDBs created. CLC has a maxUnavailable of 1 pod, and DCA has a minAvailable of 1 pod.
- run `kubectl get pods -o wide` to see which node the DCA and CLC pods are on. 
- run `k drain <node> --ignore-daemonsets --delete-emptydir-data`. You should get a message like this: `Cannot evict pod as it would violate the pod's disruption budget.`
- run `kubectl get pods -o wide` to see the status of the nodes now. There should only be one running pod for DCA, and only one pending pod for CLC. 
- run `kubectl uncordone <node>`
- run `kubectl get pods -o wide`-- everything should be back to running.
- remove the flags for PDB and reinstall the agent -- there are no more PDBs created as it is not enabled by default. If you drain a node without PDB, you shouldn't see any error statements and all the pods on that node will have a status of Pending until the node is uncordoned.

### Checklist


estherk15 (approved):

api/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -1423,6 +1423,11 @@ type DatadogAgentComponentOverride struct {
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
 
+	// Set CreatePodDisruptionBudget to true to create a PodDisruptionBudget for this component.
+	// Not applicable for the Node Agent. A Cluster Agent PDB is set with 1 min available pod, and a Cluster Checks Runner PDB is set with 1 max unavailable pod.
```
levan-m, line 1427:
```suggestion
	// Not applicable for the Node Agent. A Cluster Agent PDB is set with 1 minimum available pod, and a Cluster Checks Runner PDB is set with 1 maximum unavailable pod.
```

levan-m (commented):
Left some minor comments/questions. Would be nice if you could add a test [here](https://github.com/DataDog/datadog-operator/blob/main/internal/controller/datadogagent/controller_v2_test.go).
Otherwise looks good to me!

internal/controller/datadogagent/component/clusteragent/utils.go
```diff
@@ -53,6 +58,27 @@ func GetClusterAgentService(dda metav1.Object) *corev1.Service {
 	return service
 }
 
+func GetClusterAgentPodDisruptionBudget(dda metav1.Object) *policyv1.PodDisruptionBudget {
```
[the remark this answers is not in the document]
swang392: I put it in `utils.go` because `GetClusterAgentService` is in `utils.go` and it follows a similar structure as `GetClusterAgentPodDisruptionBudget`. Should I move it to `default.go` for consistency between DCA and CCR?

internal/controller/datadogagent/override/dependencies.go
```diff
@@ -42,11 +44,34 @@ func Dependencies(logger logr.Logger, manager feature.ResourceManagers, dda *v2a
...
+		} else if component == v2alpha1.ClusterChecksRunnerComponentName &&
+			(dda.Spec.Features.ClusterChecks.UseClusterChecksRunners == nil ||
+				*dda.Spec.Features.ClusterChecks.UseClusterChecksRunners) {
+			pdb := componentccr.GetClusterChecksRunnerPodDisruptionBudget(dda)
+			if err := manager.Store().AddOrUpdate(kubernetes.PodDisruptionBudgetsKind, pdb); err != nil {
+				errs = append(errs, err)
+			}
+		}
+	} else if createPdb != nil {
+		logger.Error(nil, "Pod disruption budget is not created by default")
```
levan-m, line 70: This will be logged when `create != nil && *createPdb == false`, why do we want to log error for this on every reconcile? the else block is probably not really necessary.

levan-m (approved):

PR #1379, opened 2024-08-29
expose kubernetes_resources_labels/annotations_as_tags

adel121:
### What does this PR do?

This PR exposes `kubernetesResourcesLabelsAsTags` and `kubernetesResourcesAnnotationsAsTags` in the DD operator.

Feature implemented in [this](https://github.com/DataDog/datadog-agent/pull/27369) PR.

### Motivation

Exposing the feature.

### Additional Notes

We also create DCA rbacs necessary for the feature to work as expected. 

### Minimum Agent Versions

Are there minimum versions of the Datadog Agent and/or Cluster Agent required?

Agent: v7.58.0
Cluster Agent: v7.58.0

### Describe your test plan

Deploy the following DatadogAgent with the operator:

```
apiVersion: datadoghq.com/v2alpha1
kind: DatadogAgent
metadata:
  name: datadog
  namespace: system
spec:
  global:
    credentials:
      apiSecret:
        secretName: datadog-secret
        keyName: api-key
    kubelet:
      tlsVerify: false

    kubernetesResourcesLabelsAsTags:
      pods:
        foo: "bar"
        bar: "baz"
      deployments.apps:
        foo: "bar"
        bar: "baz"

    kubernetesResourcesAnnotationsAsTags:
      pods:
        foo: "bar1"
        bar: "baz"
      deployments.apps:
        foo: "bar"
        bar: "baz"
      daemonsets.apps:
        agent.datadoghq.com/agentspechash: hash
```

Create a deployment with these labels and annotations set on the deployment object and the children pods:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: python-process-deployment
  labels:
    app: python-process-app
    bar: label1
    foo: label2
  annotations:
    bar: label3
    foo: label4
spec:
  replicas: 3
  selector:
    matchLabels:
      app: python-process-app
  template:
    metadata:
      labels:
        app: python-process-app
        bar: label1
        foo: label2
      annotations:
        bar: label3
        foo: label4
    spec:
      containers:
        - name: python-process-container
          image: python:3.7 # Replace with your Python image
          command: ["python3", "-u", "your-python-script.py"]
          volumeMounts:
            - name: python-script-volume
              mountPath: /your-python-script.py
              subPath: your-python-script.py
      volumes:
        - name: python-script-volume
          configMap:
            name: python-script-configmap
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: python-script-configmap
data:
  your-python-script.py: |
    while True:
      pass
```

[...]

PR #1522, opened 2024-11-12
move processing of labels and annotations as tags

celenechang:
### What does this PR do?

- Fix issue where a Cluster Agent ClusterRole and ClusterRolebinding was not being cleaned up when DatadogAgent was deleted. This is because of the order of processing of overrides compared to features.
- Changed behavior to only create ClusterRole if it will not be empty
- Also noticed a small typo in one of the conditionals which is now fixed

### Motivation

The presence of the ClusterRole and ClusterRoleBinding was causing a requeue of a nonexistant object

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Test according to instructions in https://github.com/DataDog/datadog-operator/pull/1379

In addition:
- Verify that deletion of the DatadogAgent resource results in deletion of the ClusterRole and ClusterRoleBinding called `datadog-cluster-agent-annotations-and-labels-as-tags`
- Verify that removal of both `kubernetesResourcesLabelsAsTags` and `kubernetesResourcesAnnotationsAsTags` from the DatadogAgent manifest results in the absence of the ClusterRole `datadog-cluster-agent-annotations-and-labels-as-tags`.

### Checklist


internal/controller/datadogagent/feature/enabledefault/feature.go
```diff
@@ -101,6 +102,10 @@ type secretInfo struct {
 type clusterAgentConfig struct {
 	serviceAccountName        string
 	serviceAccountAnnotations map[string]string
+
+	resourceMetadataAsTagsClusterRoleName string
+	kubernetesResourcesLabelsAsTags       map[string]map[string]string
+	kubernetesResourcesAnnotationsAsTags  map[string]map[string]string
```
adel121, line 108: `kubernetesResourcesLabelsAsTags` and `kubernetesResourcesAnnotationsAsTags` are not specific to cluster agent, it should be available to other components as well (i.e. node agents).

adel121: To elaborate on this:

we previously had, for example, `podLabelsAsTags` which is made available to all components. 

On node agents, this implies that the node agent will use this config to generate tags for pods collected directly from the kubelet.

On CLC runners, this implies that the runner will generate pod tags for unschedulable pods that are collected via informers.

On DCA instance, it means the DCA will collect pods from the API server using workloadmeta and will generate tags based on this config.

So in short, all components need to have access to these 2 configuration values, but only the DCA needs the rbacs

adel121 (commented):
Thanks @celenechang , I've left a comment

.github/workflows/pr-linter.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v1
     - name: Verify Pull Request Labels
-      uses: jesusvasquez333/verify-pr-label-action@v1.3.1
+      uses: jesusvasquez333/verify-pr-label-action@v1.4.0
```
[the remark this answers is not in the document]
celenechang: this change was removed from the PR

adel121:
`kubernetesResourcesLabelsAsTags` and `podLabelsAsTags` are very similar in the sense that they should be applied to all 3 agents (node, cluster and runner), and the cluster agent needs an rbac to watch pods. 

I am curious why these 2 features are implemented differently in the operator? 
If we had an rbac issue with `kubernetesResourcesLabelsAsTags` shouldn't we also have the same issue for `podLabelsAsTags`?

adel121 (commented):
Question: 

`kubernetesResourcesLabelsAsTags` and `podLabelsAsTags` are very similar in the sense that they should be applied to all 3 agents (node, cluster and runner), and the cluster agent needs an rbac to watch pods.

I am curious why these 2 features are implemented differently in the operator?
If we had an rbac issue with `kubernetesResourcesLabelsAsTags` shouldn't we also have the same issue for `podLabelsAsTags`?

adel121 (approved):
LGTM

fanny-jiang (approved):

PR #1547, opened 2024-12-04
Add make command to run a specific e2e test

swang392:
### What does this PR do?

Adds a new make command to run a specific E2E test by setting the `E2E_RUN_REGEX` environment variable in your terminal to the name of the desired test.
Removes parentheses from E2E test names

### Motivation

Sometimes we don't want to run the entire E2E test suite, especially when trying to check or develop one test. 

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

- [ ] set `E2E_RUN_REGEX` in your terminal window to the full name of an E2E test. ex: `export E2E_RUN_REGEX=TestLocalKindSuite/TestGenericK8s/Minimal_DDA_config`
- [ ] run `make e2e-tests`
- [ ] check that only the test you specified has run. You should see a line printing out the value of `E2E_RUN_REGEX` in the output. 
- [ ] remove the variable you set for `E2E_RUN_REGEX` -- `export E2E_RUN_REGEX=`
- [ ] run `make e2e-tests`
- [ ] check that all of the e2e tests run 

### Checklist


Makefile
```diff
@@ -217,6 +217,15 @@ e2e-tests: manifests $(KUSTOMIZE) ## Run E2E tests and destroy environment stack
...
 
+.PHONY: e2e-test-specific 
+e2e-test-specific:
+	@if [ -z "$(E2ETESTNAME)" ]; then \
+		echo "The environment variable E2ETESTNAME is not set. Please set E2ETESTNAME to the name of the test you want to run."; \
+		exit 1; \
+	else \
+		go test ./test/e2e/tests/... -count=1 --tags=e2e -v -run $(E2ETESTNAME); \
+	fi
+
```
fanny-jiang, line 228: Can we put this logic in the `e2e-tests` directive instead and make the env var optional? Also, what do you think of using the env var name, `E2E_RUN_REGEX` or `TEST_RUN_REGEX`?

```suggestion
.PHONY: e2e-tests 
e2e-tests:
	@if [ -n "$(E2E_RUN_REGEX)" ]; then \
                go test ./test/e2e/tests/... -count=1 --tags=e2e -v -run="$(E2E_RUN_REGEX)"; \
	else \
		go test ./test/e2e/tests/... -count=1 --tags=e2e -v -run="TestAWSKindSuite"; \
	fi

```

fanny-jiang (approved):

PR #1560, opened 2024-12-11
[cleanup] move most common constants to v2alpha1

celenechang:
### What does this PR do?

Move nearly all common constants to v2alpha1, because they are no longer common and only used in the v2alpha1 API.

Primary change is moving constants from api/datadoghq/common/const.go -> api/datadoghq/v2alpha1/const.go.

### Motivation

Clean up

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Make sure the binary builds and runs as expected. The change is noop.

### Checklist


sblumenthal (approved):

davidor (approved):
:+1: for container-platform files

PR #1562, opened 2024-12-12
[cleanup] move common envvars to v2alpha1 envvars

celenechang:
### What does this PR do?

Clean up task to move envvars that are no longer common to v2alpha1 .

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Make sure operator binary builds and runs without issue. This change is noop.

### Checklist


api/datadoghq/common/envvar.go
```diff
@@ -5,194 +5,77 @@
...
+	DDAppKey      = "DD_APP_KEY"
+	DDddURL       = "DD_DD_URL"
+	DDURL         = "DD_URL"
+	DDSite        = "DD_SITE"
+	DDLogLevel    = "DD_LOG_LEVEL"
+	DDClusterName = "DD_CLUSTER_NAME"
 
-	EnvVarTrueValue  = "true"
-	EnvVarFalseValue = "false"
+	// move to v2alpha1 envvar.go
```
levan-m, line 20: is the comment intentional?

api/datadoghq/common/envvar.go
```diff
@@ -5,194 +5,77 @@
...
-
-	DDOTLPgRPCEndpoint = "DD_OTLP_CONFIG_RECEIVER_PROTOCOLS_GRPC_ENDPOINT"
-	DDOTLPHTTPEndpoint = "DD_OTLP_CONFIG_RECEIVER_PROTOCOLS_HTTP_ENDPOINT"
-
-	// KubernetesEnvvarName Env var used by the Datadog Agent container entrypoint
-	// to add kubelet config provider and listener
-	KubernetesEnvVar = "KUBERNETES"
 
-	ClusterChecksConfigProvider = "clusterchecks"
+	// keep in common? will this eventually be needed for dap?
```
levan-m, line 11: it's fine to move these too imo. if we end up separating crd and controller modules, probably we will revisit what stays in the module anyway.

levan-m (approved):
looks good, left non-blocking comments.

sblumenthal (approved):

api/datadoghq/v2alpha1/envvar.go
clamoriniere, line 1: IMO, they should not be part of the `api/` packages because they are specific to the agent config.

Have you evaluated to move this list to a dedicate package in "internal/controller/*"? it seems that the envvars are only used in the controller logic.  Do we know if another project is using them?

api/datadoghq/common/envvar.go
celenechang: 🤦‍♀️ thanks

api/datadoghq/v2alpha1/envvar.go
clamoriniere: > agent-specific constants defined in api/.../const.go . Should we move all those out too?

I think so, It will be beneficial with the on-going effort to create a dedicated go.mod for the `api/` packages. But IMO it can be done in a futur PR.

celenechang: Thanks, Levan and I spoke about it and we decided the same - will move in a future PR.

adel121 (approved):

PR #1576, opened 2024-12-17
[cleanup] Remove operator reference in /api folder, refactor v2alpha files

swang392:
### What does this PR do?

Moved:
- some consts and a `v2alpha1/utils` from `api/datadoghq/v2alpha1` to a new directory `pkg/constants/`.
- `v2alpha1/test` to `pkg/testutils`
- `v2alpha1/datadogagent_default` to a new folder, `internal/controller/defaults`

### Motivation

We want to create a new go module for `/api`, so files that are not specific to the CRDs should be moved out of the `/api` folder. There are some files in `/api` that depended on `datadog-operator/pkg/defaulting` -- this dependency should be removed because a project depending on the `/api` module would indirectly depend on `datadog-operator` too. 

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Make sure the binary builds and runs as expected. The change is noop.

### Checklist


levan-m:
See this discussion here https://github.com/DataDog/datadog-operator/pull/1562#discussion_r1883439244. 
Given there are  api->controller dependencies not considered earlier, instead of eliminating the dependencies we should just move _everything_ not related to CRDs to the main module now - things like env vars, constants, some utilities (secrets.go, condition.go, utils.go, etc). 
cc @celenechang

estherk15 (approved):

levan-m (approved):

levan-m:
this is a non-functional change and can be merged without approvals from other teams.

PR #1608, opened 2025-01-07
[profiles] Do not error if pod needing deletion is not found

khewonc:
### What does this PR do?

Occasionally, when we need to delete existing agent pods that don't match a node's affinity, there's a chance that some pods are already deleted between the time the cached list of pods the operator queries is updated and the operator sends the deletion request. When that happens, the operator receives a pod not found error from the deletion request. Instead of causing a reconciler error, this PR ignores the not found error.

### Motivation

CECO-1383

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

This PR is difficult to test since the issue only happens sporadically. It was reported when a new DatadogAgentProfile was created that triggered the operator to delete existing agent pods for not meeting the updated nodeaffinity requirements and affected approximately 1 out of 8 agent pods in that instance.

One way to test could be to create a DatadogAgentProfile in staging and ensuring you don't see a reconcile error log similar to the following during rollout of the profile:

```
{"level":"ERROR","ts":"2024-08-08T19:46:12.565Z","msg":"Reconciler error","controller":"datadogagent","controllerGroup":"datadoghq.com","controllerKind":"DatadogAgent","DatadogAgent":{"name":"datadog-agent","namespace":"datadog-agent"},"namespace":"datadog-agent","name":"datadog-agent","reconcileID":"xxx","error":"pods \"datadog-agent-xxxxx-xxxxx\" not found"}
```

### Checklist

### DIFF
--- a/internal/controller/datadogagent/controller_reconcile_agent_test.go
+++ b/internal/controller/datadogagent/controller_reconcile_agent_test.go
@@ -1841,3 +1841,226 @@ func Test_labelNodesWithProfiles(t *testing.T) {
 		})
 	}
 }
+
+func Test_cleanupPodsForProfilesThatNoLongerApply(t *testing.T) {
+	sch := runtime.NewScheme()
+	_ = scheme.AddToScheme(sch)
+	ctx := context.Background()
+
+	testCases := []struct {
+		name           string
+		description    string
+		profilesByNode map[string]types.NamespacedName
+		ddaNamespace   string
+		existingPods   []client.Object
+		wantPods       []corev1.Pod
+	}{
+		{
+			name:        "delete agent pod that shouldn't be running",
+			description: "pod-2 should be deleted",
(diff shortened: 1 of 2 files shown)


internal/controller/datadogagent/controller_reconcile_agent_test.go
```diff
@@ -1841,3 +1841,226 @@ func Test_labelNodesWithProfiles(t *testing.T) {
...
+					},
+					Spec: corev1.PodSpec{
+						NodeName: "node-1",
+					},
+				},
+			},
+		},
+		{
+			name:        "delete default agent on profile node",
+			description: "pod-2 should be deleted",
```
celenechang, line 1923: nit: should this be `pod-default`?

celenechang (approved):

PR #1618, opened 2025-01-10
Move condition logic to main module

davidor:
### What does this PR do?

Moves the logic in `api/datadoghq/v2alpha1/condition.go` to the main module.

That file depends on EDS and we'd rather avoid having that dependency in the api module. Otherwise, projects importing it will indirectly depend on EDS too. It makes sense to move this logic to the main module because it doesn't have any CRD definitions, which is the point of the api module.

### Describe your test plan

It's a refactor. Tests should be enough.

### Checklist


pkg/controller/utils/datadog/condition.go
```diff
@@ -31,28 +33,19 @@ const (
 	DatadogAgentStateFailed DatadogAgentState = "Failed"
 )
 
-// UpdateDatadogAgentStatusConditionsFailure used to update the failure StatusConditions
```
davidor, line 34: For reviewers: I deleted this function and others that were not used.

go.mod
```diff
@@ -43,6 +43,7 @@ require (
 	github.com/DataDog/datadog-agent/pkg/config/model v0.59.0-rc.5
 	github.com/DataDog/datadog-agent/pkg/config/remote v0.59.0-rc.5
 	github.com/DataDog/datadog-agent/pkg/remoteconfig/state v0.59.0-rc.5
+	github.com/DataDog/datadog-operator/api v0.0.0-20250113195927-3d36de7daa54
```
davidor, line 46: I don't think this is correct. This dependency should by managed by `go work` without pinning to a specific version.

davidor: I added a new commit with the fix.

levan-m: Thanks for catching this!

levan-m (approved):

PR #1650, opened 2025-02-03
Add PodResources mount to agent

gjulianm:
### What does this PR do?

This PR adds the mount for the PodResources socket.

### Motivation

The pod resources API is necessary for correctly assigning pods to GPU devices in GPU monitoring.

### Additional Notes

### Minimum Agent Versions

* Agent: v7.62

### Describe your test plan

Tested locally by deploying the operator in a cluster, deploying an agent with default settings and ensuring that:
- The agent container has a mount for `/var/lib/kubelet/pod-resources/kubelet.sock` 
- The environment variable `DD_KUBERNETES_KUBELET_POD_RESOURCES_SOCKET` is correctly set to `/var/lib/kubelet/pod-resources/kubelet.sock`.

### Checklist

### DIFF
--- a/internal/controller/datadogagent/override/global_test.go
+++ b/internal/controller/datadogagent/override/global_test.go
@@ -7,6 +7,7 @@ package override
 
 import (
 	"fmt"
+	"slices"
 	"testing"
 
 	"github.com/DataDog/datadog-operator/pkg/constants"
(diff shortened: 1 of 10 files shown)


estherk15 (approved):

internal/controller/datadogagent/override/global_test.go
```diff
@@ -32,6 +33,7 @@ import (
 const (
 	hostCAPath           = "/host/ca/path/ca.crt"
 	agentCAPath          = "/agent/ca/path/ca.crt"
+	podResourcesSocket   = v2alpha1.DefaultKubeletPodResourcesSocket
```
celenechang, line 36: do you mind defining the constant here instead of in api/v2alpha1, as the others? i know it's not ideal to define it in more than one place, but we're actively trying to trim down the constants defined in the api/ folder

gjulianm: Removed from `api/v2alpha1` and replicated here

celenechang (approved):

PR #1664, opened 2025-02-06
Increase prometheus receiver scrape interval

mackjmr:
### What does this PR do?

Current interval leads to too much metric volume
Same as: DataDog/datadog-agent#33714 but for operator defaults.

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Run agent with otelCollector feature enabled (`features.otelCollector.enabled`) without passing a collector config (it will then use default we are changing here), and check that the `otel-config.yaml` mounted on the otel-agent container matches the one from this PR.

### Checklist


celenechang (approved):
@mackjmr If we want to verify this change, is the best way to check the `otel-config.yaml` mounted on the agent container? Thank you.

celenechang (changes requested):
Apologies for the hasty approval; can you check the CI failures?

mackjmr:
> If we want to verify this change, is the best way to check the otel-config.yaml mounted on the agent container?

You need to check the one mounted inside the `otel-agent` container.

tbavelier (approved):

levan-m (approved):

PR #1667, opened 2025-02-06
[profiles] Wait to update profile DS when canary present

khewonc:
### What does this PR do?

When using EDS and profiles and a canary, the profile DaemonSet is immediately updated while the EDS goes through a canary phase. This PR prevents profile DSs from being updated until the canary is successful.

### Motivation

https://datadoghq.atlassian.net/browse/CECO-1947

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Example setup:
* kind cluster with at least 2 nodes, e.g. 1 control plane + 1 worker, 1 control plane + multiple workers
* EDS controller setup: `helm repo update && helm install eds datadog/extendeddaemonset`
  * Double check that the `ExtendedDaemonSet` and `ExtendedDaemonSetReplicaSet` CRDs are present in the cluster
  * Install the EDS plugin, follow the [EDS readme instructions](https://github.com/DataDog/extendeddaemonset?tab=readme-ov-file#kubectl-plugin). `make kubectl-eds` will install the plugin in the `bin` folder in the extendeddaemonset repo so you can run commands using `./bin/kubectl-eds canary pause <eds-name>`
* Operator setup:
  * Double check the operator can get, list, and watch ExtendedDaemonSetReplicaSets:
```yaml
- apiGroups:
  - datadoghq.com
  resources:
  - extendeddaemonsetreplicasets
  verbs:
  - get
  - list
  - watch
```
  * Enable/disable profiles using [datadogAgentProfile.enabled](https://github.com/DataDog/helm-charts/blob/c65f22450ed666129286c2123d0de864d3b26a90/charts/datadog-operator/values.yaml#L70-L72) or [datadogAgentProfileEnabled](https://github.com/DataDog/datadog-operator/blob/63b056f22e5e283381172616508238879ac83ac0/cmd/main.go#L161)
  * Enable/disable EDS support using [supportExtendedDaemonset](https://github.com/DataDog/helm-charts/blob/c65f22450ed666129286c2123d0de864d3b26a90/charts/datadog-operator/values.yaml#L73-L74) or [supportExtendedDaemonset](https://github.com/DataDog/datadog-operator/blob/63b056f22e5e283381172616508238879ac83ac0/cmd/main.go#L167)
* Example DDA
```yaml
kind: DatadogAgent
metadata:
  name: datadog-agent
spec:
  global:
    kubelet:
      tlsVerify: false
    credentials:
      <credentials>
  override:
    nodeAgent:
      tolerations:
        # optional: allow node agents to be scheduled on the control plane node
        - key: node-role.kubernetes.io/control-plane
          operator: Exists
          effect: NoSchedule
      containers:
        agent:
          env:
            - name: FOO
              value: "foo"

```

Tests:
Make sure changes to the DDA are immediately reflected in the DS/EDS as normal when:
  * EDS is not enabled but profiles are enabled
  * Profiles are not enabled but EDS is enabled

With EDS and profiles enabled and an active profile DS applied:
Check that profile DSs aren't updated until after a canary is finished:
  * Change the DDA's spec so that a canary is created (`kubectl get ers`). Ensure that the profile DS does not update to the new spec while the canary is active (`kubectl describe ds <profile-ds>`). In the operator logs, you should see the log `Waiting to update profile DaemonSet because of an active canary`
  * While the canary is active, pause it using the kubectl-eds plugin (`./bin/kubectl-eds canary pause <eds>`) and check the operator logs for the line `Waiting to update profile DaemonSet because the canary was paused`
  * Unpause the canary using `./bin/kubectl-eds canary unpause <eds>` and check that the profile DS is still not updated
  * After the canary duration (default 10min) is over and the canary is successful, check that the profile DS is updated
With EDS and profiles

Check that 1) profile DSs are updated when the EDS is manually validated, 2) subsequent DDA updates don't trigger immediate profile DS updates, and 3) subsequent canary failures don't update the profile DS:
* Change the DDA's spec so that a canary is created. The profile DS should not update yet
* Manually validate the EDS using `./bin/kubectl-eds canary validate <eds>`
* The profile DS should update soon with the operator log `Updating profile DaemonSet because the canary was validated`
* Change the DDA spec again and ensure that the profile DS does not update the new DDA spec immediately. The log `Waiting to update profile DaemonSet after DatadogAgent update` should be present in the operator
* After the previous canary is finished, change the DDA's spec again. The profile DS should not update yet
* Fail the canary: `./bin/kubectl-eds canary fail <eds>`
* Ensure the profile DS does not update even after the canary duration (default 10min)

### Checklist

### DIFF
--- a/internal/controller/datadogagent/controller_reconcile_v2_common.go
+++ b/internal/controller/datadogagent/controller_reconcile_v2_common.go
@@ -493,8 +493,9 @@ func (r *Reconciler) shouldUpdateProfileDaemonSet(profile *v1alpha1.DatadogAgent
 		}
 
 		// eds canary was validated manually
-		if eds.Annotations[edsv1alpha1.ExtendedDaemonSetCanaryValidAnnotationKey] == ers.Name {
-			r.log.Info("Updating profile DaemonSet because the canary was validated")
+		if eds.Annotations[edsv1alpha1.ExtendedDaemonSetCanaryValidAnnotationKey] == ers.Name &&
+			eds.Annotations[datadoghqv2alpha1.MD5AgentDeploymentAnnotationKey] == ers.Annotations[datadoghqv2alpha1.MD5AgentDeploymentAnnotationKey] {
+			r.log.Info("Updating profile DaemonSet because the canary was validated and EDS and ERS hashes match")
 			return true, nil
 		}
 
@@ -506,6 +507,7 @@ func (r *Reconciler) shouldUpdateProfileDaemonSet(profile *v1alpha1.DatadogAgent
 
 		// if eds and ers agentspechash match, canary was successful
 		if eds.Annotations[datadoghqv2alpha1.MD5AgentDeploymentAnnotationKey] == ers.Annotations[datadoghqv2alpha1.MD5AgentDeploymentAnnotationKey] {
+			r.log.Info("Updating profile DaemonSet because the EDS and ERS hashes match")
 			return true, nil
 		}
 	}
--- a/internal/controller/datadogagent/controller_reconcile_v2_common.go
+++ b/internal/controller/datadogagent/controller_reconcile_v2_common.go
@@ -7,13 +7,16 @@ package datadogagent
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
+	apicommon "github.com/DataDog/datadog-operator/api/datadoghq/common"
 	"github.com/DataDog/datadog-operator/api/datadoghq/v1alpha1"
 	datadoghqv2alpha1 "github.com/DataDog/datadog-operator/api/datadoghq/v2alpha1"
 	"github.com/DataDog/datadog-operator/pkg/agentprofile"
 	"github.com/DataDog/datadog-operator/pkg/condition"
+	"github.com/DataDog/datadog-operator/pkg/constants"
 	"github.com/DataDog/datadog-operator/pkg/controller/utils/comparison"
 	"github.com/DataDog/datadog-operator/pkg/controller/utils/datadog"
 	"github.com/DataDog/datadog-operator/pkg/kubernetes"
@@ -23,6 +26,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
@@ -156,7 +160,7 @@ func (r *Reconciler) createOrUpdateDaemonset(parentLogger logr.Logger, dda *data
 	}
 
 	if alreadyExists {
-		now := metav1.NewTime(time.Now())
+		now := metav1.Now()
 		if agentprofile.CreateStrategyEnabled() {
 			if profile.Status.CreateStrategy != nil {
 				profile.Status.CreateStrategy.PodsReady = currentDaemonset.Status.NumberReady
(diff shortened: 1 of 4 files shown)


internal/controller/datadogagent/controller_reconcile_v2_common_test.go
```diff
@@ -215,9 +226,589 @@ func Test_shouldCheckCreateStrategyStatus(t *testing.T) {
...
+			existingObjects: []client.Object{
+				&edsdatadoghqv1alpha1.ExtendedDaemonSet{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "test-eds",
+						Namespace: testNS,
+						Labels:    testEDSLabels,
+					},
+					Status: edsdatadoghqv1alpha1.ExtendedDaemonSetStatus{
+						Canary: &edsdatadoghqv1alpha1.ExtendedDaemonSetStatusCanary{},
+					},
```
tbavelier, line 394:
```suggestion
					Status: edsdatadoghqv1alpha1.ExtendedDaemonSetStatus{
						Canary: &edsdatadoghqv1alpha1.ExtendedDaemonSetStatusCanary{
							ReplicaSet: "test-ers-2",
							Nodes:      []string{"node-foo"},
						},
					},
```
nit: Not entirely sure about the indent with Github UI, but suggestion is to have a "real" canary status object, not just for the sake of the test passing since it's not nil.

internal/controller/datadogagent/controller_reconcile_v2_common_test.go
```diff
@@ -215,9 +226,589 @@ func Test_shouldCheckCreateStrategyStatus(t *testing.T) {
...
+						Namespace: testNS,
+						Labels:    testERSLabels,
+					},
+				},
+			},
+			expectedShouldUpdate: false,
+			errorMessage:         nil,
+		},
+		{
+			name: "Canary validated annotation present but name doesn't match",
```
tbavelier, line 527: I don't understand this specific test, what does it validate ? I understand we should update instantly if the valid ERS is `test-ers-2` while the active ERS is `test-ers-1`, but I don't understand the scenario under which it happens: a manual canary validation ?

tbavelier (approved):
left non blocking comments, just nits and for my own understanding

internal/controller/datadogagent/controller_reconcile_v2_common.go
```diff
@@ -407,3 +419,112 @@ func shouldCheckCreateStrategyStatus(profile *v1alpha1.DatadogAgentProfile) bool
...
+
+	// profile should not be nil or the default profile
+	if profile == nil {
+		return true, nil
+	}
+	if agentprofile.IsDefaultProfile(profile.Namespace, profile.Name) {
+		return true, nil
+	}
+
+	// TODO: check that EDS is for a specific DDA
```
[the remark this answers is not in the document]
khewonc: It's not necessary at the moment to check that an EDS matches a specific DDA since we don't officially support multiple DDAs right now

internal/controller/datadogagent/controller_reconcile_v2_common.go
```diff
@@ -407,3 +419,112 @@ func shouldCheckCreateStrategyStatus(profile *v1alpha1.DatadogAgentProfile) bool
...
+			return false, errors.New("there must exist at least 1 ExtendedDaemonSetReplicaSet")
+		}
+		// wait for canary ers to be cleaned up if there are multiple ers
+		if len(ersList.Items) > 1 {
+			r.log.Info("Waiting to update profile DaemonSet until unused ExtendedDaemonSetReplicaSets are cleaned up")
+			return false, nil
+		}
+		ers := ersList.Items[0]
+		// the eds's active ers should match the ers name
+		if eds.Status.ActiveReplicaSet != ers.Name {
```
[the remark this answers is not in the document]
khewonc: I didn't have a specific example in mind that would cause this. It's more of an extra check since the names should match

internal/controller/datadogagent/controller_reconcile_v2_common_test.go
khewonc: 🤔 I might have mixed myself up when I made some edits. I'll switch it to the validation annotation being present but the agent spec hashes not matching since that can happen if a canary is manually validated and then the DDA is updated again but the canary fails

PR #1672, opened 2025-02-08
Include golangci-lint in fmt Make target

shuheiktgw:
### What does this PR do?

Currently, there's no easy way to run `golangci-lint run ./... --fix` for formatting. It would be more convenient and intuitive if we could execute it via the fmt Make target.

Thanks for your review!

### Motivation

Running `golangci-lint run ./... --fix` easier.

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,8 +68,9 @@ all: build test ## Build test
 build: manager kubectl-datadog ## Builds manager + kubectl plugin
 
 .PHONY: fmt
-fmt: ## Run go fmt against code
+fmt: bin/$(PLATFORM)/golangci-lint ## Run formatters against code
 	go fmt ./...
+	bin/$(PLATFORM)/golangci-lint run ./... --fix
 
 .PHONY: vet
 vet: ## Run go vet against code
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -306,7 +307,7 @@ patch-crds: bin/$(PLATFORM)/yq ## Patch-crds
 	hack/patch-crds.sh
 
 .PHONY: lint
-lint: bin/$(PLATFORM)/golangci-lint fmt vet ## Lint
```
shuheiktgw, line 309: I believe we should not include the fmt target in the lint target. Consider the following two cases:

1. A user makes local commits and runs the lint target for a final check. Even if there are lint failures, the fmt target fixes them automatically, causing the lint target to succeed despite local diffs. As a result, the user may push the commit without realizing there are uncommitted changes.
2. In CI, make lint runs, and even if there are lint failures, the fmt target fixes them. This could cause the CI job to pass mistakenly, despite the issues.

shuheiktgw:
https://github.com/DataDog/datadog-operator/pull/1714

PR #1681, opened 2025-02-10
[cleanup] move env vars out of /api folder

swang392:
### What does this PR do?

The environment variables in v2alpha1 aren't used in the CRDS. Moved the variables defined in /api to: 
- `internal/controller/datadogagent/common/envvar.go` if they are used in the controller and in multiple features
- a specific `envvar.go` file in a feature if it's used exclusively in one feature

### Motivation

We don't need to have the constants in /api if they aren't relevant to the CRDs. Moving the constants closer to where they are used makes the code easier to maintain.

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

should have no effect, everything builds and runs as normal

### Checklist


api/datadoghq/v2alpha1/envvar.go
tbavelier, line 1: I'd be in favor of fully removing this file now that it's empty

pkg/constants/envvar.go
```diff
@@ -0,0 +1,16 @@
...
+
+package constants
+
+const (
+	DDAPIKey      = "DD_API_KEY"
+	DDAppKey      = "DD_APP_KEY"
+	DDddURL       = "DD_DD_URL"
+	DDURL         = "DD_URL"
+	DDSite        = "DD_SITE"
+	DDLogLevel    = "DD_LOG_LEVEL"
```
tbavelier, line 14: nit: Could possibly move `DDLogLevel` over to `internal/controller/datadogagent/override/envvar.go` as it's only used in `internal/controller/datadogagent/override/` multiple times **and once** in `internal/controller/testutils` where it could be simply replaced as a string `DD_LOG_LEVEL`

tbavelier (approved):
Thanks again for the effort!

PR #1686, opened 2025-02-13
Use validation rules to validate DatadogMonitorSpec

shuheiktgw:
### What does this PR do?

This PR implements validation rules for validation rules for DatadogMonitorSpec. Thanks for your review!

### Motivation

Currently, the Datadog Operator does not implement validating webhooks and validates inputs during reconciliation. I assume this is to simplify the operator setup, but from a user experience perspective, it’s not ideal since users can only discover issues after applying their manifests.  

Kubernetes now offers a feature called [validation rules](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules), which allows input validation using CEL without requiring webhooks. I believe this would enhance the user experience by enabling users to identify errors before creating the resource.

### Additional Notes

- Kubernetes 1.23 or earlier does not seem to support the `x-kubernetes-validations` field in CRDs. However, I assume this isn’t a major issue since the operator uses Kubernetes libraries around v0.31, and no one should expect it to work on a Kubernetes 1.23 cluster anyway.  
- This PR is primarily intended to propose and demonstrate the use of validation rules. However, if this approach looks good, I’d be happy to implement validation rules for other resources as well.

### Describe your test plan

I've tested it in my local kind cluster v1.32.

```
$ cat monitor.yaml
kind: DatadogMonitor
metadata:
  name: datadog-monitor-test
spec:
  # query: "avg(last_10m):avg:system.disk.in_use{*} by {host} > 0.5"
  # type: "metric alert"
  name: "Test monitor made from DatadogMonitor"
  message: "We are running out of disk space!"
  tags:
    - "test:datadog"
```

```
$ k apply -f monitor.yaml 
The DatadogMonitor "datadog-monitor-test" is invalid:
* spec: Invalid value: "object": spec.Query must be defined
* spec: Invalid value: "object": spec.Type must be defined
```

### Checklist

### DIFF
--- a/api/datadoghq/v1alpha1/datadogmonitor_types.go
+++ b/api/datadoghq/v1alpha1/datadogmonitor_types.go
@@ -13,6 +13,10 @@ import (
 
 // DatadogMonitorSpec defines the desired state of DatadogMonitor
 // +k8s:openapi-gen=true
+// +kubebuilder:validation:XValidation:rule="has(self.name) && size(self.name) > 0",message="spec.Name must be defined"
+// +kubebuilder:validation:XValidation:rule="has(self.message) && size(self.message) > 0",message="spec.Message must be defined"
+// +kubebuilder:validation:XValidation:rule="has(self.query) && size(self.query) > 0",message="spec.Query must be defined"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && size(self.type) > 0",message="spec.Type must be defined"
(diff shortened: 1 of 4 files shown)


shuheiktgw:
Recreated https://github.com/DataDog/datadog-operator/pull/1692 in upstream.

PR #1690, opened 2025-02-13
Bump go dependencies

fanny-jiang:
### What does this PR do?

* Bump github.com/Masterminds/semver/v3 from 3.1.1 to 3.3.1 https://github.com/DataDog/datadog-operator/pull/1572
* Bump k8s.io/apiextensions-apiserver from 0.31.1 to 0.32.1 https://github.com/DataDog/datadog-operator/pull/1636
* Bump github.com/onsi/gomega from 1.33.1 to 1.36.2 https://github.com/DataDog/datadog-operator/pull/1595
* Bump sigs.k8s.io/controller-runtime from 0.19.0 to 0.20.1 https://github.com/DataDog/datadog-operator/pull/1642

### Motivation

Dependabot

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist


docs/configuration.v2alpha1.md
```diff
@@ -419,6 +419,7 @@ In the table, `spec.override.nodeAgent.image.name` and `spec.override.nodeAgent.
...
...ivileged and unprivileged Pods sharing the same volume on the same node.  "MountOption" mounts all eligible Pod volumes with `-o context` mount option. This requires all Pods that share the same volume to use the same SELinux label. It is not possible to share the same volume among privileged and unprivileged Pods. Eligible volumes are in-tree FibreChannel and iSCSI volumes, and all CSI volumes whose CSI driver announces SELinux support by setting spec.seLinuxMount: true in their CSIDriver instance. Other volumes are always re-labelled recursively. "MountOption" value is allowed only when SELinuxMount feature gate is enabled.  If not specified and SELinuxMount feature gate is enabled, "MountOption" is used. If not specified and SELinuxMount feature gate is disabled, "MountOption" is used for ReadWriteOncePod volumes and "Recursive" for all other volumes.  This field affects only Pods that have SELinux label set, either in PodSecurityContext or in SecurityContext of all containers.  All Pods that use the same volume should use the same seLinuxChangePolicy, otherwise some pods can get stuck in ContainerCreating state. Note that this field cannot be set when spec.os.name is windows. |
```
rtrieu, line 422:
```suggestion
| [key].securityContext.seLinuxChangePolicy | seLinuxChangePolicy defines how the container's SELinux label is applied to all volumes used by the Pod. It has no effect on nodes or volumes that do not support SELinux. Valid values are "MountOption" and "Recursive". Here, "Recursive" means the relabeling of all files on all Pod volumes by the container runtime. This may be slow for large volumes, but it allows the mixing of privileged and unprivileged Pods sharing the same volume on the same node.  "MountOption" mounts all eligible Pod volumes with `-o context` mount option. This requires all Pods that share the same volume to use the same SELinux label. It is not possible to share the same volume among privileged and unprivileged Pods. Eligible volumes are in-tree FibreChannel and iSCSI volumes, and all CSI volumes whose CSI driver announces SELinux support by setting spec.seLinuxMount: true in their CSIDriver instance. Other volumes are always re-labelled recursively. The "MountOption" value is allowed only when SELinuxMount feature gate is enabled.  If not specified and SELinuxMount feature gate is enabled, "MountOption" is used. If not specified and SELinuxMount feature gate is disabled, "MountOption" is used for ReadWriteOncePod volumes and "Recursive" for all other volumes.  This field affects only Pods that have SELinux label set, either in PodSecurityContext or in SecurityContext of all containers.  All Pods that use the same volume should use the same seLinuxChangePolicy, otherwise some pods can get stuck in the ContainerCreating state. Note that this field cannot be set when spec.os.name is windows. |
```

rtrieu (changes requested):
Some minor feedback for your consideration.

tbavelier:
Done in https://github.com/DataDog/datadog-operator/pull/1924

PR #1712, opened 2025-02-20
Remove deprecated dependencies

shuheiktgw:
### What does this PR do?

I've removed deprecated libraries and added a linter to prevent them from being imported again.

### Motivation

These two libraries are deprecated and we should migrate them to new ones.
- math/rand
- github.com/pkg/errors

### Additional Notes

N/A

### Minimum Agent Versions

N/A

### Describe your test plan

N/A

### Checklist


.golangci.toml
```diff
@@ -33,6 +33,13 @@
       ' \+',
     ]
 
+  [[linters-settings.depguard.rules.main.deny]]
```
shuheiktgw, line 36: I found an issue with the `make lint` command—golangci-lint doesn’t actually run in `api/` and `e2e/` because it doesn’t recognize `go.work`. I'll address this in a separate PR.

swang392 (approved):

PR #1714, opened 2025-02-21
Include golangci-lint in fmt Make target

shuheiktgw:
### What does this PR do?

I've recreated https://github.com/DataDog/datadog-operator/pull/1672 in upstream.

### Motivation

See https://github.com/DataDog/datadog-operator/pull/1672.

### Additional Notes

N/A

### Minimum Agent Versions

N/A

### Describe your test plan

N/A

### Checklist


Makefile
```diff
@@ -306,7 +307,7 @@ patch-crds: bin/$(PLATFORM)/yq ## Patch-crds
 	hack/patch-crds.sh
 
 .PHONY: lint
-lint: bin/$(PLATFORM)/golangci-lint fmt vet ## Lint
```
shuheiktgw, line 309: I believe we should not include the fmt target in the lint target. Consider the following two cases:

1. A user makes local commits and runs the lint target for a final check. Even if there are lint failures, the fmt target fixes them automatically, causing the lint target to succeed despite local diffs. As a result, the user may push the commit without realizing there are uncommitted changes.
2. In CI, make lint runs, and even if there are lint failures, the fmt target fixes them. This could cause the CI job to pass mistakenly, despite the issues.

Makefile
```diff
@@ -190,7 +191,7 @@ docker-push-check-img:
 ##@ Test
 
 .PHONY: test
-test: build manifests generate fmt vet verify-licenses gotest integration-tests integration-tests-v2 ## Run unit tests and integration tests
```
shuheiktgw, line 193: I believe `lint` is preferable to `fmt` since developers usually expect tests to have no side effects. Please note that the `lint` target also executes `vet`.

swang392 (approved):

PR #1730, opened 2025-02-25
enhancement: per-container image override support

tobz:
### What does this PR do?

Adds support to override the image used on a per-container basis.

A new "experimental" image override annotation can be used to pass JSON-based configuration that allows setting an image name and/or tag that overrides the container's image. This override logic is applied _after_ the component-level overrides, allowing for hierarchical override support.

The annotation (`experimental.agent.datadoghq.com/image-override-config`) takes a JSON-based map of container name to override image name and/or tag, similar to the component-level image override structure. The image name and tag values are also processed with the same logic as the component-level image overrides, such that the resulting container image is identical to how it would be overridden when using component-level image overrides.

Examples of the annotation's value are shown below in the test plan section.

### Motivation

Per-container image override support forms part of the long-term strategy around shrinking down the Agent container size by allowing us to build dedicated container images for each subagent. This work also supports the continued deployment of ADP in our internal environments by allowing us to decouple from the full Agent image, in order to allow normal testing of custom Agent versions, and Agent RC versions, to be deployed independently of ADP.

### Additional Notes

I've put all of this code into its own section that is meant to be a holding area for all experimental-y things like this, since I can imagine we may want to introduce more features in this way in the future. I didn't go so far as adding a bunch of interfaces or boilerplate or anything, but just wanted to sort of give it its own place to live.

Additionally, I've only applied this logic to the Agent reconciler, and not the DCA or CCR reconcilers. We _could_ apply it to those -- their container names are distinct so I don't think there would be any issue in terms of the override logic -- but I just wanted to keep things more narrowly scoped for now.

### Minimum Agent Versions

* Agent: N/A
* Cluster Agent: N/A

### Describe your test plan

I've added a number of unit tests to minimally verify that the image override logic updates the full image path as intended, but this was primarily tested in a local Kind cluster.

1. Following the contributing guide, creating a Kind cluster and deploying an Operator build based on this branch.
2. Deploy a minimal Datadog Agent configuration, which should result in an Agent pod with the Core Agent (`agent`) and Trace Agent (`trace-agent`) being deployed with version `7.63.2`.
3. Update the configuration to add the new experimental annotation, which we'll use to override the Trace Agent image:

```yaml
metadata:
  annotations:
    experimental.agent.datadoghq.com/image-override-config: |
      {
        "trace-agent": {
          "tag": "7.63.1-rc.1"
        }
      }
spec:
  global:
    env:
     # We use this environment variable to ensure a change is driven to update
     # the Agent deployment, as annotation changes alone won't do it: 
      - name: REVISION
        value: "1"
```

4. Deploy the updated manifest and verify that the Core Agent is still at version `7.63.2` while the Trace Agent is now at version `7.63.1-rc.1` based on its per-container image override.

5. Update the `DatadogAgent` resource to override the image tag on the node agent level, and also update our annotation to ensure they reference different versions:

```yaml
metadata:
  annotations:
    experimental.agent.datadoghq.com/image-override-config: |
      {
        "trace-agent": {
          "tag": "7.63.1-rc.2"
        }
      }
spec:
  global:
    env:
     # We use this environment variable to ensure a change is driven to update
     # the Agent deployment, as annotation changes alone won't do it: 
      - name: REVISION
        value: "2"
  override:
    nodeAgent:
      image:
        tag: 7.63.1-rc.1
```
5. Deploy the updated manifest and verify that the Core Agent is using version `7.63.1-rc.1` via the node agent-level override, and that the Trace Agent using version `7.63.1-rc.2` based on its per-container image override.
6. Finally, we can update the image override for the Trace Agent to show using an entirely different image _or_ registry, even:

```yaml
metadata:
  annotations:
    # This would lead to the Trace Agent using `gcr.io/datadoghq/trace-agent:7.63.2`:
    experimental.agent.datadoghq.com/image-override-config: |
      {
        "trace-agent": {
          "name": "trace-agent"
        }
      }
```

```yaml
metadata:
  annotations:
    # This would lead to the Trace Agent using `myregistry.io/trace-agent:7.64.0`:
    experimental.agent.datadoghq.com/image-override-config: |
      {
        "trace-agent": {
          "name": "myregistry.io/trace-agent:7.64.0"
        }
      }
```

```yaml
metadata:
  annotations:
    # This would lead to the Trace Agent using `gcr.io/datadoghq/trace-agent:7.62.0`:
    experimental.agent.datadoghq.com/image-override-config: |
      {
        "trace-agent": {
          "name": "trace-agent",
          "tag": "7.62.0"
        }
      }
```

### Checklist


AliDatadog (approved):

vboulineau (commented):
Code LGTM, but not sure we want to offer this flexibility. Should we instead implement the standalone logic for relevant containers.

maycmlee (commented):
Hi @tobz! Is this a dupe of: https://github.com/DataDog/datadog-operator/pull/1719?
I had provided some editorial suggestions there, could you take a look at those for this PR, if this is getting merged?

tobz:
@maycmlee Ah, apologies!

This one is superseding #1719, but has all of the same relevant changes that you had commented on in #1719... so I'll go through and try and port over all of your suggestions. 👍🏻

maycmlee (approved):

internal/controller/datadogagent/experimental/image_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+func (i *imageOverrideTestDef) ImageOverrideJSON(containerName string) string {
+	imageOverrides := imageOverrides{
+		containerName: imageOverride{
+			Name: i.name,
+			Tag:  i.tag,
+		},
+	}
+
+	json, err := imageOverrides.JSON()
```
levan-m, line 34: nit - this test basically marshals `imageOverride` struct into json, sets it in annotation and then tests the logic, including unmarshalling. Is the goal of the test to verify unmarshalling as well? if not, why can't we just use `imageOverrides` map?
imo, having constant jsons in setup would be more readable and also test unmarshalling of predefined jsons.

internal/controller/datadogagent/experimental/image.go
```diff
@@ -0,0 +1,88 @@
...
+	// * `<REGISTRY>/<NAME>:<TAG>` - Override the entire image. `Tag` is ignored.
+	Name string `json:"name,omitempty"`
+
+	// Image tag to use.
+	//
+	// Used only when non-empty and `Name` does not specify a tag.
+	Tag string `json:"tag,omitempty"`
+}
+
+func (i imageOverrides) JSON() (string, error) {
```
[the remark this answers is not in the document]
tobz: That's fair. I'll remove it and put the logic right into `imageOverridesTestDef::ImageOverridesJSON`.

internal/controller/datadogagent/experimental/image_test.go
tobz: The intent is to make sure that we we're passing the intended structure: if `imageOverrides` changes over time, our test doesn't become stale because we're always generating it from the same struct that we deserialize into.

All of that said, I'm not terribly partial to this aspect and if you think it's more readable to generate the JSON directly, I'll make that change.

levan-m (approved):

PR #1759, opened 2025-03-05
[datadogpodautoscalers] Add local fallback policy

jennchenn:
### What does this PR do?

Add a policy to allow users to control the use of local fallback.

### Motivation

We are currently using the same targets for local fallback as with normal autoscaling. Users may want the flexibility to define different targets to scale on in the case that fallback should be activated.

### Additional Notes

### Minimum Agent Versions

* Cluster Agent: v7.65.0

### Describe your test plan

Will be tested as part of changes made to autoscaling controller for local fallback.

### Checklist


api/datadoghq/v1alpha2/datadogpodautoscaler_types.go
```diff
@@ -83,6 +83,17 @@ type DatadogPodAutoscalerSpec struct {
 	// +optional
 	Objectives []common.DatadogPodAutoscalerObjective `json:"objectives,omitempty"`
 
+	// FallbackApplyPolicy defines how recommendations should be applied when in fallback mode.
```
vboulineau, line 86: This feels like we override `ApplyPolicy` in local fallback, while we actually do. I think the idea is to allow users to view local fallback values without necessarily applying them in case of fallback, but it's not an `ApplyMode` I think.
I'm also not sure it's actually needed, but if we want this option:

We could perhaps simplify:
```
# There might be a better name than `fallback`?
fallback:
  # not `Apply` as it does not control apply, just generation of recommendations 
  mode: enabled|preview|disabled 
  # I am thinking that we could have some other triggers of local fallback. Also wondering if we should highlight horizontal only?
  triggers:
    noRecommendationsSeconds: 300
```

api/datadoghq/v1alpha2/datadogpodautoscaler_types.go
```diff
@@ -119,6 +145,31 @@ type DatadogPodAutoscalerApplyPolicy struct {
...
+	// +optional
+	// +kubebuilder:default=Apply
+	Mode DatadogPodAutoscalerFallbackMode `json:"mode"`
+
+	// ErrorWindowSeconds defines the time window the controller will wait after detecting an error before applying recommendations.
+	// +optional
+	// +kubebuilder:default=600
+	// +kubebuilder:validation:Minimum=180
+	// +kubebuilder:validation:Maximum=1200
+	ErrorWindowSeconds int32 `json:"errorWindowSeconds,omitempty"`
```
vboulineau, line 163: Should we ask users to distinguish these cases? I think the goal is to trigger if we have no data for X seconds

api/datadoghq/v1alpha2/datadogpodautoscaler_types.go
```diff
@@ -99,6 +108,18 @@ const (
 	DatadogPodAutoscalerApplyModePreview DatadogPodAutoscalerApplyMode = "Preview"
 )
 
+// DatadogPodAutoscalerFallbackMode specifies if the controller should apply/generate recommendations in fallback mode.
+// +kubebuilder:validation:Enum:=Apply;Disabled
+type DatadogPodAutoscalerFallbackMode string
```
vboulineau, line 113: Wondering if we want to keep the `Mode` if it's only enabled/disabled

jennchenn: Good point, I changed it to `Enabled  bool`  instead! (unless we foresee wanting to add another mode in the future?)

vboulineau (approved):

api/datadoghq/v1alpha2/datadogpodautoscaler_types.go
```diff
@@ -83,6 +87,11 @@ type DatadogPodAutoscalerSpec struct {
 	// +optional
 	Objectives []common.DatadogPodAutoscalerObjective `json:"objectives,omitempty"`
 
+	// HorizontalFallback defines how recommendations should be applied when in fallback mode.
+	// +optional
+	// +kubebuilder:default={}
+	HorizontalFallback *DatadogPodAutoscalerHorizontalFallbackPolicy `json:"horizontalFallback,omitempty"`
```
vboulineau, line 93: Perhaps we should do:
```
fallback:
  horizontal:
```

Just in case we want to add more later, it's cleaner from the top level

vboulineau (approved):

khewonc (approved):

PR #1778, opened 2025-03-11
Use object reconciler

khewonc:
### What does this PR do?

Use the object reconciler interface and separate out DDA validation

### Motivation

Newer and shinier way to do the same thing. Saves a few lines of code in the reconciler

[CECO-2101](https://datadoghq.atlassian.net/browse/CECO-2101)

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

* Should be no-op. Apply a DDA manifest. All components should spin up as normal

### Checklist

[CECO-2101]: https://datadoghq.atlassian.net/browse/CECO-2101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/datadoghq/v2alpha1/datadogagent_validation.go
+++ b/api/datadoghq/v2alpha1/datadogagent_validation.go
@@ -1,0 +1,18 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+package v2alpha1
+
+import "fmt"
+
+// ValidateDatadogAgent is used to check if a DatadogAgent is valid
+func ValidateDatadogAgent(dda *DatadogAgent) error {
+	// TODO
+	// Ensure required credentials are configured.
+	if dda.Spec.Global == nil || dda.Spec.Global.Credentials == nil {
+		return fmt.Errorf("credentials not configured in the DatadogAgent, can't reconcile")
+	}
+	return nil
+}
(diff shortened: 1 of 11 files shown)


internal/controller/datadogagent/controller_test.go
```diff
@@ -14,20 +14,9 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-func newRequest(ns, name string) reconcile.Request {
```
khewonc, line 22: I think we can delete this file entirely 🤔 Will do this in a separate PR though

api/datadoghq/v2alpha1/datadogagent_validation.go
```diff
@@ -0,0 +1,18 @@
...
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+package v2alpha1
+
+import "fmt"
+
+// ValidateDatadogAgent is used to check if a DatadogAgent is valid
+func ValidateDatadogAgent(dda *DatadogAgent) error {
```
tbavelier, line 11: Not for this PR, but we should get rid of this validation in code and instead use kubebuilder markers: something like `+kubebuilder:validation:Required` or `+kubebuilder:validation:MinLength` https://book.kubebuilder.io/reference/markers/crd-validation

internal/controller/datadogagent/controller_test.go
tbavelier: Makes sense, was indeed the only function used

tbavelier (approved):
🔥

PR #1792, opened 2025-03-13
Allow disabling features that are enabled by default

khewonc:
### What does this PR do?

* Splits features into configured and enabled features, where configured features are features that are not enabled but may require configuration
* Changes the way we define a `RequiredComponent`'s `IsEnabled` and `IsConfigured`
  * `IsEnabled`: `nil` = feature disabled, `true` = feature enabled, `false` should not be used
  * `IsConfigured`: `true` = at least one container specified in `RequiredComponent.Containers`, `false` = no containers in `RequiredComponent.Containers`
* Adds `RequiredComponent.Containers` to features that previously only set `RequiredComponent.IsEnabled`
* Adds a path to disable the orchestrator explorer feature

### Motivation

https://datadoghq.atlassian.net/browse/CECO-1895

### Additional Notes

Have only checked the orchestrator explorer and remote config features. Other features may be enabled by default in the agent but have not had config added to allow them to be disabled

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Disable the orchestrator explorer and/or remote config feature and ensure the appropriate `xxx_enabled: false` env vars are set

### Checklist

### DIFF
--- a/internal/controller/datadogagent/feature/clusterchecks/feature.go
+++ b/internal/controller/datadogagent/feature/clusterchecks/feature.go
@@ -72,9 +72,18 @@ func (f *clusterChecksFeature) Configure(dda *v2alpha1.DatadogAgent) (reqComp fe
 
 		f.useClusterCheckRunners = apiutils.BoolValue(dda.Spec.Features.ClusterChecks.UseClusterChecksRunners)
 		reqComp = feature.RequiredComponents{
-			Agent:               feature.RequiredComponent{IsRequired: apiutils.NewBoolPointer(true)},
-			ClusterAgent:        feature.RequiredComponent{IsRequired: apiutils.NewBoolPointer(true)},
-			ClusterChecksRunner: feature.RequiredComponent{IsRequired: &f.useClusterCheckRunners},
+			Agent: feature.RequiredComponent{
+				IsRequired: apiutils.NewBoolPointer(true),
+				Containers: []apicommon.AgentContainerName{apicommon.CoreAgentContainerName},
+			},
+			ClusterAgent: feature.RequiredComponent{
+				IsRequired: apiutils.NewBoolPointer(true),
+				Containers: []apicommon.AgentContainerName{apicommon.ClusterAgentContainerName},
+			},
+			ClusterChecksRunner: feature.RequiredComponent{
+				IsRequired: &f.useClusterCheckRunners,
+				Containers: []apicommon.AgentContainerName{apicommon.CoreAgentContainerName},
+			},
 		}
 	}
 
--- a/internal/controller/datadogagent/feature/orchestratorexplorer/feature.go
+++ b/internal/controller/datadogagent/feature/orchestratorexplorer/feature.go
@@ -45,6 +45,7 @@ func buildOrchestratorExplorerFeature(options *feature.Options) feature.Feature 
 }
 
 type orchestratorExplorerFeature struct {
+	enabled                  bool
 	runInClusterChecksRunner bool
 	scrubContainers          bool
 	extraTags                []string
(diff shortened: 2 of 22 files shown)


internal/controller/datadogagent/feature/clusterchecks/feature.go
```diff
@@ -72,9 +72,18 @@ func (f *clusterChecksFeature) Configure(dda *v2alpha1.DatadogAgent) (reqComp fe
 
 		f.useClusterCheckRunners = apiutils.BoolValue(dda.Spec.Features.ClusterChecks.UseClusterChecksRunners)
 		reqComp = feature.RequiredComponents{
-			Agent:               feature.RequiredComponent{IsRequired: apiutils.NewBoolPointer(true)},
-			ClusterAgent:        feature.RequiredComponent{IsRequired: apiutils.NewBoolPointer(true)},
-			ClusterChecksRunner: feature.RequiredComponent{IsRequired: &f.useClusterCheckRunners},
+			Agent: feature.RequiredComponent{
+				IsRequired: apiutils.NewBoolPointer(true),
+				Containers: []apicommon.AgentContainerName{apicommon.CoreAgentContainerName},
```
tbavelier, line 77: That's out of scope for this PR, but technically, we don't need core Agent if useclustercheckrunners is enabled

internal/controller/datadogagent/feature/orchestratorexplorer/feature.go
```diff
@@ -181,6 +185,11 @@ func (f *orchestratorExplorerFeature) ManageDependencies(managers feature.Resour
 // ManageClusterAgent allows a feature to configure the ClusterAgent's corev1.PodTemplateSpec
 // It should do nothing if the feature doesn't need to configure it.
 func (f *orchestratorExplorerFeature) ManageClusterAgent(managers feature.PodTemplateManagers) error {
+	managers.EnvVar().AddEnvVar(f.getEnabledEnvVar())
+	if !f.enabled {
+		return nil
+	}
```
tbavelier, line 191:
```suggestion
        // We need to explicitly add the env var to disable this feature, otherwise, we rely on Agent default
        // where the feature is enabled by default, and would error out.
	managers.EnvVar().AddEnvVar(f.getEnabledEnvVar())
	if !f.enabled {
		return nil
	}
```
Wording subject to change, but I think it's worth pointing out why this needs to be the first line, as my first thought would have been to exit early if f is not enabled

internal/controller/datadogagent/feature/orchestratorexplorer/feature.go
```diff
@@ -231,11 +245,18 @@ func (f *orchestratorExplorerFeature) ManageSingleContainerNodeAgent(managers fe
 // ManageNodeAgent allows a feature to configure the Node Agent's corev1.PodTemplateSpec
 // It should do nothing if the feature doesn't need to configure it.
 func (f *orchestratorExplorerFeature) ManageNodeAgent(managers feature.PodTemplateManagers, provider string) error {
+	containers := []apicommon.AgentContainerName{apicommon.CoreAgentContainerName}
+	if f.processAgentRequired {
+		containers = append(containers, apicommon.ProcessAgentContainerName)
+	}
+
+	managers.EnvVar().AddEnvVarToContainers(containers, f.getEnabledEnvVar())
```
tbavelier, line 253: same as above on adding a comment on truly disabling the feature

internal/controller/datadogagent/feature/orchestratorexplorer/feature.go
```diff
@@ -245,6 +266,11 @@ func (f *orchestratorExplorerFeature) ManageNodeAgent(managers feature.PodTempla
 // It should do nothing if the feature doesn't need to configure it.
 func (f *orchestratorExplorerFeature) ManageClusterChecksRunner(managers feature.PodTemplateManagers) error {
 	if f.runInClusterChecksRunner {
+		managers.EnvVar().AddEnvVar(f.getEnabledEnvVar())
```
tbavelier, line 269: ditto

tbavelier (approved):

internal/controller/datadogagent/feature/clusterchecks/feature.go
khewonc: Agree. That'll be good for a future PR

PR #1801, opened 2025-03-19
Enable pod-resources mount only with GPU monitoring enabled

gjulianm:
### What does this PR do?

This PR ensures that the pod-resources socket mount is only mounted when the GPU monitoring feature is enabled.

### Motivation

https://datadoghq.atlassian.net/browse/EBPF-681?atlOrigin=eyJpIjoiODQzODBkODQxMzYxNDM3MDhiMDU3YTU1MjM5OGNlZTAiLCJwIjoiaiJ9

### Additional Notes

### Minimum Agent Versions

* Agent: N/A
* Cluster Agent: N/A

### Describe your test plan

pod-resources socket isn't mounted by default, only mounted when `feature.gpu` is set to true.

Unit tests should cover this case.

### Checklist


internal/controller/datadogagent/feature/gpu/feature_test.go
```diff
@@ -145,11 +165,19 @@ func Test_GPUMonitoringFeature_Configure(t *testing.T) {
 				Value: "all",
 			},
 		}
+
+		wantAgentEnvVars := append([]*corev1.EnvVar{
+			{
+				Name:  override.DDKubernetesPodResourcesSocket,
```
celenechang, line 171:
```suggestion
				Name:  common.DDKubernetesPodResourcesSocket,
```

internal/controller/datadogagent/override/global.go
```diff
@@ -286,27 +285,6 @@ func applyGlobalSettings(logger logr.Logger, manager feature.PodTemplateManagers
 					Value: agentCAPath,
 				})
 			}
-			if config.Kubelet.PodResourcesSocketPath != "" {
```
celenechang, line 289: sorry i didn't block for this part to removed in my suggestion - isn't this needed?

gjulianm: No, because now we're only configuring it from the GPU feature. I did not see a way to enable common configs such as this from the specific feature (for example, the common system-probe config is repeated in each feature that requires system-probe).

levan-m (approved):

PR #1833, opened 2025-03-31
[kubectl plugin] Fallback to configmap leader attempt for `clusteragent` if lease failed

tbavelier:
### What does this PR do?

Falls back to try getting leader from configmap for `clusteragent` even if leases are supported, in the case getting leader from lease failed

### Motivation

Allows to get leader from configmap even if leases are supported in the environment, in a situation where configmap is still being used for leader election of DCA

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

* As https://github.com/DataDog/datadog-operator/pull/1375, check by default, we try to retrieve leader when `leader_election_default_resource` is set to `lease`
* In an env with DCA leader election in configmap (https://github.com/DataDog/datadog-agent/blob/5e42ef8869290dcdec42772b0b71538485879f8b/pkg/config/setup/config.go#L515) but leases supported in the cluster, ensure you get the DCA leader

### Checklist

### DIFF
--- a/cmd/kubectl-datadog/clusteragent/leader/leader.go
+++ b/cmd/kubectl-datadog/clusteragent/leader/leader.go
@@ -33,7 +33,8 @@ type options struct {
 	common.Options
 	args []string
 
-	DDAName string
+	DDAName        string
+	forceConfigMap bool
 }
 
 type leaderResponse struct {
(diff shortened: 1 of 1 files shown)


clamoriniere (approved):
for now we don't make a difference between lease not found or RBAC issue. 
maybe a jira card can be created to track this improvement, but IMO not needed now. 

thanks for the fix 🙇

levan-m (approved):

PR #1477, opened 2024-10-22
[DatadogAgent][Flare] Additional environment variables and RBAC to permit user manifests retrieval

tbavelier:
### What does this PR do?

* Adds 3 env vars to the cluster Agent:
    * `AGENT_DAEMONSET`: name of the node Agent ds
    * `CLUSTER_AGENT_DEPLOYMENT`: name of the DCA deployment
    * `DATADOGAGENT_CR_NAME`: name of the `DatadogAgent` custom resource
* Adds RBAC allowing the DCA to `get` the `DatadogAgent` custom resource from its namespace

### Motivation

Similar to https://github.com/DataDog/helm-charts/pull/965, we need to inject a few variables/give some permissions to be able to collect the Agent deployment files while creating a flare to help with troubleshooting.

### Additional Notes

The flare feature depending on these variables is available at: https://github.com/DataDog/datadog-agent/pull/30372
### Minimum Agent Versions

Are there minimum versions of the Datadog Agent and/or Cluster Agent required?

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

* Deploy the operator
* Deploy a `DatadogAgent` with cluster Agent enabled (default):
```yaml
apiVersion: datadoghq.com/v2alpha1
kind: DatadogAgent
metadata:
  name: foo
```
* Assert on DCA:
    * `AGENT_DAEMONSET`: `foo-agent`, value of `kubectl get ds`
    * `CLUSTER_AGENT_DEPLOYMENT`: `foo-cluster-agent`, value of `kubectl get deploy`
    * `DATADOGAGENT_CR_NAME`: `foo`, value of `kubectl get dd`
* Add overrides:
```yaml
apiVersion: datadoghq.com/v2alpha1
kind: DatadogAgent
metadata:
  name: foo
override:
  nodeAgent:
    name: "x"
  clusterAgent:
    name: "y"
```
* Assert on DCA:
    * `AGENT_DAEMONSET`: `x`, value of `kubectl get ds`
    * `CLUSTER_AGENT_DEPLOYMENT`: `y`, value of `kubectl get deploy`
    * `DATADOGAGENT_CR_NAME`: `foo`, value of `kubectl get dd`

### Checklist

- [x] PR has at least one valid label: `bug`, `enhancement`, `refactoring`, `documentation`, `tooling`, and/or `dependencies`
- [x] PR has a milestone or the `qa/skip-qa` label

PR #1866, opened 2025-04-16
Move install info and seccomp out of enabledefault feature

khewonc:
### What does this PR do?

* Moves install info and default seccomp into global
* [enhancement] Only create the default seccomp when a custom seccomp configmap isn't specified for the system probe container
* [bug] 🐛 Fixes flare env var names messed up in a previous PR
* Removes `RequiredComponents` arg from `ManageDependencies `
* Split global dependencies into global and component specific dependencies so we only try to add the install info configmap to the store once per reconcile rather than once for each component enabled

### Motivation

https://datadoghq.atlassian.net/browse/CECO-2105

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

* Tested that the install info configmap was created
* Tested that a seccomp configmap was created when enabling the system probe container
* Tested that we don't create a seccomp configmap when a custom seccomp configmap is referenced in the DDA

Note: We don't delete the seccomp configmap if it was already created and then unneeded in a subsequent DDA update. I tested on 1.13.0 and I don't believe we do there either. Creating a low priority card for this in the backlog since it's not blocking anything and I doubt this is a seccomp specific issue

### Checklist


internal/controller/datadogagent/component/agent/default.go
```diff
@@ -87,6 +87,232 @@ func DefaultCapabilitiesForSystemProbe() []corev1.Capability {
 	}
 }
 
+// DefaultSeccompConfigDataForSystemProbe returns configmap data for the default seccomp profile
```
khewonc, line 90: Copied from [configmap.go](https://github.com/DataDog/datadog-operator/pull/1866/files#diff-aad89724ca50ab8255422b7cab682937ec0eede3e4a24f5be5dabec90571a72f)

internal/controller/datadogagent/component/clusteragent/envvar.go
```diff
@@ -10,7 +10,7 @@ const (
 	DDKubeResourcesNamespace         = "DD_KUBE_RESOURCES_NAMESPACE"
 	DDPodName                        = "DD_POD_NAME"
 	DDClusterAgentServiceAccountName = "DD_CLUSTER_AGENT_SERVICE_ACCOUNT_NAME"
-	DDAgentDaemonSet                 = "DD_AGENT_DAEMONSET"
-	DDClusterAgentDeployment         = "DD_CLUSTER_AGENT_DEPLOYMENT"
-	DDDatadogAgentCustomResource     = "DD_DATADOG_AGENT_CUSTOM_RESOURCE"
+	DDAgentDaemonSet                 = "AGENT_DAEMONSET"
```
khewonc, line 13: Flare env vars should match the ones in https://github.com/DataDog/datadog-operator/pull/1477

internal/controller/datadogagent/global/utils.go
```diff
@@ -51,3 +54,24 @@ func setImageRegistry(manager feature.PodTemplateManagers, dda *v2alpha1.Datadog
...
+		toolVersion = envVar
+	}
+
+	return fmt.Sprintf(installInfoDataTmpl, toolVersion, version.Version)
+}
+
+func useCustomSeccomp(dda *v2alpha1.DatadogAgent) bool {
+	if componentOverride, ok := dda.Spec.Override[v2alpha1.NodeAgentComponentName]; ok {
+		if container, ok := componentOverride.Containers[apicommon.SystemProbeContainerName]; ok {
+			// Only ConfigMap is supported for now
```
tbavelier, line 70: Don't we also support providing directly configData ? The operator will create a CM for it, but don't think it will be in the dda spec here ?

khewonc: I don't think we do anything with seccomp's configData in the code. I could only find custom seccomp data being used in [the overrides](https://github.com/DataDog/datadog-operator/blob/1aa8dc0fc9641dda6f0de46700a82d1746fe052b/internal/controller/datadogagent/override/container.go#L172) and I don't see any codepaths using `ConfigData`. I'll make a card for it in our backlog

tbavelier (approved):

Issue #1844, opened 2025-04-04
DataDogSLO Monitor type with DataDogMonitor name support


Alex-Caesar:
Currently, if you set a DataDogSLO CRD to be a monitor type then you must set the monitor_ids as seen [here](https://docs.datadoghq.com/api/latest/service-level-objectives/#create-an-slo-object:~:text=%5Binteger%5D-,A%20list%20of%20monitor%20IDs%20that%20defines%20the%20scope%20of%20a%20monitor%20service%20level%20objective.%20Required%20if%20type%20is%20monitor.,-name%C2%A0%5Brequired). 

If you want to point a DataDogSLO to DataDogMonitor then you need to know the id which is only available post creation and operator reconcile loop. This presents difficulties to automate CRDs. 

I would like to suggest that users can specifiy a list of monitor_names (DatadogMonitor.spec.name) so that we can create a couple of DatadogMonitors then directly reference with a DatadogSLO monitory type

joelizard:
This would be an incredibly useful feature. Similarly, the requirement to manually specify the SLO ID when setting up a burn rate monitor presents similar issues & an equivalent solution (ie. being able to reference the DatadogMonitor custom resource) would be welcomed

levan-m:
Please see the update on the request here
https://github.com/DataDog/datadog-operator/pull/1871#issuecomment-2940370704

13013SwagR:
We very interested in this, I'll mention it to our DD rep

Issue #1022, opened 2023-12-20
SLO burn rate


melchiormoulin:
Hello, since there a customressource to define SLO https://github.com/DataDog/datadog-operator/pull/807
How can we define burn rate alert based on this SLO CR ?
It seems there is some internal discussions ( this comment https://github.com/DataDog/datadog-operator/pull/807#issuecomment-1781049308 )

vlinevych:
I've found an [example of slo alert](https://github.com/DataDog/datadog-operator/blob/main/examples/datadogmonitor/slo-alert-monitor-test.yaml), but the problem is that it requires `slo-hash-id` in the query. 

It would be great to have slo burn rate alert implemented in the operator.

melchiormoulin:
Hello do you think there is any chance to see this feature please @fanny-jiang @ddjamesfrullo  ? :)

jurabek:
There is possibility to create CRD only for burn rate and pass those values as attributes instead hard coded string `query: "burn_rate(\"slo-hash-id\").over(\"30d\").long_window(\"1h\").short_window(\"5m\") > 14.4"` value, and this gives us referencing parent SLO into this alert more natively.

melchiormoulin:
Hey do you think it can be implemented ? Maybe @jurabek ?

melchiormoulin:
i've created https://github.com/DataDog/datadog-operator/pull/1871 wdyt ?

jacjtito:
Hi, 

I'll close this issue since the Feature Request is tracked in the PR: https://github.com/DataDog/datadog-operator/pull/1871. 

Thanks!

PR #1871, opened 2025-04-21
Enable to add a reference of DatadogSLO in DatadogMonitor SLO alert type.

melchiormoulin:
The goal is to be able to link DatadogSLO and DatadogMonitor for SLO alert type.

### What does this PR do?

When a user set a SLORef in DatadogMonitor CRD.
The controller query the corresponding DatadogSLO and fetch the ID from the status.
It then replace the ID inside the DatadogMonitor query.
### Motivation

Enable the user to create a datadogmonitor SLO alert type without the need of knowing the SLO ID.

It would solves thoses issues : 
https://github.com/DataDog/datadog-operator/issues/1022
https://github.com/DataDog/datadog-operator/issues/1844

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Create A DatadogSLO.
Create a DatadogMonitor and refer the previous DatadogSLO in SLORef. 
The controller should update the id inside the query

### Checklist


celenechang:
Hi @melchiormoulin , thank you for your contribution. Do you mind if I make a few edits on your branch?

melchiormoulin:
I don't mind at all.

celenechang:
@melchiormoulin Sorry for the delay; we've been having internal discussions about this change, and need some more time to consider it. For now we are going to move the milestone to the next Operator release. We will keep you posted on any updates

levan-m:
@melchiormoulin, sorry for additional delay. Unfortunately we can’t provide any ETA when and if we can merge this change.

This has nothing to do with the the implementation in the PR, rather how we see purpose of resource CRDs. When we created resource CRDs the assumption was that users would managed those independently. This PR introduces dependency between two resources, creating a precedent potentially pushing us into general dependency management which is a complex undertaking considering how monitors, SLOs, composites, synthetics etc. can reference each other.

We are contemplating if this is something we want to commit to and surveying customers how they manage dependencies now.

melchiormoulin:
Hello thanks for the response ! I understand that you don't want to link those 2 CRD in a tech perspectives but as a user and for automation ( particularly if we packaged it those 2 CR with helm ) it will be a lot more convenient.

levan-m:
Removing milestone and keeping open since we haven't yet prioritized looking deeper into this to make any decision. Can't provide ETA now when we can prioritize it.

ahatzz11:
Bumping this so it doesn't get closed - Please allow the linking between a `DatadogMonitor` and a `DatadogSLO`. The need to manage these separately is quite annoying.

zg-brian:
This would be an extremely valuable feature for creating burn rate monitors, and not have to specify the SLO ID in the monitor definition.

drewwells:
We can not use this operator without a declarative approach to managing metric-based SLOs. Designing in a way that users have to go into the datadog api and copy out fields, then update their k8s is very cumbersome. I have not seen any other operators that do it this way. The ref approach is common through k8s ecosystem and supporting CNCF tools.

13013SwagR:
Very interested in this, do you have an update on ETA @levan-m  ?

levan-m:
@13013SwagR such extension isn't on our roadmap in foreseeable future. We aren't convinced that Operator is the right tool to validate (100s-1000s of) resources, build DAGs, execute ordered creation while reconciling individual resources in independent control loops.

I'm closing this PR, please share your thoughts/questions on https://github.com/DataDog/datadog-operator/issues/2702.

drewwells:
That’s really unfortunate. I don’t actually advertise this as an operator to developers. K8s and k8s operators are declarative and this is a highly manual/imperative API. It’s going to handicap your adoption for k8s solutions. The resource tracking does not make a ton of sense either. This would only be an issue if you were doing some sort of loop based tracking. Datadog objects and dependencies would rarely change

PR #1889, opened 2025-04-30
Clean up finalizer

khewonc:
### What does this PR do?

Clean up existing finalizer. Most dependencies should be deleted through owner refs so we don't need to recreate the dependencies in the store. Cluster scoped dependencies are delete manually based on label

### Motivation

https://datadoghq.atlassian.net/browse/CECO-2210

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Create a DatadogAgent and delete it. Ensure dependencies related to the DDA are deleted, e.g. configmaps, secrets, etc. Be sure to check that cluster scoped resources such as clusterroles and clusterrolebindings are deleted.

### Checklist

### DIFF
--- a/internal/controller/datadogagent_controller.go
+++ b/internal/controller/datadogagent_controller.go
@@ -68,6 +68,11 @@ type DatadogAgentReconciler struct {
 // +kubebuilder:rbac:groups=authorization.k8s.io,resources=rolebindings,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=pods/exec,verbs=create
 
+// Finalizer
+// +kubebuilder:rbac:groups=authorization.k8s.io,resources=clusterroles,verbs=deletecollection
+// +kubebuilder:rbac:groups=authorization.k8s.io,resources=clusterrolebindings,verbs=deletecollection
+// +kubebuilder:rbac:groups=authorization.k8s.io,resources=apiservices,verbs=deletecollection
+
 // Configure Admission Controller
 // +kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=validatingwebhookconfigurations;mutatingwebhookconfigurations,verbs=*
 // +kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=get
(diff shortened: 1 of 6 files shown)


internal/controller/datadogagent_controller.go
```diff
@@ -54,8 +54,8 @@ type DatadogAgentReconciler struct {
 // +kubebuilder:rbac:groups=datadoghq.com,resources=datadogagents/finalizers,verbs=get;list;watch;create;update;patch;delete
 
 // RBAC Management
-// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterrolebindings,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles,verbs=get;list;watch;create;update;patch;delete;deletecollection
+// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterrolebindings,verbs=get;list;watch;create;update;patch;delete;deletecollection
```
tbavelier, line 58: I added `deletecollection` here as we already have this section for the `rbac.authorization.k8s.io`, but we could instead add it further below in the Finalizer section to highlight it for clarity as you did in the original PR (before my commits)

internal/controller/datadogagent_controller.go
```diff
@@ -68,6 +68,11 @@ type DatadogAgentReconciler struct {
 // +kubebuilder:rbac:groups=authorization.k8s.io,resources=rolebindings,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=pods/exec,verbs=create
 
+// Finalizer (cluster-scoped resources)
+// Note: we already have verbs=* for apiservices (External Metrics server feature requires it), so this is not needed.
+// However, we keep it here for clarity.
+// +kubebuilder:rbac:groups=apiregistration.k8s.io,resources=apiservices,verbs=deletecollection
```
tbavelier, line 74: As we already have `*` on `apiregistration.k8s.io/apiservices`, do we want to still keep this section ?

tbavelier (approved):
Approved after modifying the markers to use `rbac.authorization.k8s.io`, as `deletecollection` requires it:
```shell
kubectl auth can-i deletecollection clusterroles --as system:serviceaccount:system:datadog-operator-controller-manager
ok
Warning: resource 'clusterroles' is not namespace scoped in group 'rbac.authorization.k8s.io'

no
```
was what I was getting from the original PR, and clusterroles/bindings were not being removed.

PR #1891, opened 2025-04-30
Add DDA to DDAI reconciler

khewonc:
### What does this PR do?

* Adds a reconcile path for DDA -> DDAI, called `reconcileInstanceV3`
* Change DAPs to use DatadogAgentSpec directly. I don't believe this should be a breaking change since the custom resource should stay the same

### Motivation

https://datadoghq.atlassian.net/browse/CECO-2014
https://datadoghq.atlassian.net/browse/CECO-2234

### Additional Notes

* DDA status is incomplete. Have not set remote config, conditions, and possibly needs more work for DCA token

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist


internal/controller/datadogagent/common/affinity.go
```diff
@@ -0,0 +1,164 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
khewonc, line 1: Moved from override/podtemplatespec.go

internal/controller/datadogagent/utils.go
```diff
@@ -135,3 +140,32 @@ func getReplicas(currentReplicas, newReplicas *int32) *int32 {
 
 	return apiutils.NewInt32Pointer(*newReplicas)
 }
+
+// getDDAICRDFromConfig is only used in tests
```
khewonc, line 144: Ended up using this that gets the CRD from the `config` directory. Go embed wasn't working because of how the directory is structured

PR #1905, opened 2025-05-07
Refactor defaulting pkg into images

celenechang:
### What does this PR do?

- Refactor `Image` struct and usage
- Rename pkg from `defaulting` to `images`
- Consolidate image updates on PodTemplateSpec from `global` config into one function

### Motivation

When editing or updating images in the PodTemplateSpec, the general workflow should be

1. Convert from a `AgentImageConfig` or `string` to `Image`
2. Manipulate as needed
3. Convert to `string`

Example use cases:
- Changing the registry
- Adding suffixes
- Overriding (parts of) images

### Additional Notes

The main changes are in pkg/images and global/global.go

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Test the following spec versions:

- [ ] default (no image override)
- [ ] fips enabled (global.useFIPSAgent)
- [ ] jmx override (image.jmxEnabled under nodeAgent override)
- [ ] fips enabled and jmx override
- [ ] agent override image tag with "-jmx" in the tag
- [ ] standard image override for agent
- [ ] standard image override for dca
- [ ] standard image override for ccr

### Checklist


pkg/images/images.go
```diff
@@ -0,0 +1,203 @@
...
+	registry string
+	name     string
+	tag      string
+	isJMX    bool
+	isFIPS   bool
+}
+
+// NewImage return a new Image instance
+// Assumes that tag suffixes have already been trimmed and booleans updated accordingly
+func NewImage(registry, name, tag string, isJMX bool, isFIPS bool) *Image {
```
levan-m, line 67:
```suggestion
func newImage(registry, name, tag string, isJMX bool, isFIPS bool) *Image {
```
no need to export this.

pkg/images/images.go
```diff
@@ -0,0 +1,203 @@
...
+
+	img := NewImage(registry, imageSpec.Name, tag, imageSpec.JMXEnabled, false)
+
+	return img.ToString()
+}
+
+// OverrideAgentImage takes an existing image reference and potentially overrides portions of it based on the provided image configuration
+func OverrideAgentImage(currentImage string, overrideImageSpec *v2alpha1.AgentImageConfig) string {
+	image := FromString(currentImage)
+	overrideImage := fromImageConfig(overrideImageSpec)
```
levan-m, line 142: I think this overlooks the fact that [`AgentImageConfig.Name`](https://github.com/DataDog/datadog-operator/blob/main/api/datadoghq/v2alpha1/datadogagent_types.go#L1225-L1232) can be fully qualified name, or just name or name:tag. That's why tests are failing.

pkg/images/images.go
```diff
@@ -0,0 +1,203 @@
...
+	return &Image{
+		registry: registry,
+		name:     name,
+		tag:      tag,
+		isJMX:    isJMX,
+		isFIPS:   isFIPS,
+	}
+}
+
+func (i *Image) WithRegistry(registry string) {
```
levan-m, line 77: nit - returning `i` would allow chaining these functions.

internal/controller/datadogagent/global/global.go
```diff
@@ -236,22 +236,35 @@ func applyGlobalSettings(logger logr.Logger, manager feature.PodTemplateManagers
...
+	if !*config.UseFIPSAgent && config.FIPS != nil && apiutils.BoolValue(config.FIPS.Enabled) {
 		applyFIPSConfig(logger, manager, dda, resourcesManager)
 	}
 
 }
 
+func updateContainerImages(config *v2alpha1.GlobalConfig, podTemplateManager feature.PodTemplateManagers) {
+	image := &images.Image{}
+	for i, container := range podTemplateManager.PodTemplateSpec().Spec.Containers {
+		image = images.FromString(container.Image)
```
levan-m, line 252: nit - `FromString` can be made a method of `Image` and this could be done like
```go
image.FromString(container.Image).
  WithRegistry(..).
  WithFips(..).ToString()
```

levan-m (commented):
Refactor looks good but we lost handling of `AgentImageConfig.Name`. Left some nit suggestions too.

internal/controller/datadogagent/global/global.go
celenechang: thanks, partially took the suggestion as i found it easier to reason if it's not a struct method

celenechang:
In the latest commit, I updated the behavior of the image override function to match what we had, with one exception - if there is a registry in the current image, and it is not overridden with a new registry, then I preserved that current registry. IMO this is more consistent with the code flow because we always initialize the container images with the default registry, and then overwrite this registry with a global registry setting, before checking the override image config.

internal/controller/datadogagent/override/podtemplatespec_test.go
```diff
@@ -204,64 +204,6 @@ func TestPodTemplateSpec(t *testing.T) {
...
-				},
-			},
-			validateManager: func(t *testing.T, manager *fake.PodTemplateManagers) {
-				assertImageConfigValues(
-					manager,
-					containerImageOptions{name: "agent:latest"},
-					t,
-				)
-			},
-		},
```
celenechang, line 264: Removed a few test cases because these situations (name:tag set for the image) cannot occur with the current flow of code.

pkg/images/images.go
```diff
@@ -0,0 +1,237 @@
...
+// (Notably, we do not accept "registry/name".)
+// Note that if the name includes a tag, then we ignore imageConfig.tag and imageConfig.JMXEnabled
+func fromImageConfig(imageConfig *v2alpha1.AgentImageConfig) *Image {
+	registry := ""
+	imageName := imageConfig.Name
+	imageTag := imageConfig.Tag
+
+	nameContainsTag := false
+	isJMX := false
+	isFIPS := false
```
levan-m, line 208: this will always be false unless `fips` is in tag so only settings `jmxEnabled` in overrdie won't produce `fips-jmx` image, not sure if it should.

celenechang: thanks, yea it's expected that if the tag is included `jmxEnabled` will be ignored (as noted on line 200). however it's a good point that `isFIPS` will always be false then no matter what, so i think i will remove this bool to reduce confusion

levan-m (approved):

PR #1926, opened 2025-05-23
update metrics forwarder, collect custom resource metrics

celenechang:
### What does this PR do?

- Rename `forwarders` to `metricsForwarders` in anticipation of adding metadata forwarder in a future PR
- Update metrics forwarder setup to collect credentials from the Operator configuration if present, and then DDA otherwise
- Add metrics forwarder to DatadogMonitor controller
- Collect resource count metric for both DDA and DM controllers

### Motivation

### Additional Notes

In subsequent PRs, 
- the metrics forwarder will be added to other controllers
- a metadata forwarder will be added

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

1. Run the Operator with both DatadogAgent and DatadogMonitor controllers enabled
Apply at least one DDA and at least one DM
Verify in the Datadog web app that the `datadog.operator.datadog_agent.custom_resource.count` and `datadog.operator.datadog_monitor.custom_resource.count` metrics are reporting, with the following tags:

- `"kube_namespace:%s"`
- `"resource_name:%s"`
- `"kube_cluster_name:%s`

2. Also make sure that metrics forwarder metrics work:
- [] with operator credentials only (including cluster name), and a DDA
- [] with operator credentials, except for cluster name, and a DDA
- [] without operator credentials, but with DDA
- [] with operator credentials and a DatadogMonitor, but no DDA
- [] with a DatadogMonitor, but no cluster name set on operator

### Checklist


internal/controller/datadogmonitor/controller.go
```diff
@@ -62,33 +62,57 @@ const requiredTag = "generated:kubernetes"
...
+		forwarders:             metricForwardersMgr,
 	}, nil
 }
 
 // Reconcile is similar to reconciler.Reconcile interface, but taking a context
 func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
-	return r.internalReconcile(ctx, request)
+	// Get instance
+	instance := &datadoghqv1alpha1.DatadogMonitor{}
+	err := r.client.Get(ctx, request.NamespacedName, instance)
```
levan-m, line 93: nit - if we are touching/moving this maybe we could refactor it using generic Reconcile interface done for agent controller https://github.com/DataDog/datadog-operator/pull/1778. This would save this Get call and error handling.

pkg/controller/utils/datadog/metrics_forwarder.go
```diff
@@ -245,13 +254,47 @@ func (mf *metricsForwarder) setStatus(newStatus *ConditionCommon) {
...
+			// API key
+			mf.apiKey = creds.APIKey
+
+			// base URL
+			mf.baseURL = defaultbaseURL
+			mf.logger.V(1).Info("Got API URL for DatadogAgent", "site", mf.baseURL)
+			if os.Getenv(constants.DDddURL) != "" {
+				mf.baseURL = os.Getenv(constants.DDddURL)
+			} else if os.Getenv(constants.DDURL) != "" {
+				mf.baseURL = os.Getenv(constants.DDURL)
```
levan-m: I don't see DD_URL in operator helm chart 🤔 
https://github.com/DataDog/helm-charts/blame/06aac8eed2462e88c483c4508178369bb656e866/charts/datadog/templates/_components-common-env.yaml#L59-L66

levan-m (approved):

pkg/controller/utils/datadog/metrics_forwarder.go
celenechang: this one would be coming from the operator chart here https://github.com/DataDog/helm-charts/blob/main/charts/datadog-operator/templates/deployment.yaml#L113 

`DD_DD_URL` is also there in that case, but technically we support both

internal/controller/datadogmonitor/controller.go
celenechang: I'm going to make this change in a different PR - it's impacting tests and I need more time to figure it out

pkg/version/version.go
```diff
@@ -81,3 +81,8 @@ func printVersionSlice() []string {
 		fmt.Sprintf("Go OS/Arch: %s/%s", runtime.GOOS, runtime.GOARCH),
 	}
 }
+
+// GetVersion returns the operator version
+func GetVersion() string {
```
levan-m: nit - this seems unused

levan-m (approved):

PR #1927, opened 2025-05-23
[CONTINT-4643] Add an option to configure KSM custom resource metrics collection

L3n41c:
### What does this PR do?

Add an datadog agent parameter to configure custom resources metrics collection introduced in DataDog/datadog-agent#31715.

### Additional Notes

This the `datadog-operator` equivalent of DataDog/helm-charts#1883.

### Minimum Agent Versions

* Agent: v7.63.0 (if `kubernetes-state-core` is running on cluster checks runners)
* Cluster Agent: v7.63.0 (otherwise)

### Describe your test plan

Create a `DatadogAgent` CR with the following piece of KSM configuration:
```yaml
spec:
  features:
    kubeStateMetricsCore:
      collectCrMetrics:
      - groupVersionKind:
          group: datadoghq.com
          kind: DatadogAgent
          version: v2alpha1
        metrics:
        - each:
            gauge:
              path:
              - status
              - agent
              - upToDate
            type: gauge
          help: Number of up-to-date agents
          name: uptodateagents
```

And validate that the corresponding custom resource metric is properly emitted:
```console
$ kubectl exec pod/dda-with-operator-cluster-agent-79988798b9-9t7f7 -c cluster-agent -- env -u DD_LOG_LEVEL agent check -r kubernetes_state_core --table
[…]
  kubernetes_state_customresource.uptodateagents             gauge  1755816713  1             kube_cluster_name:lenaic-kind, customresource_group:datadoghq.com, customresource_kind:DatadogAgent, customresource_version:v2alpha1, orch_cluster_id:596fb22b-2124-4b0a-91a9-dca1b37078f8
```

### Checklist


estherk15 (approved):

api/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -806,6 +806,151 @@ type KubeStateMetricsCoreFeatureConfig struct {
 	// This must point to a ConfigMap containing a valid cluster check configuration.
 	// +optional
 	Conf *CustomConfig `json:"conf,omitempty"`
+
+	// `CollectCustomResources` defines custom resources for the kube-state-metrics core check to collect.
```
fanny-jiang: Can we add config instructions for `[]resource` in this description so that it'll be populated in the docs?

api/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -806,6 +806,151 @@ type KubeStateMetricsCoreFeatureConfig struct {
...
+	// MetricNamePrefix defines a prefix for all metrics of the resource.
+	// If set to "", no prefix will be added.
+	// Example: If set to "foo", MetricNamePrefix will be "foo_<metric>".
+	// +optional
+	MetricNamePrefix *string `json:"metricNamePrefix" yaml:"metricNamePrefix"`
+
+	// GroupVersionKind of the custom resource to be monitored.
+	GroupVersionKind GroupVersionKind `json:"groupVersionKind" yaml:"groupVersionKind"`
+
+	// Labels are added to all metrics. If the same key is used in a metric, the value from the metric will overwrite the value here.
```
fanny-jiang, line 849: This is a little confusing. Are metric values really overwriting label values here? Or are configured label values overwriting existing label values if a metric already has the same label key?

fanny-jiang: This might be more related to the KSM custom resource metrics collection behavior in the agent, but this is what I'm seeing while testing: 

```
    kubeStateMetricsCore:
      enabled: true
      collectCustomResources:
        - groupVersionKind:
            group: datadoghq.com
            kind: DatadogAgent
            version: v2alpha1
          commonLabels:
            foo: "bar"
            bar: "zip"
            kube_cluster_name: "blahlalala"
            kubernetes_state_customresource.uptodateagents: "lalala"
          metrics:
            - each:
                gauge:
                  path:
                    - status
                    - agent
                    - upToDate
                type: gauge
              help: Number of up-to-date agents
              name: uptodateagents
```

Result: 

```
  kubernetes_state_customresource.uptodateagents             gauge  1756842646  12            kube_cluster_name:fanny-kind, bar:zip, customresource_group:datadoghq.com, customresource_kind:DatadogAgent, customresource_version:v2alpha1, foo:bar, kube_cluster_name:blahlalala, orch_cluster_id:a55c312b-e4e3-45cd-b8ab-2c1ee95824b7, kubernetes_state_customresource.uptodateagents:lalala
```

When I configure an existing label key with a new value, the label value does not get overwritten. See the duplicate labels, `kube_cluster_name:fanny-kind` and `kube_cluster_name:blahlalala`

When I add a new label key with the same name as the metric name, this new label also gets added to the metric, but with the configured label value: `kubernetes_state_customresource.uptodateagents:lalala`. 

It would be good if we can clarify the expected behavior in the description.

api/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -806,6 +806,151 @@ type KubeStateMetricsCoreFeatureConfig struct {
 	// This must point to a ConfigMap containing a valid cluster check configuration.
 	// +optional
 	Conf *CustomConfig `json:"conf,omitempty"`
+
+	// `CollectCustomResources` defines custom resources for the kube-state-metrics core check to collect.
+	// +optional
+	// +listType=atomic
+	CollectCustomResources []Resource `json:"collectCustomResources,omitempty"`
```
levan-m: Since we are investing in simplifing migration could we align configs as closely as possible with helm?
https://github.com/DataDog/helm-charts/blob/ff7cf3fbee4c440bc493a105fae27a2e2c1c3958/charts/datadog/values.yaml#L178-L185

api/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -806,6 +806,151 @@ type KubeStateMetricsCoreFeatureConfig struct {
...
+	// +optional
+	ResourcePlural string `json:"resourcePlural" yaml:"resourcePlural"`
+}
+
+// GroupVersionKind is the Kubernetes group, version, and kind of a resource.
+type GroupVersionKind struct {
+	Group   string `json:"group"   yaml:"group"`
+	Version string `json:"version" yaml:"version"`
+	Kind    string `json:"kind"    yaml:"kind"`
+}
```
lavigne958, line 865: could this struct be imported from kube source code ?
it will _never_ change but for simplicity ?

internal/controller/datadogagent/feature/kubernetesstatecore/rbac.go
```diff
@@ -129,5 +134,36 @@ func getRBACPolicyRules(collectorOpts collectorOptions) []rbacv1.PolicyRule {
...
+	if len(collectorOpts.customResources) > 0 {
+		// Group custom resources by API group using sets to avoid duplicates
+		groupedResources := make(map[string]map[string]struct{})
+		for _, cr := range collectorOpts.customResources {
+			apiGroup := cr.GroupVersionKind.Group
+			// Use the resource plural if specified, otherwise derive it from the Kind
+			resourceName := cr.ResourcePlural
+			if resourceName == "" {
+				resourceName = strings.ToLower(flect.Pluralize(cr.GroupVersionKind.Kind))
+			}
```
lavigne958, line 147: if we force to lower the plural name we build here, should we force the plural name we read from the custom resource too ? is it always in lowercase by rune from kubernetes or something ?

api/datadoghq/v2alpha1/datadogagent_types.go
L3n41c: Here is the definition of `GroupVersionKind` in `apimachinery`:
https://github.com/kubernetes/apimachinery/blob/268a6d0fb19c92c9665e8f5cd85b564557038dc1/pkg/apis/meta/v1/group_version.go#L82-L90.
Unfortunately, it misses the `yaml` tags that we need here to embed a `GroupVersionKind` in a custom resource.

internal/controller/datadogagent/feature/kubernetesstatecore/rbac.go
L3n41c: This comes from the core Kubernetes concepts.

The **GVK** is what we find inside resource manifests. The **kind** is the object type. It’s capitalized and singular.
For ex., a resource contains:
```yaml
apiVersion: apps/v1
kind: Deployment
```

The **GVR** is what is used in RESTful API path. The **resource** is lower cased and plural.
For ex., the API to operate on the above resource is:
```
/apis/apps/v1/namespaces/{ns}/deployments
```

The logic here is the same as the one implemented in `kube-state-metrics` upstream: https://github.com/kubernetes/kube-state-metrics/blob/e332175940b8d8f76648c24d79c1e0d59a9b7926/pkg/customresourcestate/config.go#L82-L90

api/datadoghq/v2alpha1/datadogagent_types.go
L3n41c: In fact, all this configuration is passed verbatim to upstream `kube-state-metrics` which parses it.
That’s why this whole structure has been copied, with the comments, from upstream `kube-state-metrics` as-is:
https://github.com/kubernetes/kube-state-metrics/blob/e332175940b8d8f76648c24d79c1e0d59a9b7926/pkg/customresourcestate/config.go#L53

I’d rather not try to over-interpret `kube-state-metrics` spec and rather reference the upstream doc.

api/datadoghq/v2alpha1/datadogagent_types.go
L3n41c: I’ve renamed the field to make it match the Helm configuration in https://github.com/DataDog/datadog-operator/pull/1927/commits/28c60bdd738bb1932720260ae2902c6319e26f53.

api/datadoghq/v2alpha1/datadogagent_types.go
L3n41c: I’ve added a link to the documentation of the details of the fields of this structure here: https://github.com/DataDog/datadog-operator/commit/28c60bdd738bb1932720260ae2902c6319e26f53#diff-bb61879086cb5d8e4bd2d033770ca5943d58d914262b1c1f6cab53ad5a1b263a.

PR #1960, opened 2025-06-06
Use DDA-generated ConfigMap to store APM telemetry info

tbavelier:
### What does this PR do?

* Uses a ConfigMap (like with Helm) to store APM Telemetry KPI information
* For DDAI, manage this ConfigMap by the DDA controller

### Motivation

* At the moment, these env vars are generated dynamically based on the DDA/DDAI UID -> this causes the spec of the daemonset/deploy to change when enabling DDAI flag as they are re-generated. On top of sending false signals, the spec change causes a re-deployment that shouldn't be necessary. By using a ConfigMap to store the information, we ensure it remains stable (instead of being re-generated, refer to [CECO-1087](https://datadoghq.atlassian.net/browse/CECO-1087)) based on the DDA whether or not DDAI is enabled, and the DaemonSet / Deployment spec is un-changed, thus preventing the restart
image

### Additional Notes

* The ConfigMap name is hard-coded to `datadog-apm-telemetry-kpi` instead of being dynamic based on the DDA name: this is because when profiles are enabled, DDAIs will be generated with a name different from the DDA that will create additional configmaps, and thus DaemonSet spec changes compared to when DDAI is disabled
    * We could instead imagine inside the function to retrieve/generate the ConfigMap name to remove the `with-profile` and retrieve back the DDA name

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Checklist

[CECO-1087]: https://datadoghq.atlassian.net/browse/CECO-1087?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/controller/datadogagent/component/agent/default.go
```diff
@@ -606,16 +605,37 @@ func envVarsForCoreAgent(dda metav1.Object) []corev1.EnvVar {
 func envVarsForTraceAgent(dda metav1.Object) []corev1.EnvVar {
 	envs := []corev1.EnvVar{
 		{
-			Name:  common.DDAPMInstrumentationInstallId,
-			Value: utils.GetDatadogAgentResourceUID(dda),
+			Name: common.DDAPMInstrumentationInstallId,
+			ValueFrom: &corev1.EnvVarSource{
+				ConfigMapKeyRef: &corev1.ConfigMapKeySelector{
+					LocalObjectReference: corev1.LocalObjectReference{
+						Name: "datadog-apm-telemetry-kpi",
```
khewonc: Since the cm name and keys are used multiple times in different places, can we put them in [constants](https://github.com/DataDog/datadog-operator/blob/main/pkg/constants/const.go) and reference from there?

tbavelier: Done in 4b2b04f1d62bfd06769b08282ba0a60ca8c0a1f3

PR #1983, opened 2025-06-17
New OTEL Image

alessandro-avendano:
### What does this PR do?

changed the otel image default that is being used, now includes the -full suffix

### Motivation

Otel image was outdated, using beta tag

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

On a local kind cluster, have spec.features.otelCollector.enabled set to true in the DDA manifest. Then describe an agent, and go under Containers.otel-agent and make sure that the image is gcr.io/datadoghq/agent:7.66.1-full (whatever agent version you use, just make sure it has "-full" as the suffix).

### Checklist

### DIFF
--- a/internal/controller/datadogagentinternal/component/agent/default.go
+++ b/internal/controller/datadogagentinternal/component/agent/default.go
@@ -334,9 +334,7 @@ func agentImage() string {
 }
 
 func otelAgentImage() string {
-	// todo(mackjmr): Update once OTel agent is GA (7.64.0), as the ot-beta tag will be discontinued.
-	return fmt.Sprintf("%s/%s:%s", images.DefaultImageRegistry, images.DefaultAgentImageName, images.OTelAgentBetaTag)
-
+	return fmt.Sprintf("%s/%s:%s%s", images.DefaultImageRegistry, images.DefaultAgentImageName, images.AgentLatestVersion, images.FullTagSuffix)
 }
 
 func initContainers(dda metav1.Object, requiredContainers []apicommon.AgentContainerName) []corev1.Container {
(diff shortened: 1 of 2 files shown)


internal/controller/datadogagentinternal/component/agent/default.go
```diff
@@ -334,9 +334,7 @@ func agentImage() string {
 }
 
 func otelAgentImage() string {
```
levan-m: This [PR](https://github.com/DataDog/datadog-operator/pull/1916/files#diff-99f3b0c011db2b1425d672097cb8258747a30d66cf70e79275681ffb992bbcd4)  updated this function in DatadogAgent controller in a bit different way so better to follow same approach.

alessandro-avendano: change has been made to stay consistent with the previous PR

internal/controller/datadogagentinternal/component/agent/default.go
```diff
@@ -439,7 +437,7 @@ func processAgentContainer(dda metav1.Object) corev1.Container {
 func otelAgentContainer(_ metav1.Object) corev1.Container {
 	return corev1.Container{
 		Name:  string(apicommon.OtelAgent),
-		Image: otelAgentImage(),
+		Image: fullAgentImage(),
```
AliDatadog, line 440: if otel is enabled, all containers should use the full image. However, since the `ddot-collector` is already released with 7.67, shouldn't we just use `ddot-collector` ?

levan-m: @AliDatadog we planned to switch this in 1.17. Any reason to do it now?

PR #1984, opened 2025-06-17
[AGENTONB-2268] Add control plane monitoring for Openshift >4.14 and EKS

swang392:
### What does this PR do?

Adds openshift control plane configuration with introspection [CECO-2268]
- Adds openshift and AKS label detection 436595e8f06f5c7a0b1b2c65b8f7cf172c5e4ee1
- Adds a new feature, `controlPlaneMonitoring`, which is enabled by default and sets up endpoint checks for compatible providers

### Motivation

Currently, configuring control plane monitoring requires manual setup by the user, with platform-specific instructions. 

### Additional Notes

- some fixes to the DCA reconcile loop have been made in https://github.com/DataDog/datadog-operator/pull/2031, which expands on this branch
- depends on https://github.com/DataDog/datadog-agent/pull/38218 
- A seperate PR adding control plane monitoring for EKS has been merged into this PR. Specific testing instructions for EKS are here https://github.com/DataDog/datadog-operator/pull/2031

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: v7.69 (if testing before release, use the image `registry.ddbuild.io/datadog-cluster-agent:dev-tbavelier-resolve-ip-endpoint-file-cce558be` or `7.69.0-rc.1`)

### Describe your test plan

**To verify feature works:**
- update bundles `make VERSION=0.0.1-dev bundle`
- start up a openshift cluster ([instructions](https://datadoghq.atlassian.net/wiki/spaces/TS/pages/2520056015/Creating+OpenShift+Environment+With+Code-Ready-Containers))
- Build a custom image of the operator and push it to docker 
- in `bundle/manifests/datadog-operator.clusterserviceversion.yaml`:
     - enable control plane introspection in `spec.containers.args`
     - Add custom operator image built in the previous step in `spec.containers.image`
image

- build the bundle image and install onto openshift cluster: `make bundle-build-push BUNDLE_IMG=sarahwang579/operator-bundle:TAG` and `operator-sdk run bundle docker.io/sarahwang579/operator-bundle:TAG`
- apply a DDA config enabling cluster checks runner. Control Plane Monitoring is already enabled by default. For example
```
kind: DatadogAgent
metadata:
  name: datadog
spec:
  global:
    clusterName: sarah-openshift
    credentials:
      apiSecret:
        secretName: datadog-secret
        keyName: api-key
    kubelet:
      tlsVerify: false
  features:
    clusterChecks:
      enabled: true
      useClusterChecksRunners: true
```
- Copy this secret to allow mounting TLS certs: 
```
oc get secret etcd-metric-client -n openshift-etcd-operator -o yaml | sed 's/name: etcd-metric-client/name: etcd-metric-client/' | sed 's/namespace: openshift-etcd-operator/namespace: default/' | oc create -f -
```
- Exec into the cluster agent pod and run `agent clusterchecks`. 
- Verify that you see the following 4 checks:
     - `kube_apiserver_metrics`
     - `etcd`
     - `kube_controller_manager`
     - `kube_scheduler`
- Verify that metrics are visible on [Datadog](https://dddev.datadoghq.com/metric/explorer?fromUser=false&state=H4sIAAAAAAAAE12R3WrDMAyF30WXw5S669o1rzKKMbGcCvyTWk4hDXn3oaTNYHfS4fPRkTzBAwtTTtDAYX84Gq3NXoOCrtj-xtD8TODQU6K6QBNUqgGhAVBraZie0uvTpthAnRgG9FXEsRegUkTGQsigoOB9QK7rgILc58RofC7R1v_sfVgrQZON4iXSKDGdrdZwHkorcsRaqH2_GaEB--gaHrli3LX9sBsYy_Qxw3xVIMOGYFffV_NnPV_n6ywY94GqcRQxyZ0E39Q2J0_dYhAoUoVG7xVwLlUulYvDAg045BYWL1nLl2WFCbha4fT5S39ezqfD8VufFGBym3Z5aQV9Qb6ZmJ1syYEcpQ4U9HZgdNB4GxhnBZ5CxfKKuEU2D3qa9yfkPhBXmH8B94EUm_kBAAA&start=1751397624816&end=1751397924816&paused=false#N4Ig7glgJg5gpgFxALlAGwIYE8D2BXJVEADxQEYAaELcqyKBAC1pEbghkcLIF8qo4AMwgA7CAgg4RKUAiwAHOChASAtnADOcAE4RNIKtrgBHPJoQaUAbVBGN8qVoD6gnNtUZCKiOq279VKY6epbINiAiGOrKQdpYZAYgUJ4YThr42gDGSsgg6gi6mZaBZnHKGniqyJlsmQDWTtp4IgB0dXgARnBOGPIQfgBuOi3y2jj5bHgaLWwYaEwtOHXA7V1OmWhTCDpOkerIGhjaGIwAtDiKIhqMEIIIPC0YGuv4IggAFACUIHygezkgWJYABMiWSCFS6TwWQB+UKxUBpRouQqVRqcHqjWaLUQmSgIzGEzgUxmcDmCyWK063Q2Wx2-wORxO50u11u90ez0yrw+31+ESiAKBAGYwSk0hlsso4RAiokgeVKtVag0mq1Vt0NOioHg0MNRuNEJNprN5oxFssNetNhpttpdoLGcczhc4FcbncHk8Xs1eT8KH9BTEkQAWMUQiXQqW5GVyko6ZEgVHKjGq7FW7lvMZoPX2jyReDaAmGpjEk1ks0WqlrWm2+mOw7Ollutmezk+t5fH4AXSorncuqe1lA-dUg+DCYS-NH49yQNB07cY8wE7iosXA5Xc9DPaotqwepkIHkc0Q22UUBwMGtTw0ssSaFE3TkinKOEfUAfT6c9CYyhES5zP6ST9PImBYE4L4Ao+IhKDw3Y8HwSZgeIADCUjCDAKAiLqaA8EAA), replace cluster name with your cluster name

**To verify no changes to default**
- create a kind cluster 
- Deploy the operator with the `introspectionEnabled` flag added and deploy a minimal DDA config. `controlPlaneMonitoring` is enabled by default.
- Verify no error logs occur

### Checklist

[CECO-2268]: https://datadoghq.atlassian.net/browse/CECO-2268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/controller/datadogagent/feature/controlplaneconfiguration/configmap.go
```diff
@@ -0,0 +1,91 @@
...
+	if provider == kubernetes.DefaultProvider {
+		configMap = &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      configMapName,
+				Namespace: f.owner.GetNamespace(),
+			},
+			Data: map[string]string{
+				"foo.yaml": "bar",
+			},
+		}
```
tbavelier: This should be removed when un-drafted

internal/controller/datadogagent/feature/controlplaneconfiguration/feature.go
```diff
@@ -0,0 +1,212 @@
...
+// ManageNodeAgent allows a feature to configure the Node Agent's corev1.PodTemplateSpec
+// It should do nothing if the feature doesn't need to configure it.
+func (f *controlPlaneConfigurationFeature) ManageNodeAgent(managers feature.PodTemplateManagers, provider string) error {
+	f.provider = provider
+	return nil
+}
+
+// ManageClusterChecksRunner allows a feature to configure the ClusterChecksRunner's corev1.PodTemplateSpec
+func (f *controlPlaneConfigurationFeature) ManageClusterChecksRunner(managers feature.PodTemplateManagers) error {
+	if f.provider != kubernetes.OpenshiftRHCOSType {
```
tbavelier: This confuses me, shouldn't we add the volume if the provider is OpenShift ?

internal/controller/datadogagent/feature/controlplaneconfiguration/configmap.go
```diff
@@ -0,0 +1,91 @@
...
+
+				"etcd.yaml": `advanced_ad_identifiers:
+  - kube_endpoints:
+      name: "etcd"
+      namespace: "openshift-etcd"
+cluster_check: true
+init_config: {}
+instances:
+  - prometheus_url: "https://%%host%%:%%port%%/metrics"
+    tls_ca_cert: "/etc/etcd-certs/etcd-client-ca.crt"
```
tbavelier: I think our doc might be wrong/missing a detail: the client ca certificate is not present in the secret we ask to copy over. We should either remove `tls_ca_cert` and instead use `ssl_verify: false` to trust the CA, or get the CA certificate from the ConfigMap `etcd-metrics-ca-bundle` which has `ca-bundle.crt` that would fit I believe (to be tested). We'll still need in either case the client certificates from the secret which are correctly configured

tbavelier (commented):
didn't review the code, simply some observations from testing

internal/controller/datadogagent/feature/controlplaneconfiguration/feature.go
swang392: yep, i was selecting lines to commit after testing some secrets stuff and missed this

internal/controller/datadogagent/feature/controlplaneconfiguration/configmap.go
swang392: Would `ssl_verify: false` be necessary to add if we remove `tls_ca_cert`? If I add the annotations with this reference to the client ca certificate and without `ssl_verify: false`, I see the etcd check working. `

tbavelier: Well it's either we provide the CA certificate then we do not need `ssl_verify: false` since the CA cert will be used to trust. Or we don't provide it, and we do need `ssl_verify` as the endpoint is only behind https, not http and the CA is not known to the Agent I think ? At least, when I was using `curl`, I had to use `-k` (ssl_verify: false) or provide the CA certificate.
The `tls_key` and `tls_cert` are needed in both scenarios for authenticartion.

internal/controller/datadogagent/feature/controlplaneconfiguration/configmap.go
swang392: addressed in [4be5e82](https://github.com/DataDog/datadog-operator/pull/1984/commits/4be5e82a76c5f4634bba00bb5d7d60af12f920d0)

internal/controller/datadogagent/feature/controlplaneconfiguration/configmap.go
swang392: addressed in [d271842](https://github.com/DataDog/datadog-operator/pull/1984/commits/d27184293e4985883bed8b4074ae38f3fe3c159f), the secrets are named `tls.crt` and `tls.key`, removed `tls_ca_cert` as its not copied in the secret or needed

internal/controller/datadogagent/controller_reconcile_dca.go
```diff
@@ -30,66 +30,103 @@ import (
...
 	}
-	if len(featErrors) > 0 {
-		err := utilerrors.NewAggregate(featErrors)
-		updateStatusV2WithClusterAgent(deployment, newStatus, now, metav1.ConditionFalse, "ClusterAgent feature error", err.Error())
-		return result, err
-	}
+	// TODO: does it make sense to put provider list BEFORE reconciler? this is where the real reconcilation starts
+	// Reconcile cluster agent for each provider
+	var errs []error
+	for provider := range providerList {
```
levan-m: Would this create multiple DCA deployments if we have two providers (like cos and ubuntu)? 

Can we assume that only one provider will affect it's config, given DCA is responsible for cluster-level resources/telemetry and technically we can't have two DCA deployments.

internal/controller/datadogagent/controller_reconcile_v2_helpers.go
```diff
@@ -94,15 +94,17 @@ func (r *Reconciler) reconcileAgentProfiles(ctx context.Context, logger logr.Log
...
 	metrics.DAPEnabled.Set(metrics.FalseValue)
-
+	logger.Info("reconcileAgentProfiles", "DatadogAgentProfileEnabled", r.options.DatadogAgentProfileEnabled, "IntrospectionEnabled", r.options.IntrospectionEnabled)
 	// If profiles or introspection is enabled, get the node list and update providers.
 	if r.options.DatadogAgentProfileEnabled || r.options.IntrospectionEnabled {
 		nodeList, err := r.getNodeList(ctx)
 		if err != nil {
 			return reconcile.Result{}, err
 		}
+		logger.Info("nodeList", "nodeList", nodeList)
```
levan-m: we will need to remove logs or change level before merging.

internal/controller/datadogagent/controller_reconcile_dca.go
swang392: fixed in EKS PR based off this branch https://github.com/DataDog/datadog-operator/pull/2031 https://github.com/DataDog/datadog-operator/pull/2031/commits/39d8c27a8195047f410b4a045fadfdd2de09b93b

internal/controller/datadogagent/controller_reconcile_v2_helpers.go
swang392: fixed in EKS PR #2031 https://github.com/DataDog/datadog-operator/pull/2031/commits/a75a591564761f4cc6499a0452757e79709629ef and https://github.com/DataDog/datadog-operator/pull/2031/commits/75319bb6d937b04fe827eeb350f571f4269d2c79

internal/controller/datadogagent/feature/controlplanemonitoring/configmap_test.go
```diff
@@ -0,0 +1,198 @@
...
+				enabled:  true,
+			},
+			configMapName: otherConfigMapName,
+			want: &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      otherConfigMapName,
+					Namespace: "foo",
+				},
+				Data: map[string]string{
+					"test.yaml": "test",
```
levan-m: Why would we create any CM for unknown providers?
Unless we (remember and) add a conditional in this block, we will create test CM for all new providers, probably not desirable.

internal/controller/datadogagent/feature/controlplanemonitoring/feature.go
```diff
@@ -0,0 +1,206 @@
...
+		return fmt.Errorf("failed to build openshift configmap: %w", err2)
+	}
+	openshiftConfigMap.Name = f.openshiftConfigMapName
+
+	if err := managers.Store().AddOrUpdate(kubernetes.ConfigMapKind, openshiftConfigMap); err != nil {
+		return fmt.Errorf("failed to add openshift configmap to store: %w", err)
+	}
+
+	// EKS ConfigMap
+	eksConfigMap, err2 := f.buildControlPlaneMonitoringConfigMap(kubernetes.EKSProviderLabel, f.eksConfigMapName)
```
levan-m: This creates second configMaps when it's not needed. Initially I thought this was just due to merge of EKS PR into this, but without access to `provider` not sure how we can determine which CM to create.

internal/controller/datadogagent/feature/controlplanemonitoring/const.go
```diff
@@ -0,0 +1,21 @@
...
+package controlplanemonitoring
+
+const (
+	openshiftConfigMapName = "datadog-controlplane-configuration-openshift"
+	defaultConfigMapName   = "datadog-controlplane-configuration-default"
+	eksConfigMapName       = "datadog-controlplane-configuration-eks"
+	otherConfigMapName     = "datadog-controlplane-configuration-unknown"
+
+	controlPlaneMonitoringVolumeName      = "controlplane-config"
+	controlPlaneMonitoringVolumeMountPath = "/etc/datadog-agent/conf.d"
```
levan-m, line 14: Should we mount the config map as a subfolder similar to KSM, OE ones?

internal/controller/datadogagent/feature/controlplanemonitoring/feature.go
```diff
@@ -0,0 +1,206 @@
...
+
+	// Select the appropriate configmap based on provider
+	var configMapName string
+	if providerLabel == kubernetes.OpenShiftProviderLabel {
+		configMapName = f.openshiftConfigMapName
+
+	} else if providerLabel == kubernetes.EKSProviderLabel {
+		configMapName = f.eksConfigMapName
+	} else {
+		configMapName = f.defaultConfigMapName
```
levan-m, line 140: nit - you can perhaps return here and avoid if check later.

internal/controller/datadogagent/feature/controlplanemonitoring/feature.go
```diff
@@ -0,0 +1,206 @@
...
+func (f *controlPlaneMonitoringFeature) ManageNodeAgent(managers feature.PodTemplateManagers, provider string) error {
+	f.provider = provider
+	return nil
+}
+
+// ManageClusterChecksRunner allows a feature to configure the ClusterChecksRunner's corev1.PodTemplateSpec
+func (f *controlPlaneMonitoringFeature) ManageClusterChecksRunner(managers feature.PodTemplateManagers) error {
+	providerLabel, _ := kubernetes.GetProviderLabelKeyValue(f.provider)
+	if providerLabel == kubernetes.OpenShiftProviderLabel {
+		etcdCertsVolume := &corev1.Volume{
```
levan-m, line 224: Why are we mounting this on CLC and not on DCA and creating config map mount in DCA and not here?

internal/controller/datadogagent/controller_reconcile_dca.go
```diff
@@ -89,7 +127,26 @@ func (r *Reconciler) reconcileV2ClusterAgent(logger logr.Logger, requiredCompone
...
 	}
 
-	return r.createOrUpdateDeployment(deploymentLogger, dda, deployment, newStatus, updateStatusV2WithClusterAgent)
+	// Add provider label to deployment
+	if deployment.Labels == nil {
+		deployment.Labels = make(map[string]string)
+	}
+	deployment.Labels[constants.MD5AgentDeploymentProviderLabelKey] = dcaProvider
+
+	res, err := r.createOrUpdateDeployment(deploymentLogger, dda, deployment, newStatus, updateStatusV2WithClusterAgent)
```
levan-m: Do we want to add this label even if introspection is disable?

internal/controller/datadogagent/feature/admissioncontroller/feature.go
```diff
@@ -323,7 +323,7 @@ func (f *admissionControllerFeature) ManageDependencies(managers feature.Resourc
 	return nil
 }
 
-func (f *admissionControllerFeature) ManageClusterAgent(managers feature.PodTemplateManagers) error {
+func (f *admissionControllerFeature) ManageClusterAgent(managers feature.PodTemplateManagers, provider string) error {
```
clamoriniere, line 331: Some suggestions regarding the function signature change:

I recommend using a struct instead of a plain string to represent the provider information. The main reason is future extensibility — as introspection evolves, we may want to associate additional metadata with a provider, such as Linux distribution, GPU availability, or other environment-specific details.

For example, we could introduce a struct like this:

```go
type IntrospectionInfo struct {
  Provider string
  // Additional fields can be added in the future as needed
}
```

This approach would make the interface more future-proof and easier to extend without breaking changes.

Additionally, if you look at the existing Helm chart configuration for [providers](https://github.com/DataDog/helm-charts/blob/3425916e07ec0f345cab28adb4b876c9de0c5f05/charts/datadog/values.yaml#L2592), you’ll see that we already support several sub-options for a given provider. This reinforces the idea that a more structured representation would align better with our current and future configuration patterns.

internal/controller/datadogagent/feature/controlplanemonitoring/configmap_test.go
swang392: Yeah, I had that just for the sake of testing an unknown provider but it's not necessary and can be removed if we don't think it's useful. A ConfigMap corresponding to an unknown provider would never get created because `buildControlPlaneMonitoringConfigMap` takes in the provider

internal/controller/datadogagent/feature/controlplanemonitoring/feature.go
swang392: The `etcd` certs are needed by the CLC, which is why I added the volume and volume mount here. But the etcd annotation configmaps should not be mounted to the CLC, since the endpoint check is still being dispatched by the DCA.

internal/controller/datadogagent/feature/controlplanemonitoring/feature.go
swang392: fixed in [9269552](https://github.com/DataDog/datadog-operator/pull/1984/commits/926955244d090be952d5c6e134e4d06e39590c9e)

internal/controller/datadogagent/feature/controlplanemonitoring/feature.go
swang392: fixed in [9269552](https://github.com/DataDog/datadog-operator/pull/1984/commits/926955244d090be952d5c6e134e4d06e39590c9e). I kept the same names for configmaps but coudl also change to use one general configmap name with different contents depending on the provider.

internal/controller/datadogagent/controller_reconcile_dca.go
swang392: fixed in [9269552](https://github.com/DataDog/datadog-operator/pull/1984/commits/926955244d090be952d5c6e134e4d06e39590c9e)

internal/controller/datadogagent/feature/controlplanemonitoring/configmap_test.go
swang392: removed in [aff5fcb](https://github.com/DataDog/datadog-operator/pull/1984/commits/aff5fcbe4fb0354bb3e4f296d1a8a63976213686)

internal/controller/datadogagent/feature/controlplanemonitoring/const.go
swang392: mounting to a subfolder leads to the endpoint checks being renamed to the subfolder, I'll keep it as is.

internal/controller/datadogagent/feature/admissioncontroller/feature.go
swang392: Hi Cedric, thanks for the feedback! I'll put this suggestion on hold and make a [card](https://datadoghq.atlassian.net/browse/AGENTONB-2432) for it in our backlog. This change would only be applicable for the current DDA setup, whereas supporting control plane monitoring with DDAI hasn't been implemented yet and is higher priority.

internal/controller/datadogagent/controller_v2_test.go
```diff
@@ -1100,6 +1100,248 @@ func Test_AutopilotOverrides(t *testing.T) {
...
+		return err
+	}
+	assert.Equal(t, 1, len(daemonSetList.Items))
+	assert.Equal(t, expectedName, daemonSetList.Items[0].ObjectMeta.Name)
+
+	cms := corev1.ConfigMapList{}
+	if err := c.List(context.TODO(), &cms, client.InNamespace(resourcesNamespace)); err != nil {
+		return err
+	}
+	cpCm := corev1.ConfigMap{}
```
levan-m, line 1310: should we assert anything on cpCm?

internal/controller/datadogagent/controller_v2_test.go
```diff
@@ -1100,6 +1100,248 @@ func Test_AutopilotOverrides(t *testing.T) {
...
+						Labels: map[string]string{
+							kubernetes.OpenShiftProviderLabel: "rhel",
+						},
+					},
+				},
+			},
+			want:    reconcile.Result{RequeueAfter: defaultRequeueDuration},
+			wantErr: false,
+			wantFunc: func(t *testing.T, c client.Client) error {
+				return verifyDCADeployment(t, c, resourcesName, resourcesNamespace, dcaName, "openshift")
```
levan-m, line 1160: Would be nice to assert etcd cert mount in DS as well; could be a follow-up PR.

internal/controller/datadogagent/controller_v2_test.go
swang392: We could but I think that would be redundant testing, since we already verify in `internal/controller/datadogagent/feature/controlplanemonitoring/configmap_test.go` that the contents of the configmap is accurate corresponding to the provider. Here we just assert that the correct configmap is created.

jennchenn (approved):

PR #1999, opened 2025-06-20
add metadata forwarder

celenechang:
### What does this PR do?

Adds metadata forwarder in the Operator.

This runs outside of the controllers, but can fallback to grab DatadogAgent credentials if they are not set in the Operator deployment.

### Motivation

Add DatadogOperator metadata in REDAPL
 
### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

- [ ] works with operator credentials, cluster name set
- [ ] works with operator credentials, changing one of the flags (genericresource)
- [ ] works with operator credentials, cluster name not set
- [ ] works with datadogagent credentials, no operator credentials
- [ ] check behavior if operator credentials are set and there is no DDA
- [ ] check behavior if operator credentials are not set and there is no DDA

### Checklist


pkg/controller/utils/metadata/metadata_forwarder.go
```diff
@@ -0,0 +1,404 @@
...
+)
+
+const (
+	apiHTTPHeaderKey       = "Dd-Api-Key"
+	userAgentHTTPHeaderKey = "User-Agent"
+	contentTypeHeaderKey   = "Content-Type"
+	acceptHeaderKey        = "Accept"
+
+	defaultURLScheme     = "https"
+	defaultURLHost       = "app.datadog.com"
```
levan-m:
```suggestion
	defaultURLHost       = "app.datadoghq.com"
```

pkg/controller/utils/metadata/metadata_forwarder.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+
+	return mdf
+}
+
+// Start starts the metadata forwarder
+func (mdf *MetadataForwarder) Start() {
+	err := mdf.setCredentials()
+	if err != nil {
+		mdf.logger.Error(err, "Could not set credentials; not starting metadata forwarder")
```
levan-m, line 122: In the test case 
> works with datadogagent credentials, no operator credentials

I get two errors with DDA approach:
`DatadogAgent not found` if Operator starts without DDA in the clusters. 
`the cache is not started, can not read objects` DDA, secret are already there and Operator is restarted. Not sure how to make it work with DDA. 

Either way this will attempt loading only once at startup and never retry, which is probably fine since pod lifecycle is a lot shorter than DDA/secret lifecycle so we at least need to address second error.

cmd/main.go
```diff
@@ -328,20 +337,63 @@ func run(opts *options) error {
...
+
 	// +kubebuilder:scaffold:builder
 
 	setupLog.Info("starting manager")
 	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
 		return setupErrorf(setupLog, err, "Problem running manager")
 	}
 
+	// Stop metadata forwarder
+	mdf.Stop()
```
levan-m: Did we move the above block to be able to have `mdf` initialized and stopped in main? version info handling seems a bit low level stuff for main, I'm also wondering if we really need to call Stop() here.

pkg/controller/utils/metadata/metadata_forwarder.go
```diff
@@ -0,0 +1,404 @@
...
+	reader := bytes.NewReader(mdf.getPayload())
+	req, err := http.NewRequestWithContext(context.TODO(), "POST", mdf.requestURL, reader)
+	if err != nil {
+		mdf.logger.Error(err, "Error creating request", "url", mdf.requestURL, "reader", reader)
+	}
+	req.Header = mdf.payloadHeader
+
+	resp, err := mdf.client.Do(req)
+	if err != nil {
+		return fmt.Errorf("Error sending request: %w", err)
```
levan-m:
```suggestion
		return fmt.Errorf("error sending request: %w", err)
```

pkg/controller/utils/metadata/metadata_forwarder.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+
+	defer resp.Body.Close()
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Errorf("Failed to read response body: %w", err)
```
levan-m:
```suggestion
		return fmt.Errorf("failed to read response body: %w", err)
```

pkg/controller/utils/metadata/metadata_forwarder.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+
+	mdf.logger.V(1).Info("Read response", "status code", resp.StatusCode, "body", string(body))
+	return nil
+}
+
+func (mdf *MetadataForwarder) setupFromOperator() error {
+	if mdf.credsManager == nil {
+		return fmt.Errorf("Credentials Manager is undefined")
```
levan-m:
```suggestion
		return fmt.Errorf("credentials Manager is undefined")
```

cmd/main.go
celenechang: - Since the process is intended to be running for the life of the Operator, and there's no need for a "graceful" shut down of the process, I agree that a `stop` is not necessary
- The metadata forwarder setup was intentionally placed in `main.go` to indicate its status as unaffiliated with any controllers; i.e. even if no controllers are enabled, the metadata forwarder should work (assuming creds are present). Unfortunately this does lead to the necessity of setting up the DiscoveryClient here, though. If you think the cons outweigh the pros I can incorporate your suggestion to move the forwarder set up down a level

pkg/controller/utils/metadata/metadata_forwarder.go
celenechang: Were you getting the cache error on your branch off of this one? I haven't been able to reproduce on this branch. (I do think I got a cache-related error when I tried at some point to pass the context from the main manager to get the DDA, but maybe that's not related.)

levan-m: I'm getting when testing this branch.

cmd/main.go
levan-m: * I'm fine not having graceful shutdown but it should be conscious decision. Would there be a any value in trying to submit last payload?
* Haven't looked from that perspective. Given client passed is derived from `mgr` and only thing we need is versionInfo, does it make sense to move that large block elsewhere to have less clutter in main?

celenechang: > but it should be conscious decision

Yes, it is consciously decided on my part. As for submitting one last payload, I don't think it's valuable because the interval is already at 1 minute.

> does it make sense to move that large block elsewhere to have less clutter in main?

I'll work on factoring that out

pkg/controller/utils/metadata/metadata_forwarder.go
```diff
@@ -0,0 +1,392 @@
...
+	}
+
+	mdf.payloadHeader = getHeader(mdf.apiKey)
+
+	mdf.logger.Info("Starting metadata forwarder")
+
+	ticker := time.NewTicker(defaultInterval)
+	go func() {
+		for {
+			select {
```
levan-m: nit, I don't get this message but it seems `select` on a case ticker is redundant.

levan-m (approved):

PR #2003, opened 2025-06-23
[CECO-2029] Add profile label to non-default profile DDAIs and do not create EDS if non-default profile DDAI

tbavelier:
### What does this PR do?

* When creating profile DDAIs, include the label `agent.datadoghq.com/datadogagentprofile` in DDAI meta
* Removes `scheme` argument from `setProfileDDAIMeta` as it is not used
* When reconciling Agent, check if non-default profile label and if yes, use DS instead of EDS if enabled

### Motivation

* Keeping behaviour with DDA controller: we do not create EDS for profiles

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

- [ ] EDS enabled 
    * Create a profile
    * Ensure label is present on DDAI
    * Ensure DS gets created instead of EDS for profile pods, while EDS is present for non-profile/default pods
- [ ] EDS disabled
    * Create a profile
    * Ensure label is present on DDAI

### Checklist

### DIFF
--- a/internal/controller/datadogagent/profile.go
+++ b/internal/controller/datadogagent/profile.go
@@ -13,7 +13,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


tbavelier:
Rebased in https://github.com/DataDog/datadog-operator/pull/2004

PR #2004, opened 2025-06-23
[CECO-2029][DatadogAgentInternal] Add profile label to non-default profile DDAIs and do not create EDS if non-default profile DDAI

tbavelier:
### What does this PR do?

* When creating profile DDAIs, include the label `agent.datadoghq.com/datadogagentprofile` in DDAI meta
* Removes `scheme` argument from `setProfileDDAIMeta` as it is not used
* When reconciling Agent, check if non-default profile label and if yes, use DS instead of EDS if enabled

### Motivation

* Keeping behaviour with DDA controller: we do not create EDS for profiles

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

- [ ] EDS enabled 
    * Create a profile
    * Ensure label is present on DDAI
    * Ensure DS gets created instead of EDS for profile pods, while EDS is present for non-profile/default pods
- [ ] EDS disabled
    * Create a profile
    * Ensure label is present on DDAI

### Checklist

### DIFF
--- a/internal/controller/datadogagent/profile.go
+++ b/internal/controller/datadogagent/profile.go
@@ -13,7 +13,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 
 	v1alpha1 "github.com/DataDog/datadog-operator/api/datadoghq/v1alpha1"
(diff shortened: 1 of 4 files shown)


internal/controller/datadogagent/profile.go
```diff
@@ -77,13 +76,10 @@ func (r *Reconciler) computeProfileMerge(ddai *v1alpha1.DatadogAgentInternal, pr
 	// Copy the original DDAI and apply profile spec to create a fake "DDAI" to merge
 	profileDDAI := ddai.DeepCopy()
 	baseDDAI := ddai.DeepCopy()
-	if !agentprofile.IsDefaultProfile(profile.Namespace, profile.Name) {
-		// Clear owner reference to tie GC to the profile
-		baseDDAI.OwnerReferences = []metav1.OwnerReference{}
-	}
```
tbavelier, line 83: Removed this since we removed tie between profile and DDAIs in https://github.com/DataDog/datadog-operator/pull/2001

internal/controller/datadogagent/profile.go
```diff
@@ -77,13 +76,10 @@ func (r *Reconciler) computeProfileMerge(ddai *v1alpha1.DatadogAgentInternal, pr
...
+
 	// Add profile settings to "fake" DDAI
 	setProfileSpec(profileDDAI, profile)
-	if err := setProfileDDAIMeta(profileDDAI, profile, r.scheme); err != nil {
```
tbavelier, line 86: We are not using `r.scheme`

internal/controller/datadogagent/profile.go
```diff
@@ -137,11 +133,21 @@ func setProfileDDAIAffinity(ddai *v1alpha1.DatadogAgentInternal, profile *v1alph
...
 	// Managed fields
 	ddai.ManagedFields = []metav1.ManagedFieldsEntry{}
+	// Include the profile label in the DDAI metadata only for non-default profiles.
+	// This is used to determine whether or not the EDS should be created (only for default profile).
+	// This could possibly be used for GC of the profile DDAIs too.
+	if !agentprofile.IsDefaultProfile(profile.Namespace, profile.Name) {
+		// This should not happen, as we add the DDA label upon creation of the DDAI.
+		if ddai.Labels == nil {
+			ddai.Labels = make(map[string]string)
+		}
```
tbavelier, line 148: Just in case, but we should always have label (DDA name) from the DDA controller

khewonc (approved):

PR #2011, opened 2025-06-27
[CECO-2029][DatadogAgentInternal] Setup dependencies with real owner while using copy for naming

tbavelier:
### What does this PR do?

* Provides a store option to use a different object than the owner to get labels/annotations, while still setting ownerref to the provided owner
* Uses this option solely in the DDAI controller to use the profile DDAI copy (that's using the DDA name) for labels/annotations while still setting the ownerref to an existing/real object

### Motivation

* Avoid creating dependencies with non-existing object in ownerref causing it to be GCd: this would happen when using profiles and DDAI

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist


tbavelier:
Closing in favour of https://github.com/DataDog/datadog-operator/pull/2014

PR #2014, opened 2025-06-27
Reduce update frequency for dependencies

khewonc:
### What does this PR do?

Reduce updates for DDA/DDAI dependencies:
* ownerref changes when dap is enabled
* ownerref changes for dca service
* equality check fails for external metrics service target port not being defined

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist

### DIFF
--- a/internal/controller/datadogagent/feature/externalmetrics/feature.go
+++ b/internal/controller/datadogagent/feature/externalmetrics/feature.go
@@ -152,9 +152,10 @@ func (f *externalMetricsFeature) ManageDependencies(managers feature.ResourceMan
 	// service
 	emPorts := []corev1.ServicePort{
 		{
-			Protocol: corev1.ProtocolTCP,
-			Port:     f.port,
-			Name:     externalMetricsPortName,
+			Protocol:   corev1.ProtocolTCP,
+			Port:       f.port,
+			TargetPort: intstr.FromInt(int(f.port)),
+			Name:       externalMetricsPortName,
 		},
 	}
 	selector := map[string]string{
--- a/internal/controller/datadogagentinternal/controller_reconcile_v2.go
+++ b/internal/controller/datadogagentinternal/controller_reconcile_v2.go
@@ -20,6 +20,7 @@ import (
 	"github.com/DataDog/datadog-operator/internal/controller/datadogagent/common"
 	"github.com/DataDog/datadog-operator/internal/controller/datadogagent/defaults"
 	"github.com/DataDog/datadog-operator/internal/controller/datadogagent/feature"
+	"github.com/DataDog/datadog-operator/pkg/agentprofile"
 	"github.com/DataDog/datadog-operator/pkg/condition"
 	"github.com/DataDog/datadog-operator/pkg/controller/utils"
 	pkgutils "github.com/DataDog/datadog-operator/pkg/controller/utils/datadog"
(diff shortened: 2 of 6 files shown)


internal/controller/datadogagent/feature/externalmetrics/feature.go
```diff
@@ -152,9 +152,10 @@ func (f *externalMetricsFeature) ManageDependencies(managers feature.ResourceMan
 	// service
 	emPorts := []corev1.ServicePort{
 		{
-			Protocol: corev1.ProtocolTCP,
-			Port:     f.port,
-			Name:     externalMetricsPortName,
+			Protocol:   corev1.ProtocolTCP,
+			Port:       f.port,
+			TargetPort: intstr.FromInt(int(f.port)),
```
tbavelier, line 157: Is this expected (addition of `TargetPort`) ?

internal/controller/datadogagentinternal/controller_reconcile_v2.go
```diff
@@ -64,19 +65,22 @@ func (r *Reconciler) reconcileInstanceV2(ctx context.Context, logger logr.Logger
 	r.updateMetricsForwardersFeatures(instanceCopy, enabledFeatures)
 
 	// 1. Manage dependencies.
-	depsStore, resourceManagers := r.setupDependencies(instanceCopy, logger)
+	// set the original DDAI as the owner of dependencies
+	depsStore, resourceManagers := r.setupDependencies(instance, logger)
```
tbavelier, line 62: We're using `instance` again instead of `instanceCopy` -> the default DDAI instance uses the same name as DDA, should we remove the lines 60 and 61 ?

internal/controller/datadogagentinternal/controller_reconcile_v2.go
```diff
@@ -64,19 +65,22 @@ func (r *Reconciler) reconcileInstanceV2(ctx context.Context, logger logr.Logger
...
-	if err = r.manageGlobalDependencies(logger, instanceCopy, resourceManagers, requiredComponents); err != nil {
-		return r.updateStatusIfNeededV2(logger, instance, newStatus, reconcile.Result{}, err, now)
-	}
-	if err = r.manageFeatureDependencies(logger, enabledFeatures, resourceManagers); err != nil {
-		return r.updateStatusIfNeededV2(logger, instance, newStatus, reconcile.Result{}, err, now)
-	}
-	if err = r.overrideDependencies(logger, resourceManagers, instanceCopy); err != nil {
-		return r.updateStatusIfNeededV2(logger, instance, newStatus, reconcile.Result{}, err, now)
+	// only manage dependencies for default DDAIs
+	if instance.Labels[agentprofile.ProfileLabelKey] == "" {
```
tbavelier:
```suggestion
	if !isDDAILabeledWithProfile(instance) {
```
Let's use the helper that is "safe" (check for nil labels -> that should not happen as we always inject the DDA name with `generateObjMetaFromDDA` but since the function already exists, might as well use it)

internal/controller/datadogagentinternal/controller_reconcile_v2.go
```diff
@@ -109,8 +113,11 @@ func (r *Reconciler) reconcileInstanceV2(ctx context.Context, logger logr.Logger
 	}
 
 	// 4. Apply and cleanup dependencies.
-	if err = r.applyAndCleanupDependencies(ctx, logger, depsStore); err != nil {
-		return r.updateStatusIfNeededV2(logger, instance, newStatus, result, err, now)
+	// only manage dependencies for default DDAIs
+	if instance.Labels[agentprofile.ProfileLabelKey] == "" {
```
tbavelier: Same as above, let's use the helper `isDDAILabeledWithProfile`

internal/controller/datadogagent/feature/externalmetrics/feature.go
khewonc: Yes, there's a note in this section about k8s automatically setting it: https://pkg.go.dev/k8s.io/kubernetes/pkg/apis/core#ServicePort
> If this is not specified, the value of the 'port' field is used

PR #2016, opened 2025-06-30
[CECO-2265] Reduce clusterroles update

tbavelier:
### What does this PR do?

* Ensures stable order when generating rules for non-deterministic CCRs (orchestrator explorer and annotationsAndLabelsAsTags)

### Motivation

* Frequent un-necessary updates for 2 CCRs: orchestrator and annotationsAndLabelsAsTags
* Root cause: non-deterministic generation of the `PolicyRule` slice within these `ClusterRoles`. The code iterated over maps to build the list of rules; because map iteration order is not guaranteed in Go, the resulting slice of rules could have a different order between reconciliations. This caused the deep equality check to fail, triggering an update even when the set of permissions was semantically identical.
* Solution: generate rules in a stable, deterministic order:
    * The approach involves a two-level sort in the functions that generate these rules:
        1. Sort resources within each API group: The list of resources (e.g., pods, services) for a given API group is sorted alphabetically.
        2. Sort the API groups themselves: The PolicyRule objects for each API group are then sorted alphabetically by the group name.

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

* Using either feature, check that before (e.g. using public build) and after this PR, the CCRs remain the same for these 2 features
* Internally, verified the frequency of clusterroles updates has disappeared: https://ddstaging.datadoghq.com/logs?query=service%3Akube-apiserver-audit%20%40usr.id%3A%22system%3Aserviceaccount%3Adatadog-agent%3Adatadog-operator%22%20%40objectRef.resource%3Aclusterroles%20k8s_cluster%3Aabra&agg_m=count&agg_m_source=base&agg_q=%40http.method%2C%40objectRef.name%2C%40http.url_details.path&agg_q_source=base%2Cbase%2Cbase&agg_t=count&analyticsOptions=%5B%22bars%22%2C%22dog_classic%22%2Cnull%2Cnull%2C%22value%22%5D&clustering_pattern_field_path=message&cols=host%2Cservice&messageDisplay=inline&refresh_mode=paused&sort_m=%2C%2C&sort_m_source=%2C%2C&sort_t=%2C%2C&storage=hot&stream_sort=desc&top_n=10%2C10%2C10&top_o=top%2Ctop%2Ctop&viz=stream&x_missing=true%2Ctrue%2Ctrue&from_ts=1748442574401&to_ts=1748443474401&live=false

### Checklist

### DIFF
--- a/internal/controller/datadogagent/component/clusteragent/rbac.go
+++ b/internal/controller/datadogagent/component/clusteragent/rbac.go
@@ -6,6 +6,8 @@
 package clusteragent
 
 import (
+	"sort"
+
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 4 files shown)


tbavelier:
We'll most likely go with the normalized approach instead https://github.com/DataDog/datadog-operator/pull/2020/s

PR #2019, opened 2025-06-30
Delete DAP DDAIs when DAP deleted

khewonc:
### What does this PR do?

Delete DAP DDAIs when a DAP is deleted

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist

### DIFF
--- a/internal/controller/datadogagent/ddai.go
+++ b/internal/controller/datadogagent/ddai.go
@@ -6,12 +6,15 @@
 package datadogagent
 
 import (
+	"context"
+
+	"github.com/go-logr/logr"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 
 	apicommon "github.com/DataDog/datadog-operator/api/datadoghq/common"
-	datadoghqv1alpha1 "github.com/DataDog/datadog-operator/api/datadoghq/v1alpha1"
-	datadoghqv2alpha1 "github.com/DataDog/datadog-operator/api/datadoghq/v2alpha1"
+	"github.com/DataDog/datadog-operator/api/datadoghq/v1alpha1"
+	"github.com/DataDog/datadog-operator/api/datadoghq/v2alpha1"
 	"github.com/DataDog/datadog-operator/internal/controller/datadogagent/global"
 	"github.com/DataDog/datadog-operator/internal/controller/datadogagent/object"
 	"github.com/DataDog/datadog-operator/internal/controller/datadogagent/override"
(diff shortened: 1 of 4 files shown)


internal/controller/datadogagent/ddai.go
```diff
@@ -69,3 +72,28 @@ func getDDAILabels(dda metav1.Object) map[string]string {
...
+		valid := false
+		for _, validDDAI := range validDDAIs {
+			if ddai.Name == validDDAI.Name && ddai.Namespace == validDDAI.Namespace {
+				valid = true
+				break
+			}
+		}
+		if !valid {
+			logger.Info("Deleting unused DDAI", "namespace", ddai.Namespace, "name", ddai.Name)
+			if err := r.client.Delete(ctx, &ddai); err != nil {
```
tbavelier:
```suggestion
func (r *Reconciler) cleanUpUnusedDDAIs(ctx context.Context, validDDAIs []*v1alpha1.DatadogAgentInternal) error {
	// Use a map for efficient lookup of valid DDAIs.
	validDDAIMap := make(map[string]struct{}, len(validDDAIs))
	for _, ddai := range validDDAIs {
		validDDAIMap[client.ObjectKeyFromObject(ddai).String()] = struct{}{}
	}

	// List only metadata of DDAIs to reduce memory footprint and API server load.
	ddaiList := &metav1.PartialObjectMetadataList{}
	ddaiList.SetGroupVersionKind(v1alpha1.GroupVersion.WithKind("DatadogAgentInternal"))
	if err := r.client.List(ctx, ddaiList); err != nil {
		return err
	}

	for i := range ddaiList.Items {
		ddai := &ddaiList.Items[i]
		if _, isValid := validDDAIMap[client.ObjectKeyFromObject(ddai).String()]; !isValid {
			ddaiName := ddai.GetName()
			ddaiNamespace := ddai.GetNamespace()
			r.log.Info("Deleting unused DDAI", "namespace", ddaiNamespace, "name", ddaiName)
			// Create a minimal DDAI object for deletion.
			ddaiToDelete := &v1alpha1.DatadogAgentInternal{
				ObjectMeta: metav1.ObjectMeta{
					Name:      ddaiName,
					Namespace: ddaiNamespace,
				},
			}
			if err := r.client.Delete(ctx, ddaiToDelete); err != nil {
```

Offering some possible improvements:
* Using a map to avoid nested loops and instead use direct lookups
* Only retrieve metadata as we only care about name and namespace for our comparison
* Using `r.log` instead of the logger argument since `cleanUpUnusedDDAIs` is a function of the Reconciler anyway
* Loop using the index to access directly the pointer instead of making a copy as `_, item` does

internal/controller/datadogagent/ddai.go
```diff
@@ -6,21 +6,24 @@
 package datadogagent
 
 import (
+	"context"
+
+	"github.com/go-logr/logr"
```
tbavelier:
```suggestion
```
I think we can remove the logger argument from the function, and directly rely on `r.log` since it's a method of the reconciler, see below

tbavelier (approved):

PR #2030, opened 2025-07-03
Alessandro/aks provider

alessandro-avendano:
### What does this PR do?

Adding AKS as a supported provider within the introspection feature, and adds DD_ADMISSION_CONTROLLER_ADD_AKS_SELECTORS env var to be true if the admission controller is enabled and introspection sees that a node is running on AKS. 

### Motivation

Customers do not have to manually add the correct configurations on the Datadog Agent to run on a AKS node, the operator will add the basic configurations needed automatically. 

### Additional Notes

https://docs.datadoghq.com/containers/kubernetes/distributions/?tab=datadogoperator#AKS
Based on previous documentation, we needed to add the admission controller env var AND make kubelet changes in order to run the agent on AKS. Because of some recent certificate changes that AKS made, the only thing the operator needs to adjust for AKS environments is the admission controller env var if the admission controller is enabled. 

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

- Enable introspection feature in the operator (manager.yaml)
- Create AKS cluster with custom operator image
- Once the operator is deployed:
     - Deploy the DDA with admission controller enabled (automatically enabled) and check to see in the cluster agent that the DD_ADMISSION_CONTROLLER_ADD_AKS_SELECTORS env var is set to true
     - Deploy the DDA with admission controller disabled and check to see in the cluster agent that DD_ADMISSION_CONTROLLER_ADD_AKS_SELECTORS is not added as an env var

### Checklist

### DIFF
--- a/internal/controller/datadogagent/controller_reconcile_dca.go
+++ b/internal/controller/datadogagent/controller_reconcile_dca.go
@@ -38,7 +38,7 @@ import (
 func (r *Reconciler) reconcileV2ClusterAgent(ctx context.Context, logger logr.Logger, requiredComponents feature.RequiredComponents, features []feature.Feature, dda *datadoghqv2alpha1.DatadogAgent, resourcesManager feature.ResourceManagers, newStatus *datadoghqv2alpha1.DatadogAgentStatus) (reconcile.Result, error) {
 	var result reconcile.Result
 	now := metav1.NewTime(time.Now())
-  
+
 	// Get provider list for introspection
 	providerList := map[string]struct{}{kubernetes.LegacyProvider: {}}
 	if r.options.IntrospectionEnabled {


internal/controller/datadogagent/controller_reconcile_dca.go
```diff
@@ -30,66 +30,103 @@ import (
...
-
-	// Apply features changes on the Deployment.Spec.Template
-	var featErrors []error
-	for _, feat := range features {
-		if errFeat := feat.ManageClusterAgent(podManagers); errFeat != nil {
-			featErrors = append(featErrors, errFeat)
+	// Get provider list for introspection
+	providerList := map[string]struct{}{kubernetes.LegacyProvider: {}}
+	if r.options.IntrospectionEnabled {
+		nodeList, err := r.getNodeList(ctx)
```
celenechang, line 45: since this is a costly operation, I wonder if it makes sense to make this call up one level and pass it to both this function and `reconcileAgentProfiles` ?

internal/controller/datadogagent/controller_reconcile_dca.go
```diff
@@ -30,66 +30,103 @@ import (
...
-					metav1.NewTime(time.Now()),
-					common.OverrideReconcileConflictConditionType,
-					metav1.ConditionTrue,
-					"OverrideConflict",
-					"ClusterAgent component is set to disabled",
-					true,
-				)
+	// Reconcile cluster agent for each provider
+	var errs []error
+	for provider := range providerList {
```
celenechang, line 55: I don't think we want to loop over all the providers, because we still want one DCA per cluster. (Let me know if I'm misunderstanding something here.) Perhaps we can add some logic to choose the "best" provider - for instance, if the entire cluster is AKS then use AKS. If it's mixed then make a deterministic choice

swang392: I made a fix for this in the EKS control plane monitoring PR https://github.com/DataDog/datadog-operator/pull/2031 (39d8c27a8195047f410b4a045fadfdd2de09b93b)

alessandro-avendano: I will be waiting for Sarah to merge her PR into main, and will then update my branch since the issue has been resolved

PR #2032, opened 2025-07-07
operator now emits INFO messages to stdout

alessandro-avendano:
### What does this PR do?

Operator now emits INFO messages to stdout

### Motivation

Operator was emitting INFO messages to stderr, and stdout seems to be the appropriate place for that.

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

* Create AKS cluster with custom operator image
* Once the operator is deployed, go into Azure Insights within the Azure portal and check to see the source of logs. All INFO logs should have the source be stdout, and all logs that are not INFO logs should have the source be stderr. 

Logs should look like this below, where INFO logs are sent to stdout and the rest of the logs are sent to stderr:
Image 7-11-25 at 9 23 AM

### Checklist


celenechang:
Instead of using zapr, you may be able to keep the controller-runtime zap and add an option for the destination https://github.com/kubernetes-sigs/controller-runtime/blob/1ac370e560b19321af779df0610622c03106ec77/pkg/log/zap/zap.go#L172-L174

Do you mind trying this route and seeing if it works?

alessandro-avendano:
Made the change where instead of using zapr I only used zap and added a core through the `zapOpts` field in the `Options` struct like mentioned in this [issue](https://github.com/kubernetes-sigs/controller-runtime/issues/2242).

PR #2040, opened 2025-07-10
Support custom seccomp profile configData

alessandro-avendano:
### What does this PR do?

Allow use of configdata in custom seccomp. Checksum annotation is also added if configData is used.

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

1. Create cluster with custom operator image (any cluster except kind since it has problems with the system-probe container)
2. Once the operator is deployed, enable a feature so that the system-probe container will be created (like the npm feature)
3. Check: 
  * the default `datadog-system-probe-seccomp` configMap is created
  * no annotations are present on the configMap nor daemonset
  * The default configMap is mounted in the volumes
4. Deploy a DDA with a custom configData (see example below). 
5. Check: 
* The `datadog-system-probe-seccomp` configMap contains the custom data and annotation. 
* The daemonset should contain a matching annotation with the same hash ID. 
6. Update the DDA with a different configData value (i.e. update to `"defaultAction": "SCMP_ACT_LOG"`) 
7. Check: 
* The `datadog-system-probe-seccomp` configMap should contain the updated data and a new annotation hash.
* The daemonset annotation hash should also get updated with the new annotation hash. 
8. Deploy a custom seccomp profile configMap (see below) and update the DDA with the custom configMap
9. Check: 
* The custom configMap should take precedence. Make sure the custom configMap volume is mounted. 
* Check that `datadog-system-probe-seccomp` configmap is not created
* The daemonset does NOT have an annotation

Below is an example of how configData is used in the DDA manifest. 

```yaml
  features:
    apm:
      enabled: true
    logCollection:
      enabled: true
    npm:
      enabled: true
      enableConntrack: true
  override:
    nodeAgent:
      containers:
        system-probe:
          seccompConfig:
            customProfile:
              configData: |
                {
                    "defaultAction": "SCMP_ACT_ALLOW"
                }
#              configMap:
#                name: custom-seccomp-profile
#                items:
#                  - key: system-probe-seccomp.json
#                    path: system-probe-seccomp.json
```

Custom seccomp profile configmap: 
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: custom-seccomp-profile
  namespace: system
data:
  system-probe-seccomp.json: |
    {
      "defaultAction": "SCMP_ACT_LOG"
    }
```

### Checklist

### DIFF
--- a/go.work.sum
+++ b/go.work.sum
@@ -261,7 +261,6 @@ github.com/Azure/azure-sdk-for-go/sdk/keyvault/internal v0.7.1/go.mod h1:9V2j0jn
 github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.3.1 h1:fXPMAmuh0gDuRDey0atC8cXBuKIlqCzCkL8sm1n9Ov0=
 github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.3.1/go.mod h1:SUZc9YRRHfx2+FAQKNDGrssXehqLpxmwRv2mC/5ntj4=
 github.com/Azure/go-autorest v14.2.0+incompatible h1:V5VMDjClD3GiElqLWO7mz2MxNAK/vTfRHdAubSIPRgs=
-github.com/Azure/go-autorest v14.2.0+incompatible/go.mod h1:r+4oMnoxhatjLLJ6zxSWATqVooLgysK6ZNox3g/xq24=
 github.com/Azure/go-autorest/autorest v0.11.20 h1:s8H1PbCZSqg/DH7JMlOz6YMig6htWLNPsjDdlLqCx3M=
 github.com/Azure/go-autorest/autorest v0.11.20/go.mod h1:o3tqFY+QR40VOlk+pV4d77mORO64jOXSgEnPQgLK6JY=
 github.com/Azure/go-autorest/autorest/adal v0.9.13 h1:Mp5hbtOePIzM8pJVRa3YLrWWmZtoxRXqUEzCfJt3+/Q=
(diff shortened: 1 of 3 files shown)


internal/controller/datadogagent/override/container.go
```diff
@@ -217,6 +217,21 @@ func overrideSeccompProfile(containerName apicommon.AgentContainerName, manager
...
+				VolumeSource: corev1.VolumeSource{
+					ConfigMap: &corev1.ConfigMapVolumeSource{
+						LocalObjectReference: corev1.LocalObjectReference{
+							Name: ddaName + "-" + common.SystemProbeAgentSecurityConfigMapSuffixName,
+						},
+					},
+				},
+			}
+			manager.Volume().AddVolume(&vol)
+		}
```
tbavelier:
```suggestion
```
This part is not needed. When using `configData`, you justly use the same default name for the configmap, so the volume is already present in the manager from `GetVolumeForSecurity` in `internal/controller/datadogagent/common/volumes.go`. This allows you to remove the change adding `ddaName` to `overrideSeccompProfile` and `Container`

alessandro-avendano: checksum annotation added whenever configData is used

alessandro-avendano:
Only 2 concerns / questions:
1. When both configMap and configData are used, checksum annotation is still written in the agent ds. Should the annotation only be written when only configData is used on its own, or is it fine how it is now?
2. As of right now if you change the contents of the external configMap but still have the same name, you manually need to bump it since theres no checksum annotation for it. Should we add checksum annotation so any change made in an external configMap causes a rollout even if the name of the configMap is not changed

internal/controller/datadogagent/global/dependencies.go
```diff
@@ -70,15 +70,30 @@ func addComponentDependencies(logger logr.Logger, ddaMeta metav1.Object, ddaSpec
...
-		// This is mounted in the init-volume container in the agent default code.
+		// Creates / updates system-probe-seccomp configMap to configData or default
 		for _, containerName := range rc.Containers {
 			if containerName == apicommon.SystemProbeContainerName {
-				if !useSystemProbeCustomSeccomp(ddaSpec) {
+				var seccompConfigData map[string]string
+
+				if componentOverride, ok := ddaSpec.Override[v2alpha1.NodeAgentComponentName]; ok {
+					if spContainer, ok := componentOverride.Containers[apicommon.SystemProbeContainerName]; ok {
+						if spContainer.SeccompConfig != nil && spContainer.SeccompConfig.CustomProfile != nil && spContainer.SeccompConfig.CustomProfile.ConfigData != nil {
```
tbavelier: We check here and also in `override/container.go` if ConfigData is not nil based on the user config. We could possibly add a public util function to avoid re-typing it twice 🤔

tbavelier (approved):
> When both configMap and configData are used, checksum annotation is still written in the agent ds. Should the annotation only be written when only configData is used on its own, or is it fine how it is now?

It's fine, it's in line with how it is now for other features, if both options are used, we still add the checksum annotation in the pod template iirc.

> As of right now if you change the contents of the external configMap but still have the same name, you manually need to bump it since theres no checksum annotation for it. Should we add checksum annotation so any change made in an external configMap causes a rollout even if the name of the configMap is not changed

My motto is to replicate behaviour from other similar "features". e.g. like with KSM. When either options are used (configData or (and) configMap), we still add the checksum annotation, however it does not trigger a rollout if a user edits their own configmap I believe. Ideally, it should, but can be discussed offline with the team if that's expected.

One thing is when a user uses `configData`, we should add the checksum annotation on the operator-managed configmap if possible, as that's how we do it currently for other features. Note that's only for `configData`: if a user provides their own configmap with `configMap` directly, we should not annotate it.

PR #2045, opened 2025-07-11
setting default log level to logLevel

alessandro-avendano:
### What does this PR do?

sets default log level to logLevel for new logger since the previous logger showed all logs and did not set the default log level to logLevel. additionally, INFO logs and DEBUG logs will be sent to stdout.

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

- Create AKS cluster with custom operator image
- Once the operator is deployed, go into Azure Insights within the Azure portal and check to see the source of logs. All INFO and DEBUG logs should have the source be stdout, and all logs that are not INFO or DEBUG logs should have the source be stderr.
- Can test with different log levels by setting the default log level in `manager.yaml` with the flag `- --loglevel=<logLevel>`

Logs should look like this below, where INFO and DEBUG logs are sent to stdout and the rest of the logs are sent to stderr:
- EX: default is INFO logs (set by default, no flag needed)
Image 7-11-25 at 2 53 PM

- EX: default is DEBUG logs (everything including DEBUG logs and above will show, meaning that INFO and DEBUG logs will be shown and sent to stdout while the other logs showing (ERROR, etc) will be sent to stderr)
Image 7-15-25 at 10 41 AM

### Checklist

### DIFF
--- a/pkg/controller/utils/metadata/metadata_forwarder.go
+++ b/pkg/controller/utils/metadata/metadata_forwarder.go
@@ -382,7 +382,11 @@ func getURL() string {
 	// check url env var
 	// example: https://app.datadoghq.com
 	if urlFromEnvVar := os.Getenv("DD_URL"); urlFromEnvVar != "" {
-		mdfURL.Host = urlFromEnvVar
+		tempURL, err := url.Parse(urlFromEnvVar)
+		if err == nil {
+			mdfURL.Host = tempURL.Host
+			mdfURL.Scheme = tempURL.Scheme
+		}
 	}
 
 	return mdfURL.String()
--- a/pkg/controller/utils/metadata/metadata_forwarder_test.go
+++ b/pkg/controller/utils/metadata/metadata_forwarder_test.go
@@ -1,0 +1,55 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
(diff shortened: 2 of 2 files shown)


cmd/main.go
```diff
@@ -375,14 +375,15 @@ func customSetupLogging(logLevel zapcore.Level, logEncoder string) error {
...
 
 	core := zap.WrapCore(func(c zapcore.Core) zapcore.Core {
 		infoLevel := zap.LevelEnablerFunc(func(level zapcore.Level) bool {
-			return level == zapcore.InfoLevel
+			return level == zapcore.InfoLevel && zapOpts.Level.Enabled(level)
 		})
 
 		otherLevel := zap.LevelEnablerFunc(func(level zapcore.Level) bool {
-			return level != zapcore.InfoLevel
+			return level != zapcore.InfoLevel && zapOpts.Level.Enabled(level)
```
celenechang: looks better, thanks. i didn't think of this on the other PR, are there other levels we should write to stdout (v. stderr)?

alessandro-avendano: the only possibility i could see is having DEBUG logs sent to stdout if we want to keep stderr with only errors. but that could make stdout a little more verbose, which may not be wanted. i think i am leaning a little more towards keeping stderr for just strictly errors and adding DEBUG logs into stdout, what are your thoughts?

celenechang: yea i think DEBUG would be more appropriate in stdout than stderr, so you could add a conditional to make sure that's the case if DEBUG logging is enabled

PR #2054, opened 2025-07-18
Update label selector

khewonc:
### What does this PR do?

Update the label selector of DDA workloads to use `app.kubernetes.io/instance` instead of `agent.datadoghq.com/name`. The instance value is derived from the DDA name + component suffix or DDA name + profile name

### Motivation

https://datadoghq.atlassian.net/browse/CECO-2318

### Additional Notes

This will break network policies since the network policy pod selectors haven't been updated yet. These need to be updated in a future PR

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

* Upgrade from an operator image pre-this PR to an image with this PR. The label value `app.kubernetes.io/instance` should change to what's described above for profile pods. Default pod labels should not change. No DSs should restart, default pods should not restart, and only profile pods should restart following the DS's rollout strategy
* Apply the annotation `agent.datadoghq.com/update-selector` to the DDA and check that the daemonset is deleted, pods are still running, a new daemosnet is created with the new selector, and the pods have not restarted. Pods should not restart when adding the annotation. Removing the annotation to go back does not guarantee pods will not restart. Deployment pods will not automatically restart and must be manually deleted for the new selector to be applied
* Do the same, but with profiles enabled. Ensure the pod selector value for `app.kubernetes.io/instance` does not match the default daemonset's value

### Checklist


internal/controller/datadogagent/component/agent/default.go
```diff
@@ -771,22 +771,3 @@ func volumeMountsForAgentDataPlane() []corev1.VolumeMount {
 		common.GetVolumeMountForCgroups(),
 	}
 }
-
-func GetDefaultMetadata(owner metav1.Object, componentKind, componentName, version string, selector *metav1.LabelSelector) (map[string]string, map[string]string, *metav1.LabelSelector) {
```
khewonc, line 775: Unused function

internal/controller/datadogagent/common/utils.go
```diff
@@ -49,6 +49,14 @@ func GetDefaultMetadata(owner metav1.Object, componentKind, componentName, versi
 		for key, val := range selector.MatchLabels {
 			labels[key] = val
 		}
+		// if update selector is present, use k8s instance and component as the selector
+	} else if _, ok := owner.GetAnnotations()[apicommon.UpdateSelectorAnnotationKey]; ok {
```
tbavelier: nit: as long as the annotation is present with any value, it will update the selector. Should it be gated behind the value of the annotation, e.g. true ? Not sure if it's worth noting it

tbavelier (approved):
In QA steps:
> Upgrade from an operator image pre-this PR to an image with this PR. The label value app.kubernetes.io/instance should change to what's described above. The DS should not restart and only the pods should restart following the DS's rollout strategy

Only profile pods are rolled out with the label change, it doesn't seem the pods from default DS are rolled out ?

internal/controller/datadogagent/common/utils.go
khewonc: Thanks, changed it to check that the value == `true`. For the question above, the default DS pod labels don't change so they shouldn't update. I'll update the test instructions to make that clearer

levan-m (approved):

PR #2109, opened 2025-08-11
Percentage of unready Pods to stop autoscaling

ADKosm:
### What does this PR do?

Introduction of a parameter for a new autoscaling safety mechanism that allows users to define a threshold for percentage of unready Pods. If the actual percentage of unready Pods in a workload exceeds this configured threshold, all autoscaling activities (both scale-up and scale-down) will be temporarily paused.

### Motivation

We want to have additional safety check in order to allow autoscaling not to interfere with a turbulent cluster.

### Additional Notes

No

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

???

### Checklist

### DIFF
--- a/api/datadoghq/common/datadogpodautoscaler_types.go
+++ b/api/datadoghq/common/datadogpodautoscaler_types.go
@@ -199,6 +199,11 @@ type DatadogPodAutoscalerConstraints struct {
 	// MaxReplicas is the upper limit for the number of POD replicas. Needs to be >= minReplicas.
 	MaxReplicas int32 `json:"maxReplicas"`
 
+	// MaxUnreadyPodsPercent is the maximum percentage of unready pods allowed before pausing scaling. Disabled by default.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=100
+	MaxUnreadyPodsPercent *int32 `json:"maxUnreadyPodsPercent,omitempty"`
+
 	// Containers defines constraints for the containers.
 	Containers []DatadogPodAutoscalerContainerConstraints `json:"containers,omitempty"`
 }
--- a/api/datadoghq/common/zz_generated.deepcopy.go
+++ b/api/datadoghq/common/zz_generated.deepcopy.go
@@ -38,6 +38,11 @@ func (in *DatadogPodAutoscalerConstraints) DeepCopyInto(out *DatadogPodAutoscale
 		*out = new(int32)
 		**out = **in
 	}
+	if in.MaxUnreadyPodsPercent != nil {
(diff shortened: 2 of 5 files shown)


api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -199,6 +199,11 @@ type DatadogPodAutoscalerConstraints struct {
 	// MaxReplicas is the upper limit for the number of POD replicas. Needs to be >= minReplicas.
 	MaxReplicas int32 `json:"maxReplicas"`
 
+	// MaxUnreadyPodsPercent is the maximum percentage of unready pods allowed before pausing scaling. Disabled by default.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=100
+	MaxUnreadyPodsPercent *int32 `json:"maxUnreadyPodsPercent,omitempty"`
```
vboulineau, line 205: I am not sure if it's really a constraint. Thinking about it, it's more a policy to stop Autoscaling in that scenario, which should probably be more on the Cluster Agent side, WDYT?

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -199,6 +199,11 @@ type DatadogPodAutoscalerConstraints struct {
 	// MaxReplicas is the upper limit for the number of POD replicas. Needs to be >= minReplicas.
 	MaxReplicas int32 `json:"maxReplicas"`
 
+	// MaxUnreadyPodsPercent is the maximum percentage of unready pods allowed before pausing scaling. Disabled by default.
```
jbartosik, line 202: I think "Disabled by default." means that "When not specified autoscaling will not be stopped because of unready pods".

api/datadoghq/common/datadogpodautoscaler_types.go
jbartosik, line 205: It looks like the same constraint is applied to vertical and horizontal actuation. Is that intentional?

api/datadoghq/common/datadogpodautoscaler_types.go
ADKosm: Well this it exactly what it is

jennchenn: If scaling won't be stopped because of unready pods, is that equivalent to setting it to `100%`?

andrew--r, line 202: Would it immediately resume scaling when the percentage of unready pods drops below the threshold? Maybe it's worth mentioning in the description.

PR #2157, opened 2025-09-03
[datadogpodautoscaler] Add fallback scaling direction option and last horizontal recommendations

jennchenn:
### What does this PR do?

Add fields to support:
* defining which direction fallback should be enabled (`ScaleUp` by default)
* enable local fallback after 10m by default
* storing the last x number of unconstrained recommendations 

### Motivation

We want to support a lower fallback time / better stabilization window implementation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist


api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -278,6 +278,9 @@ type DatadogPodAutoscalerHorizontalStatus struct {
 
 	// LastActions are the last successful actions done by the controller
 	LastActions []DatadogPodAutoscalerHorizontalAction `json:"lastActions,omitempty"`
+
+	// LastRecommendations stores the most recent recommendations
+	LastRecommendations []DatadogPodAutoscalerRecommendation `json:"lastRecommendations,omitempty"`
```
vboulineau: Isn't it similar to `DatadogPodAutoscalerHorizontalTargetStatus`? We can rename and make `source` optional if needed

jennchenn: Updated - open to suggestions on naming though if current one doesn't make sense!

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -274,23 +274,27 @@ const (
...
+	// +optional
+	Source DatadogPodAutoscalerValueSource `json:"source,omitempty"`
 
 	// GeneratedAt is the timestamp at which the recommendation was generated
 	GeneratedAt metav1.Time `json:"generatedAt,omitempty"`
 
-	// Replicas is the desired number of replicas for the workload
-	Replicas int32 `json:"desiredReplicas"`
+	// Replicas is the recommended number of replicas for the workload
+	Replicas int32 `json:"replicas"`
```
vboulineau: Note: we're going to loose state on upgrade by doing so, it might be easier to keep the name until next major version

vboulineau (approved):

PR #2159, opened 2025-09-03
[CWS] add new configuration to allow CWS to send payloads without using the security agent

paulcacheux:
### What does this PR do?

This PR adds a new feature flag, disabled by default, `DirectSendFromSystemProbe` in the CWS feature group. When this feature is enabled, the CWS product is running 100% in the system-probe and thus doesn't require the security agent.

This PR adds the new config flag, mapping to `runtime_security_config.direct_send_from_system_probe` applied on the system-probe, and removes the requirement on the security agent container (in addition to removing everything that is applied to this container in the first place).

### Motivation

This PR is a pre-requisite to be able to use this feature more globally internally (because of the implications regarding which containers are required). Which is why we mention the experimental-ness of this feature in the documentation.

### Minimum Agent Versions

This feature is available in the agent starting with version 7.71.

### Describe your test plan

In addition to the simple unit tests in the PR, this has been tested on a kind cluster and on `stingchameleon` (an internal experimental cluster).

### Checklist

### DIFF
--- a/docs/configuration.v2alpha1.md
+++ b/docs/configuration.v2alpha1.md
@@ -85,7 +85,7 @@ spec:
 | features.cws.customPolicies.configData | ConfigData corresponds to the configuration file content. |
 | features.cws.customPolicies.configMap.items | Maps a ConfigMap data `key` to a file `path` mount. |
 | features.cws.customPolicies.configMap.name | Is the name of the ConfigMap. |
-| features.cws.directSendFromSystemProbe | DirectSendFromSystemProbe configures CWS to send payloads directly from the system-probe, without using the security-agent. This is an experimental feature, please do not use without validating with support first. Default: false |
+| features.cws.directSendFromSystemProbe | DirectSendFromSystemProbe configures CWS to send payloads directly from the system-probe, without using the security-agent. This is an experimental feature. Contact support before using. Default: false |
 | features.cws.enabled | Enables Cloud Workload Security. Default: false |
 | features.cws.network.enabled | Enables Cloud Workload Security Network detections. Default: true |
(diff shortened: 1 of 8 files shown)


mbertrone (approved):
Overall this looks good to me.

- The flag for enabling `directSendFromSystemProbe` is currently placed under the CWS configuration.
- In the future, other features that could benefit from direct communication may want to introduce their own `directSendFromSystemProbe` flag under their respective configs, so this pattern should work well.
- Marking it as experimental and keeping it disabled by default is a safe approach and I agree with that for now.
- On the flag name itself: I do feel it’s a bit long and could potentially be shortened, but since this naming is already in use in the Agent, it makes sense to keep it consistent for now. I don’t have strong alternative suggestions at this point.

docs/configuration.v2alpha1.md
```diff
@@ -85,6 +85,7 @@ spec:
 | features.cws.customPolicies.configData | ConfigData corresponds to the configuration file content. |
 | features.cws.customPolicies.configMap.items | Maps a ConfigMap data `key` to a file `path` mount. |
 | features.cws.customPolicies.configMap.name | Is the name of the ConfigMap. |
+| features.cws.directSendFromSystemProbe | DirectSendFromSystemProbe configures CWS to send payloads directly from the system-probe, without using the security-agent. This is an experimental feature, please do not use without validating with support first. Default: false |
```
urseberry:
```suggestion
| features.cws.directSendFromSystemProbe | DirectSendFromSystemProbe configures CWS to send payloads directly from the system-probe, without using the security-agent. This is an experimental feature. Contact support before using. Default: false |
```

paulcacheux: thanks ! I applied this in all places in 541b70118b4c5269431e1b88101fc1b35a4eff26

PR #2174, opened 2025-09-08
Autoscaling based on custom metric query

jbartosik:
### What does this PR do?

RFC: https://datadoghq.atlassian.net/wiki/spaces/CAUT/pages/5463776220/Scaling+on+any+Datadog+Query

Add option to DPA to scale based on custom query

Based on a public v2/query/timeseries:

https://docs.datadoghq.com/api/latest/metrics/?code-lang=go#query-timeseries-data-across-multiple-products

https://github.com/DataDog/datadog-api-spec/blob/88478288f31297f3659ac6d7b3ad2a37618fee45/spec/v2/query.yaml#L124

### Motivation

Adding option to scale based on custom queries

### Additional Notes

-

### Minimum Agent Versions

* Cluster Agent: (upcoming)

### Describe your test plan

### Checklist


api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -116,7 +116,7 @@ type DatadogPodAutoscalerScalingRule struct {
 }
 
 // DatadogPodAutoscalerObjectiveType defines the type of the objective.
-// +kubebuilder:validation:Enum:=PodResource;ContainerResource
+// +kubebuilder:validation:Enum:=PodResource;ContainerResource;ControllerObjective
```
vboulineau: I think that `ControllerObjective` name is not very clear for users, something like DatadogQuery or CustomQueries, can be checked with PM

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -138,8 +141,28 @@ type DatadogPodAutoscalerObjective struct {
 
 	// ContainerResource allows to set a container-level resource objective.
 	ContainerResource *DatadogPodAutoscalerContainerResourceObjective `json:"containerResource,omitempty"`
+
+	// ControllerObjective allows to set a controller-level objective.
+	ControllerObjective *DatadogPodAutoscalerControllerObjective `json:"controllerObjective,omitempty"`
+
+	// Mode describes when the objective will be active.
+	// +optional
+	Mode ObjectiveMode `json:"mode,omitempty"`
```
vboulineau: Was discussed but adding the other option there: we can add a dedicated `objective` in `fallback` section

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -167,26 +190,97 @@ type DatadogPodAutoscalerContainerResourceObjective struct {
...
 
+// DatadogPodAutoscalerControllerObjective defines a controller-level objective
+// +kubebuilder:object:generate=true
+type DatadogPodAutoscalerControllerObjective struct {
+	Query TimeseriesFormulaRequest `json:"query"`
+	// Value is the value of the objective
+	Value DatadogPodAutoscalerObjectiveValue `json:"value"`
+}
+
+// TimeseriesFormulaRequest is a subset of the v2 timeseries query API (metrics only).
```
vboulineau: We don't need to stick exactly to the API format. For instance the `Attributes` layer is not useful to us.

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -167,26 +190,97 @@ type DatadogPodAutoscalerContainerResourceObjective struct {
...
+// It mirrors the OpenAPI "TimeseriesFormulaRequestAttributes" fields relevant to autoscaling.
+// Reference: https://github.com/DataDog/datadog-api-spec/blob/94d1542b31ad0df1da915bae84686b13ba1a65ae/spec/v2/query.yaml#L124
+// +kubebuilder:object:generate=true
+type TimeseriesFormulaRequest struct {
```
vboulineau: nit: As the namespace is shared, we normally prefix with `DatadogPodAutoscaler`

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -167,26 +190,97 @@ type DatadogPodAutoscalerContainerResourceObjective struct {
...
+	// +kubebuilder:validation:Required
+	Attributes TimeseriesFormulaRequestAttributes `json:"attributes"`
+}
+
+// Spec mirrors the OpenAPI "TimeseriesFormulaRequestAttributes".
+// +kubebuilder:object:generate=true
+type TimeseriesFormulaRequestAttributes struct {
+	// Optional sampling interval in milliseconds.
+	// +optional
+	Interval *int64 `json:"interval,omitempty"`
```
vboulineau: This field is usually not used, I don't think we should expose it. However we need to allow customization of query window (with a max duration, probably 1h).

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -167,26 +190,97 @@ type DatadogPodAutoscalerContainerResourceObjective struct {
...
+}
+
+// +kubebuilder:object:generate=true
+type QueryFormula struct {
+	// +kubebuilder:validation:MinLength=1
+	Formula string `json:"formula"`
+}
+
+// TimeseriesQuery is a discriminated union. Only Metrics are supported for autoscaling.
+// +kubebuilder:validation:XValidation:rule="has(self.metrics)",message="metrics is required"
```
vboulineau: We cannot use `XValidation` as it requires a recent enough Kube version

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -167,26 +190,97 @@ type DatadogPodAutoscalerContainerResourceObjective struct {
...
+// +kubebuilder:object:generate=true
+type TimeseriesQuery struct {
+	Metrics *MetricsTimeseriesQuery `json:"metrics,omitempty"`
+}
+
+// +kubebuilder:object:generate=true
+type MetricsTimeseriesQuery struct {
+	// Data source (Metrics API): "metrics" or "cloud_cost".
```
vboulineau: We currently target metrics and APM Metrics not cloud cost

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -167,26 +190,97 @@ type DatadogPodAutoscalerContainerResourceObjective struct {
...
+type MetricsDataSource string
+
+const (
+	MetricsDataSourceMetrics   MetricsDataSource = "metrics"
+	MetricsDataSourceCloudCost MetricsDataSource = "cloud_cost"
+)
+
 // DatadogPodAutoscalerObjectiveValueType specifies the type of objective value.
-// kubebuilder:validation:Enum:=Utilization
+// +kubebuilder:validation:Enum:=Utilization;AverageValue
```
vboulineau: I'm not sure if `AverageValue` makes the most sense for external metrics. If we look at internal data, only 1.2% of WPAs are `average`, everything else is `absolute`.

api/datadoghq/common/datadogpodautoscaler_types.go
jbartosik: I asked

api/datadoghq/common/datadogpodautoscaler_types.go
jbartosik: Done

api/datadoghq/common/datadogpodautoscaler_types.go
jbartosik: Done

api/datadoghq/common/datadogpodautoscaler_types.go
jbartosik: Done

api/datadoghq/common/datadogpodautoscaler_types.go
jbartosik: Done

api/datadoghq/common/datadogpodautoscaler_types.go
jbartosik: Done

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -167,26 +190,86 @@ type DatadogPodAutoscalerContainerResourceObjective struct {
...
 
 // DatadogPodAutoscalerObjectiveValue defines the target value of the objective.
 // +kubebuilder:object:generate=true
 type DatadogPodAutoscalerObjectiveValue struct {
-	// Type specifies how the value is expressed (possible values: Utilization).
-	// +kubebuilder:validation:Enum:=Utilization
-	Type DatadogPodAutoscalerObjectiveValueType `json:"type"`
+	// Type specifies how the value is expressed (possible values: Utilization, AbsoluteValue).
+	// +kubebuilder:validation:Enum:=Utilization;AbsoluteValue
+	Type DatadogPodAutoscalerDatadogPodAutoscalerObjectiveValueType `json:"type"`
```
ADKosm: Aren't there too many DatadogPodAutoscalers ? :)

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -138,8 +141,28 @@ type DatadogPodAutoscalerObjective struct {
...
+	Mode ObjectiveMode `json:"mode,omitempty"`
 }
 
+// ObjectiveMode defines when the objective should be active.
+// +kubebuilder:validation:Enum:=RemoteOnly;LocalOnly;RemoteAndLocal
+type ObjectiveMode string
+
+const (
+	// ObjectiveModeRemoteOnly activates the objective for remote recommender only.
+	ObjectiveModeRemoteOnly ObjectiveMode = "RemoteOnly"
```
ADKosm: If I didn't read the comment here I would think that activation of an objective depends on autoscaler creation type - via UI on web or manually via CRD. I think we can polish naming here or move this business logic completely to fallback section as Vincent suggested

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -167,26 +190,86 @@ type DatadogPodAutoscalerContainerResourceObjective struct {
...
+	Queries []DatadogPodAutoscalerTimeseriesQuery `json:"queries"`
+}
+
+// +kubebuilder:object:generate=true
+type DatadogPodAutoscalerQueryFormula struct {
+	// +kubebuilder:validation:MinLength=1
+	Formula string `json:"formula"`
+}
+
+// TimeseriesQuery is a discriminated union. Only Metrics are supported for autoscaling.
```
ADKosm, line 207: For now and we have plans to expand this in the future?

config/crd/bases/v1/datadoghq.com_datadogpodautoscalers.yaml
```diff
@@ -323,10 +323,25 @@ spec:
...
+                                  Use a plain number (e.g., "11" or "11.5").
+                                  Represented as a resource.Quantity to avoid floating point in CRDs.
+                                pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                                x-kubernetes-int-or-string: true
                               type:
-                                description: 'Type specifies how the value is expressed (possible values: Utilization).'
-                                enum:
-                                  - Utilization
+                                allOf:
+                                  - enum:
```
ADKosm: Aren't there too many enums in this allOf? :)

config/crd/bases/v1/datadoghq.com_datadogpodautoscalers.yaml
```diff
@@ -323,10 +323,25 @@ spec:
...
                             properties:
+                              absoluteValue:
+                                anyOf:
+                                  - type: integer
+                                  - type: string
+                                description: |-
+                                  AbsoluteValue defines a target as an absolute value divided by the number of running pods.
+                                  Use a plain number (e.g., "11" or "11.5").
+                                  Represented as a resource.Quantity to avoid floating point in CRDs.
+                                pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
```
ADKosm, line 334: Hmm, as far as I understand this pattern, it also allows user to use units like 2.5M, 4Gi, etc. And scientific notation as well. 
Do we want to allow all of this now? I mean if user enters 500Mi how should we treat this input - just multiply to 2^20? Maybe start with just plain number for custom query?

ADKosm: I mean since it is a custom query user can always just multiply the metric to appropriate number and convert it to useful scale - and there will be no ambiguity about this input

config/crd/bases/v1/datadoghq.com_datadogpodautoscalers.yaml
```diff
@@ -342,6 +357,102 @@ spec:
...
+                                    metrics:
+                                      properties:
+                                        dataSource:
+                                          allOf:
+                                            - enum:
+                                                - metrics
+                                                - cloud_cost
+                                            - enum:
+                                                - metrics
+                                                - cloud_cost
```
ADKosm: Is it also a refundant dublication or it really serves some purpose? 🤔

api/datadoghq/common/datadogpodautoscaler_types.go
jbartosik: 😅

config/crd/bases/v1/datadoghq.com_datadogpodautoscalers.yaml
jbartosik: I initially wanted to use float here but floats are not allowed

So we can go with string - which means custom parsing

Or my preffered option - K8s standard resource.Quantity

config/crd/bases/v1/datadoghq.com_datadogpodautoscalers.yaml
jbartosik: I'll check why it gets generated like that

config/crd/bases/v1/datadoghq.com_datadogpodautoscalers.yaml
jbartosik: I'll check why it gets genertated like this

config/crd/bases/v1/datadoghq.com_datadogpodautoscalers.yaml
jbartosik: Done

config/crd/bases/v1/datadoghq.com_datadogpodautoscalers.yaml
jbartosik: Done

api/datadoghq/common/datadogpodautoscaler_types.go
jbartosik: Done

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -167,26 +173,83 @@ type DatadogPodAutoscalerContainerResourceObjective struct {
...
+	// +kubebuilder:validation:MinLength=1
+	Query string `json:"query"`
+}
+
+// +kubebuilder:validation:Enum=metrics;cloud_cost
+type DatadogPodAutoscalerMetricsDataSource string
+
+const (
+	MetricsDataSourceMetrics   DatadogPodAutoscalerMetricsDataSource = "metrics"
+	MetricsDataSourceCloudCost DatadogPodAutoscalerMetricsDataSource = "cloud_cost"
```
andrew--r: question: why is there `cloud_cost`? AFAIK we didn't consider supporting `cloud_cost` and wanted to support APM metrics instead

config/crd/bases/v1/datadoghq.com_datadogpodautoscalers.yaml
ADKosm: Am I right that standard resource.Quantity is basically custom parsing but just with ready-to-use regex from k8s?
I mean if it is the only advantage of resource.Quantity maybe it is worth it to just do our own string parsing?
In that case it will actually reduce complexity and ambiguity of usage of this field

jbartosik: resource.Quantity has also advantages of:
- being familiar to customers
- sometimes its nice to use different units (eg. you're scaling based on nanocpus and counting exactly how many zeores are in a value

ADKosm: Deal 🤝

ADKosm (approved):

api/datadoghq/common/datadogpodautoscaler_types.go
jbartosik: Looks like CustomQuery has more support so I went with that.

api/datadoghq/common/datadogpodautoscaler_types.go
andrew--r: This question is still pending

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -167,26 +173,83 @@ type DatadogPodAutoscalerContainerResourceObjective struct {
...
+
 // DatadogPodAutoscalerObjectiveValueType specifies the type of objective value.
-// kubebuilder:validation:Enum:=Utilization
+// +kubebuilder:validation:Enum:=Utilization;AbsoluteValue
 type DatadogPodAutoscalerObjectiveValueType string
 
 const (
 	// DatadogPodAutoscalerUtilizationObjectiveValueType declares an objective based on a Utilization (percentage, 0-100).
 	DatadogPodAutoscalerUtilizationObjectiveValueType DatadogPodAutoscalerObjectiveValueType = "Utilization"
+	// DatadogPodAutoscalerAbsoluteValueObjectiveValueType declares an objective based on an AbsoluteValue (absolute value divided by the number of running pods).
```
vboulineau: Comment is incorrect we don't divide by number of Pods

api/datadoghq/common/datadogpodautoscaler_types.go
vboulineau: Indeed, still need a change

api/datadoghq/common/datadogpodautoscaler_types.go
vboulineau: We should allow further extension yes

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -167,26 +173,83 @@ type DatadogPodAutoscalerContainerResourceObjective struct {
 	Container string `json:"container"`
 }
 
+// DatadogPodAutoscalerCustomQueryObjective defines a controller-level objective
+// +kubebuilder:object:generate=true
+type DatadogPodAutoscalerCustomQueryObjective struct {
+	Query DatadogPodAutoscalerTimeseriesFormulaRequest `json:"query"`
```
vboulineau, line 180: The query window is missing

vboulineau (commented):
Please update example YAML at the top of `api/datadoghq/v1alpha2/datadogpodautoscaler_types.go` with all possible fields set, mostly LGTM, some comments to be addressed though.

api/datadoghq/common/datadogpodautoscaler_types.go
jbartosik: Done.

api/datadoghq/common/datadogpodautoscaler_types.go
jbartosik: Done

jbartosik:
> Please update example YAML at the top of `api/datadoghq/v1alpha2/datadogpodautoscaler_types.go` with all possible fields set, mostly LGTM, some comments to be addressed though.

Done

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -167,26 +173,88 @@ type DatadogPodAutoscalerContainerResourceObjective struct {
...
+	// +kubebuilder:validation:MinLength=1
+	Query string `json:"query"`
+}
+
+// +kubebuilder:validation:Enum:=metrics;apm_metrics
+type DatadogPodAutoscalerMetricsDataSource string
+
+const (
+	MetricsDataSourceMetrics    DatadogPodAutoscalerMetricsDataSource = "metrics"
+	MetricsDataSourceApmMetrics DatadogPodAutoscalerMetricsDataSource = "apm_metrics"
```
andrew--r: question: how would the `DatadogPodAutoscalerMetricsTimeseriesQuery` look like with `apm_metrics` data source? AFAIK APM Metrics queries do not have the same stringified query format as regular metrics and require some custom JSON instead

jbartosik: I removed this from the PR

I hope we can add it soon

PR #2178, opened 2025-09-09
[CXP-2595] Clean up RunProcessChecksInCoreAgent config

daniel-taf:
### What does this PR do?

This PR cleans up the `RunProcessChecksInCoreAgent` config that can be used to control whether checks/features owned by container-experiences (process, container, process_discovery) should run in the process agent or core agent. The default behavior is to use the core agent unless the agent version does not support it. One could still override this behavior using env vars but we are reserving that for when customers have issues.

Deprecation notice has been in since 1.19: https://github.com/DataDog/datadog-operator/pull/2201

### Motivation

We now want to discourage the use of the process-agent where possible, so we no longer want the config to be used/available. 

### Additional Notes

Open questions: 
- Would deleting the config fully cause a breakage for this who have it in their .yaml files? 
- How should we handle these breaking tests since the config they used doesn't exist anymore:
    - `Test_AutopilotOverrides/autopilot_enabled_with_core-agent_and_process-agent`  
    - `Test_AutopilotOverrides/autopilot_enabled_with_all_containers_(core,_trace,_process)`

### Minimum Agent Versions

* Agent: N/A
* Cluster Agent: N/A

### Describe your test plan

I believe unit tests are generally sufficient since the config has been true by default for a while. However, for manual QA, one can deploy the operator and ensure:

1. The process agent does not come up
2. With the default config: the process discovery and container check is on the core agent.
3. With process collection enabled: the process check is on the core agent.

Steps 2 and 3 can be confirmed via logs or agent status output.

### Checklist

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -498,6 +498,7 @@ func setupAndStartOperatorMetadataForwarder(logger logr.Logger, client client.Re
 		DatadogSLOEnabled:             options.datadogSLOEnabled,
 		DatadogGenericResourceEnabled: options.datadogGenericResourceEnabled,
 		DatadogAgentProfileEnabled:    options.datadogAgentProfileEnabled,
+		DatadogAgentInternalEnabled:   options.datadogAgentInternalEnabled,
 		LeaderElectionEnabled:         options.enableLeaderElection,
 		ExtendedDaemonSetEnabled:      options.supportExtendedDaemonset,
 		RemoteConfigEnabled:           options.remoteConfigEnabled,
--- a/hack/patch-bundle.sh
+++ b/hack/patch-bundle.sh
@@ -40,8 +40,8 @@ $YQ -i ".spec.\"replaces\" = \"datadog-operator.v$LATEST_VERSION\"" bundle/manif
 # Add OpenShift version annotation (adding in main bundle as it's used for OpenShift Community)
 $YQ -i ".annotations.\"com.redhat.openshift.versions\" = \"v4.6\"" bundle/metadata/annotations.yaml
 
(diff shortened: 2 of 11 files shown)


internal/controller/datadogagent/controller_v2_test.go
```diff
@@ -367,32 +367,6 @@ func TestReconcileDatadogAgentV2_Reconcile(t *testing.T) {
...
-			wantFunc: func(c client.Client) error {
-				expectedContainers := []string{
-					string(apicommon.CoreAgentContainerName),
-					string(apicommon.ProcessAgentContainerName),
-					string(apicommon.TraceAgentContainerName),
-				}
-
-				return verifyDaemonsetContainers(c, resourcesNamespace, dsName, expectedContainers)
-			},
-		},
```
daniel-taf, line 396: I deleted this since container monitoring can't be run in the process agent using the builder anymore. Let me know if something else is preferred.

internal/controller/datadogagent/controller_v2_test.go
```diff
@@ -1007,7 +981,7 @@ func Test_AutopilotOverrides(t *testing.T) {
 				dda := testutils.NewInitializedDatadogAgentBuilder(resourcesNamespace, resourcesName).
 					WithAPMEnabled(true).
 					WithLiveProcessEnabled(true).
-					WithProcessChecksInCoreAgent(false).
+					WithNPMEnabled(true).
```
daniel-taf: Since these tests require the process agent, I enabled NPM but that adds the system probe container which this test doesn't check hence the failures on CI. What's the preferred way to handle this test? I am unfamiliar with its purpose.

PR #2183, opened 2025-09-17
fix: mutating admission controller setup ordering issue

scottgerring:
# Fix: Mutating Admission Controller Setup Ordering Issue

## Problem

Setting `features.apm.instrumentation.enabled = true` does not consistently cause the mutating admission controller to begin to SSI-inject pods, unless we explicitly also set the `features.admissionController.mutation.enabled = true` value. This is _despite the fact_ that the mutation feature is enabled by default. 

The upshot of this is that users can configure SSI correctly - as in, their agent yaml exactly reflects our guidance to configure it - and it simply won't turn on. @spottsdd and I have been able to reproduce this in our lab environment consistently. 

## Why?

I think there's some timing issue in the way the reconciliation interacts with the default configuration set when a mutating change is applied. Changing the guard here seems to address the issue, such that we _always_ create the two webhookConfigs and just conditionally set the `enabled` flag. 

https://github.com/DataDog/datadog-operator/blob/da95e2b6c9c0d338b732422ab7f17b43c61ace08/internal/controller/datadogagent/feature/admissioncontroller/feature.go#L101-L107

### DIFF
--- a/internal/controller/datadogagent/feature/admissioncontroller/feature.go
+++ b/internal/controller/datadogagent/feature/admissioncontroller/feature.go
@@ -99,11 +99,11 @@ func (f *admissionControllerFeature) Configure(dda metav1.Object, ddaSpec *v2alp
 	ac := ddaSpec.Features.AdmissionController
 
 	if ac != nil && apiutils.BoolValue(ac.Enabled) {
-		if ac.Validation != nil && ac.Validation.Enabled != nil {
-			f.validationWebhookConfig = &ValidationConfig{enabled: apiutils.BoolValue(ac.Validation.Enabled)}
+		f.validationWebhookConfig = &ValidationConfig{
+			enabled: ac.Validation != nil && apiutils.BoolValue(ac.Validation.Enabled),
 		}
-		if ac.Mutation != nil && ac.Mutation.Enabled != nil {
-			f.mutationWebhookConfig = &MutationConfig{enabled: apiutils.BoolValue(ac.Mutation.Enabled)}
+		f.mutationWebhookConfig = &MutationConfig{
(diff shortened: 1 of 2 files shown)


internal/controller/datadogagent/feature/admissioncontroller/feature.go
```diff
@@ -99,11 +99,11 @@ func (f *admissionControllerFeature) Configure(dda metav1.Object, ddaSpec *v2alp
 	ac := ddaSpec.Features.AdmissionController
 
 	if ac != nil && apiutils.BoolValue(ac.Enabled) {
-		if ac.Validation != nil && ac.Validation.Enabled != nil {
```
stanistan, line 102: I think the intention here is not to override cluster-agent defaults when the setting in the operator is not configured.

So something more like:

```go
if ac.Validation != nil {
  f.validationWebhookConfig = &ValidationConfig{
    enabled: apiutils.BoolValue(ac.Validation.Enabled),
  }
}
```

levan-m:
@scottgerring, I don't think this change makes any difference in practice, `features.admissionController.mutation.enabled` is initialized to internal default and it's set to `true` unless done otherwise in DDA. That's why adding  to existing manifest doesn't change DCA config and should change Admission Controller either. With the help of @spottsdd, I reproduced the issue but very sporadically and at least on one occasion restarting DCA resolved. Otherwise it worked just fine without adding `features.admissionController.mutation.enabled = true`.
I'll close the PR, please reach out if you have any questions.

scottgerring:
Hey @levan-m, i've been on break.

Yep - it is quite easy to reproduce in the workshop environment! I will try make a minimal repro with KIND and loop back; as it stands I appreciate this isn't very convincing.

PR #2184, opened 2025-09-18
Add DD_DASHBOARD_FORCE_SYNC_PERIOD environment variable

### LINKED ISSUE #2179: DatadogDashboard - add configurable DD_DASHBOARD_FORCE_SYNC_PERIOD environment variable
## Problem
DatadogDashboard controller has a hardcoded 60-minute force sync period, unlike DatadogMonitor which supports `DD_MONITOR_FORCE_SYNC_PERIOD`. This means dashboard drift from manual UI changes can persist for up to 60 minutes.

## Proposed Solution
Add `DD_DASHBOARD_FORCE_SYNC_PERIOD` environment variable to configure dashboard sync frequency, similar to the existing DatadogMonitor implementation.

puretension:
### What does this PR do?

Adds DD_DASHBOARD_FORCE_SYNC_PERIOD environment variable to configure DatadogDashboard controller sync frequency, following the same pattern as DD_MONITOR_FORCE_SYNC_PERIOD.

### Motivation

Fixes #2179  - DatadogDashboard controller has a hardcoded 60-minute force sync period, unlike DatadogMonitor which supports DD_MONITOR_FORCE_SYNC_PERIOD. This means dashboard drift from manual UI changes can persist for up to 60 minutes.

### Additional Notes

- Implementation follows the exact same pattern as DatadogMonitor controller
- Defaults to 60 minutes if environment variable is not set
- Includes error handling for invalid values
- All existing tests pass

### Minimum Agent Versions

No minimum version requirements - this is an operator-only change.

* Agent: N/A
* Cluster Agent: N/A

### Describe your test plan

- Verified existing DatadogDashboard tests still pass
- Tested environment variable parsing logic with valid/invalid values
- Confirmed lint and build checks pass

### Checklist

### DIFF
--- a/internal/controller/datadogdashboard/controller.go
+++ b/internal/controller/datadogdashboard/controller.go
@@ -2,6 +2,8 @@ package datadogdashboard
 
 import (
 	"context"
+	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -28,10 +30,11 @@ import (
 )
 
 const (
-	defaultRequeuePeriod    = 60 * time.Second
-	defaultErrRequeuePeriod = 5 * time.Second
-	defaultForceSyncPeriod  = 60 * time.Minute
-	datadogDashboardKind    = "DatadogDashboard"
+	defaultRequeuePeriod                = 60 * time.Second
+	defaultErrRequeuePeriod             = 5 * time.Second
+	defaultForceSyncPeriod              = 60 * time.Minute
+	datadogDashboardKind                = "DatadogDashboard"
+	DDDashboardForceSyncPeriodEnvVar    = "DD_DASHBOARD_FORCE_SYNC_PERIOD"
 )
 
 type Reconciler struct {
@@ -62,6 +65,18 @@ func (r *Reconciler) internalReconcile(ctx context.Context, req reconcile.Reques
 	logger := r.log.WithValues("datadogdashboard", req.NamespacedName)
 	logger.Info("Reconciling Datadog Dashboard")
 	now := metav1.NewTime(time.Now())
+
+	forceSyncPeriod := defaultForceSyncPeriod
+
+	if userForceSyncPeriod, ok := os.LookupEnv(DDDashboardForceSyncPeriodEnvVar); ok {
+		forceSyncPeriodInt, err := strconv.Atoi(userForceSyncPeriod)
+		if err != nil {
+			logger.Error(err, "Invalid value for dashboard force sync period. Defaulting to 60 minutes.")
+		} else {
+			logger.V(1).Info("Setting dashboard force sync period", "minutes", forceSyncPeriodInt)
+			forceSyncPeriod = time.Duration(forceSyncPeriodInt) * time.Minute
+		}
+	}
 
 	instance := &v1alpha1.DatadogDashboard{}
 	var result ctrl.Result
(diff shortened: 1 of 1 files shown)


puretension:
Hi! Could someone please add the 'enhancement' label and either a milestone or the 'qa/skip-qa' label to this PR? 
The CI is failing due to missing labels/milestone.

internal/controller/datadogdashboard/controller.go
```diff
@@ -106,7 +121,7 @@ func (r *Reconciler) internalReconcile(ctx context.Context, req reconcile.Reques
 		if instanceSpecHash != statusSpecHash {
 			logger.Info("DatadogDashboard manifest has changed")
 			shouldUpdate = true
-		} else if instance.Status.LastForceSyncTime == nil || ((defaultForceSyncPeriod - now.Sub(instance.Status.LastForceSyncTime.Time)) <= 0) {
+		} else if instance.Status.LastForceSyncTime == nil || ((forceSyncPeriod - now.Sub(instance.Status.LastForceSyncTime.Time)) <= 0) {
```
orkhanM, line 124: @puretension Thanks for putting this together! It looks good I think but it seems like this doesn't follow the same pattern as the monitors do exactly.  I haven't dived too deeply into the patterns used in other crd controllers here but the Monitor controller seems to check the Monitor specific `MonitorLastForceSyncTime` rather than the generic LastForceSyncTime: https://github.com/DataDog/datadog-operator/blob/900987e33738ab0b0cd26d7fd287e1b4121b5730/internal/controller/datadogmonitor/controller.go#L154-L178

The `MonitorLastForceSyncTime` seems to be set whenever a monitor is synced. I'm not sure if having a Monitor specific variable there was intentional or not or if the maintainers care about consistency in this case but thought it'd be worthwhile to bring to your attention.

orkhanM (commented):
Should any documentation be updated?

internal/controller/datadogdashboard/controller.go
puretension: @orkhanM Great catch! You're absolutely right about the consistency issue.

Looking at the Monitor controller pattern you referenced, 
I can see it uses MonitorLastForceSyncTime for the force sync check. 
I'll update the Dashboard implementation to use DashboardLastForceSyncTime 
to match this pattern for consistency across controllers.

Thanks for pointing this out!

puretension:
> Should any documentation be updated?

I don't think any documentation updates are needed for this change 
since it's an internal API field rename for consistency with the Monitor controller pattern. 
But I'm not entirely sure😅 would appreciate the maintainers' input on this!

tbavelier (changes requested):
Hello @puretension ,

Thank you for the contribution! While I understand the CRD was changed to be consistent with the monitor one, such changes could break customers that do not update the CRD while updating the operator by trying to access a field that does not exist. Moreover, changing a CRD (`_types.go`) should cause more files to be changed (`make generate` should show these changes).

Could you please remove the last commit ? I think `docs/datadog_dashboard.md` could also benefit from this change being documented.

Thanks!

puretension:
Hi @tbavelier,

Thank you for the detailed feedback and for explaining the compatibility concerns with CRD changes. I've removed the last commit that modified the CRD to avoid potential issues for customers who might update the operator without updating the CRD. I also added documentation for the `DD_DASHBOARD_FORCE_SYNC_PERIOD` environment variable in docs/datadog_dashboard.md as you suggested.

I appreciate you taking the time to explain why the CRD change could be problematic and the importance of running make generate when modifying types. Please let me know if there's anything else I missed or if I should make any other adjustments to the implementation.

Thanks again for your guidance!

docs/datadog_dashboard.md
tbavelier, line 1: This seems doubled up ? In any case, could you instead modify the step 3 in this doc to mention it there ?
```markdown
This automatically creates a new dashboard in Datadog. You can find it on the [Dashboards][8] page of your Datadog account.
    Datadog Operator occasionally reconciles and keeps dashboards in line with the given configuration. There is also a force 
    sync every hour, so if a user deletes a dashboard in the Datadog UI, Datadog Operator restores it in under an hour.
```

puretension: @tbavelier  Thanks for the feedback! I've updated the documentation based on your comments. 
I removed the duplicated Configuration section and modified step 3 to reference the `DD_DASHBOARD_FORCE_SYNC_PERIOD` environment variable in the force sync description. 
This makes the connection between the configuration section and operational behavior much clearer for users! Please let me know if any additional changes are needed. Thanks!

tbavelier:
Thank you for the contribution! Re-created a branch locally since our CI doesn't run from forks contribution: https://github.com/DataDog/datadog-operator/pull/2257 (note I also pushed change to the documentation to avoid you the hassle).
I'll be closing this one since the PR is now merged

PR #2190, opened 2025-09-19
[profiles] Reduce dap status update frequency

khewonc:
### What does this PR do?

Changes DAP status update method to compare old and new status to determine whether it needs updating instead of updating every time it's called

**Note**
* This PR makes it so the status is not updated unless a non-time-based field changes. This means `LastUpdate` in the status will not change unless a change was made to another field such as `CurrentHash`
* This behavior is different from the DDA status update

### Motivation

https://datadoghq.atlassian.net/browse/AGENTONB-1698

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

1. Frequency of updates
* Check the operator rollout dashboard internally to confirm the # k8s requests to update profiles has decreased
2. Updates behavior change
* Deploy the operator with profiles enabled
* Spin up a DDA + DAP and check the DAP status. Note the `LastUpdate` time. As long as there are no changes to the cluster, this time should not change even after a minute or so
* Change the DAP (make it invalid, config change, etc) and check that the status is updated accordingly (status fields such as `Valid` or `CurrentHash` change to match the DAP change). Check the `LastUpdate` time to ensure it updated

### Checklist

### DIFF
--- a/internal/controller/datadogagent/controller_reconcile_v2.go
+++ b/internal/controller/datadogagent/controller_reconcile_v2.go
@@ -250,21 +250,20 @@ func (r *Reconciler) updateStatusIfNeededV2(logger logr.Logger, agentdeployment 
 	return result, currentError
 }
 
-func (r *Reconciler) updateDAPStatus(logger logr.Logger, profile *datadoghqv1alpha1.DatadogAgentProfile) (reconcile.Result, error) {
+func (r *Reconciler) updateDAPStatus(ctx context.Context, logger logr.Logger, profile *datadoghqv1alpha1.DatadogAgentProfile, oldStatus *datadoghqv1alpha1.DatadogAgentProfileStatus) (reconcile.Result, error) {
 	// update dap status for non-default profiles only
 	if !agentprofile.IsDefaultProfile(profile.Namespace, profile.Name) {
-		current := datadoghqv1alpha1.DatadogAgentProfile{}
-		if err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: profile.Namespace, Name: profile.Name}, &current); err != nil {
-			return reconcile.Result{}, err
-		}
-
-		if !agentprofile.IsEqualStatus(&current.Status, &profile.Status) {
-			// copy status to current profile to avoid updating aside from status
-			current.Status = profile.Status
-			if err := r.client.Status().Update(context.TODO(), &current); err != nil {
+		if !agentprofile.IsEqualStatus(oldStatus, &profile.Status) {
+			// Update a deep copy to avoid mutating the in-memory object used later
+			toUpdate := profile.DeepCopy()
+			if err := r.client.Status().Update(ctx, toUpdate); err != nil {
 				if apierrors.IsConflict(err) {
 					logger.V(1).Info("unable to update DatadogAgentProfile status due to update conflict")
 					return reconcile.Result{RequeueAfter: time.Second}, nil
+				}
+				if apierrors.IsNotFound(err) {
+					// Profile deleted between list and update; no action needed
+					return reconcile.Result{}, nil
 				}
 				logger.Error(err, "unable to update DatadogAgentProfile status")
 				return reconcile.Result{}, err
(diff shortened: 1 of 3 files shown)


pkg/condition/condition.go
```diff
@@ -475,3 +475,25 @@ func CombineDeploymentStatus(deploymentStatus *v2alpha1.DeploymentStatus, ddaiSt
 	deploymentStatus.DeploymentName = ddaiStatus.DeploymentName
 	return deploymentStatus
 }
+
+func IsEqualConditions(current []metav1.Condition, newCond []metav1.Condition) bool {
+	if len(current) != len(newCond) {
+		return false
+	}
+	for i := range current {
+		if !IsEqualCondition(&current[i], &newCond[i]) {
```
tbavelier:
```suggestion
	// Compare order-insensitively. The CRD uses listMapKey=type so types are unique.
	ac := append([]metav1.Condition(nil), current...)
	bc := append([]metav1.Condition(nil), newCond...)
	slices.SortFunc(ac, func(a, b metav1.Condition) int { return cmp.Compare(a.Type, b.Type) })
	slices.SortFunc(bc, func(a, b metav1.Condition) int { return cmp.Compare(a.Type, b.Type) })
	for i := range ac {
		if !IsEqualCondition(&ac[i], &bc[i]) {
```

We have no guarantee on the conditions ordering, so let's sort by type which is unique to guarantee it as a safeguard ?

internal/controller/datadogagent/controller_reconcile_v2.go
```diff
@@ -250,16 +250,28 @@ func (r *Reconciler) updateStatusIfNeededV2(logger logr.Logger, agentdeployment
...
+		}
+
+		if !agentprofile.IsEqualStatus(&current.Status, &profile.Status) {
+			// copy status to current profile to avoid updating aside from status
+			current.Status = profile.Status
+			if err := r.client.Status().Update(context.TODO(), &current); err != nil {
+				if apierrors.IsConflict(err) {
+					logger.V(1).Info("unable to update DatadogAgentProfile status due to update conflict")
+					return reconcile.Result{RequeueAfter: time.Second}, nil
+				}
```
tbavelier, line 263:
```suggestion
				}
				if apierrors.IsNotFound(err) {
					// Profile deleted between list and update; no action needed
					return reconcile.Result{}, nil
				}
```

Avoid potential error/un-necessary requeue if profile gets deleted after the list from operator, but before update

internal/controller/datadogagent/controller_reconcile_v2.go
```diff
@@ -250,16 +250,28 @@ func (r *Reconciler) updateStatusIfNeededV2(logger logr.Logger, agentdeployment
...
+func (r *Reconciler) updateDAPStatus(logger logr.Logger, profile *datadoghqv1alpha1.DatadogAgentProfile) (reconcile.Result, error) {
 	// update dap status for non-default profiles only
 	if !agentprofile.IsDefaultProfile(profile.Namespace, profile.Name) {
-		if err := r.client.Status().Update(context.TODO(), profile); err != nil {
-			if apierrors.IsConflict(err) {
-				logger.V(1).Info("unable to update DatadogAgentProfile status due to update conflict")
+		current := datadoghqv1alpha1.DatadogAgentProfile{}
+		if err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: profile.Namespace, Name: profile.Name}, &current); err != nil {
+			return reconcile.Result{}, err
+		}
```
tbavelier:
```suggestion
```

We could avoid this call to the **cache** and  _some_ deepcopy allocations by passing previous status to `updateDAPStatus` signature: 
image
image
(The deepcopy is necessary for the update for resourceVersion, that's what we were doing previously anyway with the get into current)

tbavelier:
Pushed above suggestions on https://github.com/DataDog/datadog-operator/pull/2192 / https://github.com/DataDog/datadog-operator/compare/khewonc/dap-status-update...tbavelier/khewonc/dap-status-update so it's easier to understand

PR #2202, opened 2025-09-24
[DatadogAgentInternal][AGENTONB-2546] Profile DDAI name override

tbavelier:
### What does this PR do?

* When generating profile DDAIs, include a nodeAgent component name override based on the profile to ensure the daemonsets are unique **WHEN dda already includes a nodeAgent component name override**
* Generate dependencies using the original DDA name when called from DDAI controller (using the DDA name label that's only on DDAIs) -> noop from DDA perspective

### Motivation

Internal testing

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

* Tested on internal staging clusters: enable DDAI and ensure no issues, daemonsets are proper, no pod creation/deletion with flip-flopping of the ownership between each DDAI

### Checklist


internal/controller/datadogagent/profile_test.go
```diff
@@ -619,3 +622,57 @@ func Test_setProfileDDAIMeta(t *testing.T) {
...
+func Test_setProfileNodeAgentOverride(t *testing.T) {
+	testCases := []struct {
+		name                                   string
+		ddai                                   v1alpha1.DatadogAgentInternal
+		profile                                v1alpha1.DatadogAgentProfile
+		expectedNodeAgentComponentNameOverride *string
+		expectedLabels                         map[string]string
+	}{
+		{
+			name: "non-default profile should get DaemonSet name override",
```
tbavelier, line 635: N.B.: there's no test case when it's default (=no) profile since this function is only called for non-default profiles

internal/controller/datadogagent/component/agent/default.go
```diff
@@ -556,6 +556,15 @@ func initSeccompSetupContainer() corev1.Container {
...
 func commonEnvVars(dda metav1.Object) []corev1.EnvVar {
+	// Use the original DDA name for dependencies that are created by the DatadogAgent controller
+	ddaForDependencies := dda
+	if originalDDAName, ok := dda.GetLabels()[apicommon.DatadogAgentNameLabelKey]; ok {
+		ddaForDependencies = &metav1.ObjectMeta{
+			Name:      originalDDAName,
+			Namespace: dda.GetNamespace(),
+		}
+	}
+
```
levan-m: I think this might be non-functional change.  [GetClusterAgentServiceName](https://github.com/DataDog/datadog-operator/blob/8c52bc1fc73e15b8d0e45457b34068ba46e50910/internal/controller/datadogagent/component/clusteragent/default.go#L28C1-L30C2) uses [constants.GetDDAName(dda)](https://github.com/DataDog/datadog-operator/blob/8c52bc1fc73e15b8d0e45457b34068ba46e50910/pkg/constants/utils.go#L215-L220) which derives name from the label or falls back to current name
```go
func GetDDAName(dda metav1.Object) string {
	if val, ok := dda.GetLabels()[apicommon.DatadogAgentNameLabelKey]; ok && val != "" {
		return val
	}
	return dda.GetName()
}
```

internal/controller/datadogagent/component/agent/default.go
```diff
@@ -665,10 +674,19 @@ func volumeMountsForInitConfig() []corev1.VolumeMount {
...
 func volumesForAgent(dda metav1.Object, requiredContainers []apicommon.AgentContainerName) []corev1.Volume {
+	// Use the original DDA name for dependencies that are created by the DatadogAgent controller
+	ddaForDependencies := dda
+	if originalDDAName, ok := dda.GetLabels()[apicommon.DatadogAgentNameLabelKey]; ok {
+		ddaForDependencies = &metav1.ObjectMeta{
+			Name:      originalDDAName,
+			Namespace: dda.GetNamespace(),
+		}
+	}
+
```
levan-m: Same comment as above in case of `GetVolumeInstallInfo`. However, 
`GetVolumeForSecurity` uses a different [function](https://github.com/DataDog/datadog-operator/blob/8c52bc1fc73e15b8d0e45457b34068ba46e50910/internal/controller/datadogagent/common/utils.go#L89-L91) `GetDefaultSeccompConfigMapName` and gets the DDA name directly from metadata. Same function is used [here](https://github.com/DataDog/datadog-operator/blob/8c52bc1fc73e15b8d0e45457b34068ba46e50910/internal/controller/datadogagent/global/dependencies.go#L99-L107) too so I wonder if the fix should be in the `GetDefaultSeccompConfigMapName` so they use common name and this file may not need to change.

tbavelier: Thanks, great call indeed, added in 38a11de03fc4d1f8894e7b2b7945804b1762fe9e

PR #2214, opened 2025-10-01
[AGENTONB-2555][DatadogAgentInternal] Fix DDAI metric forwaders errors

tbavelier:
### What does this PR do?

* d16694d3dcb7f7ce5bebc6dd1d7f598ef65a74dc: try to determine the Go type instead of fallbacking on Go annotation to determine objectKind and do not copy over kubectl apply annotation on DDAI
    * This is because we'd copy over the annotation from DDA and when using k apply, we'd think it's a DDA for metric forwaders instead of DDAI leading to `{"level":"ERROR","ts":"2025-10-01T12:41:23.523Z","logger":"DatadogMetricForwarders","msg":"cannot retrieve DatadogAgent to get Datadog credentials, will retry later...","CustomResource.Namespace":"system","CustomResource.Name":"foo","error":"DatadogAgent.datadoghq.com \"foo\" not found"}` since there's no `foo` DDA, only `foo` DDAI from profile
* f519e04771741aa23a0677b6970f1916286acb5e: DDAI metrics forwarder would never get reconcile error, so it would always stay in init and log `last reconcile error not updated"` -> we now set the first lastReconcileErr to nil after successfully connecting to DD API
    * We would not see it on DDA as it would always error first, then reconcile after getting credentials

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

* First commit: tested on `kind`, deploy DDA and DAP, with DDAI enabled. Ensure no metric forwarders errors like the one in the description
* Second commit: tested on staging

### Checklist


pkg/controller/utils/datadog/common.go
```diff
@@ -21,10 +23,22 @@ func getObjKind(obj client.Object) string {
 	// First try to get the kind from GVK
 	objKind := obj.GetObjectKind().GroupVersionKind().Kind
 
-	// There is a known bug where the object frequently has empty GVK info. This is a workaround to get object Kind if that happens.
+	// There is a known bug where the object frequently has empty GVK info.
 	// Ref: https://github.com/kubernetes-sigs/controller-runtime/issues/1735
+	// If GVK is empty, prefer inspecting the concrete type to avoid annotation-induced misclassification
+	if objKind == "" {
```
khewonc, line 29: Not specific to this PR, but I think we can refactor this function in the future so we don't have so many `if objKind == ""` statements

levan-m: +1, all three can be inside one conditional.

levan-m (approved):
Looks good. However I think we lost unit test coverage for getting type from annotation, we create `client.Object` either using DDAI or DDA so type cast will resolve it and exit from the first `if objKind == ""` block.

pkg/controller/utils/datadog/common.go
tbavelier: Done in 1b56504c4cdc0714cfa06064f28b1b4c91496a4c

PR #2244, opened 2025-10-16
[AGENTONB-2588] refactor metadata fowarder

swang392:
### What does this PR do?

* Refactoring `metadata_forwarder.go` into `shared_metadata.go` and `operator_forwarder.go`
    * Shared functionality was moved into `shared_metadata.go`
    * Renamed `metadata_forwarder.go` to `operator_forwarder.go`
* Added unit test to verify payload

### Motivation

Two new metadata forwarders will be added to the operator. By refactoring and moving the shared components (ex: credential management, shared payload values), we can reduce code duplication. 

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Verified that the operator payload after refactor matches the payload sent before refactor. 
Deployed test operator image to staging cluster `squirtle-b` https://github.com/DataDog/k8s-datadog-agent-ops/pull/6602 with `datadogDashboard` CRD enabled. 

Verified that the metadata collected in DDSQL reflected the CRD enabling, but the schema otherwise looks exactly the same. Compared to another cluster on staging without metadata updates (`squirtle-a`)
**squirtle-b**
squirtleb

**squirtle-a** 
squirtle-a

### Checklist

### DIFF
--- a/pkg/controller/utils/metadata/operator_metadata.go
+++ b/pkg/controller/utils/metadata/operator_metadata.go
@@ -23,8 +23,7 @@ import (
 
 const (
 	userAgentHTTPHeaderKey = "User-Agent"
-
-	defaultInterval = 1 * time.Minute
+	defaultInterval        = 1 * time.Minute
 )
 
 type OperatorMetadataForwarder struct {
@@ -158,6 +157,9 @@ func (omf *OperatorMetadataForwarder) setupFromDDA(dda *v2alpha1.DatadogAgent) e
 	return omf.SharedMetadata.setupFromDDA(dda)
 }
 
+// setCredentials attempts to set up credentials and cluster name from the operator configuration first.
+// If cluster name is empty (even when credentials are successfully retrieved from operator),
+// it falls back to setting up from DatadogAgent to ensure we have a valid cluster name.
 func (omf *OperatorMetadataForwarder) setCredentials() error {
 	err := omf.setupFromOperator()
 	if err == nil && omf.clusterName != "" {


pkg/controller/utils/metadata/operator_metadata.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/go-logr/logr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/DataDog/datadog-operator/api/datadoghq/v2alpha1"
+	"github.com/DataDog/datadog-operator/pkg/version"
+)
+
+const (
+	userAgentHTTPHeaderKey = "User-Agent"
+
```
tbavelier:
```suggestion
```
nit: remove empty line

pkg/controller/utils/metadata/operator_metadata.go
```diff
@@ -0,0 +1,179 @@
...
+// setupFromDDA delegates to SharedMetadata setupFromDDA method
+func (omf *OperatorMetadataForwarder) setupFromDDA(dda *v2alpha1.DatadogAgent) error {
+	return omf.SharedMetadata.setupFromDDA(dda)
+}
+
+func (omf *OperatorMetadataForwarder) setCredentials() error {
+	err := omf.setupFromOperator()
+	if err == nil && omf.clusterName != "" {
+		return nil
+	}
```
tbavelier, line 167: Could we add a comment to clarify there (or at the top of the function) the behaviour / my understanding, as in, we won't setup metadata fwder purely from operator if cluster name is empty, despite no error, and instead fallback to DDA ? (which makes sense, we want unique entries for a given cluster, not based on a single node name)

pkg/controller/utils/metadata/operator_metadata.go
```diff
@@ -0,0 +1,179 @@
...
+	Metadata  OperatorMetadata `json:"datadog_operator_metadata"`
+}
+
+type OperatorMetadata struct {
+	OperatorVersion               string `json:"operator_version"`
+	KubernetesVersion             string `json:"kubernetes_version"`
+	InstallMethodTool             string `json:"install_method_tool"`
+	InstallMethodToolVersion      string `json:"install_method_tool_version"`
+	IsLeader                      bool   `json:"is_leader"`
+	DatadogAgentEnabled           bool   `json:"datadogagent_enabled"`
```
tbavelier, line 49:
```suggestion
	DatadogAgentEnabled           bool   `json:"datadogagent_enabled"`
	DatadogAgentInternalEnabled ...
```
should we add DDAI there ? or that's for upcoming PRs as it requires schema change ?

pkg/controller/utils/metadata/operator_metadata_test.go
```diff
@@ -0,0 +1,259 @@
...
+			// Create OperatorMetadataForwarder with the new structure
+			omf := &OperatorMetadataForwarder{
+				SharedMetadata: NewSharedMetadata(zap.New(zap.UseDevMode(true)), nil, "v1.28.0", "v1.19.0"),
+			}
+
+			tt.loadFunc()
+
+			_ = omf.setupFromOperator()
+
+			_ = omf.setupFromDDA(tt.dda)
```
tbavelier, line 149:
```suggestion
			_ = omf.setCredentials()
```
Wouldn't it be more truthful to use `setCredenntials` here to test the behaviour of the function ? Or at least add a `Test_setCredentials` ? (n.b. didn't test if it worked)

pkg/controller/utils/metadata/operator_metadata.go
swang392: I was planning to make a separate PR to update the payload for Operator Metadata since it requires a schema change, which hasn't been merged yet

pkg/controller/utils/metadata/operator_metadata_test.go
swang392: I didn't use setCredentials here because it requires a kubernetes client to run -- so I omitted the step that required the client here (`getDatadogAgent` in shared metadata). Similarly I thought testing `setCredentials` would need to mock or skip that step

PR #2249, opened 2025-10-17
add controllerrevisions collection in ksm check for agents > 7.72.x

frank-spano:
### What does this PR do?

Add rbac for controllerrevisions, and enable controllerrevisions collections in ksm check for agents > 7.72.x

### Motivation

https://github.com/DataDog/datadog-agent/pull/40864

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Deployed operator to minikube, tested with agents 7.71.0, and a few rc's 7.72.0-x. 

For QA, the branch should be checked out  and operator deployed. Then, you'll want to test deploying the agent with a version > 7.72.0, and < 7.72.0. In both cases, the ksm check should run with no errors. If you run `agent config check`, and look in the resources section, you should see controllerrevisions present for agents > 7.72.0
```
kind: DatadogAgent
metadata:
  name: datadog
spec:
  global:
    credentials:
      apiKey: <apikey>
      appKey: <appkey>
  features:
    kubeStateMetricsCore:
      enabled: true
  override:
    nodeAgent:
      image:
       #name: gcr.io/datadoghq/agent:7.71.0
        name: gcr.io/datadoghq/agent:7.72.0-rc.8
      tolerations:
        - operator: Exists
    clusterAgent:
      image:
        #name: gcr.io/datadoghq/agent:7.71.0
        name: gcr.io/datadoghq/agent:7.72.0-rc.8
      tolerations:
        - operator: Exists
```

### Checklist

### DIFF
--- a/internal/controller/datadogagent/feature/kubernetesstatecore/feature.go
+++ b/internal/controller/datadogagent/feature/kubernetesstatecore/feature.go
@@ -22,7 +22,6 @@ import (
 	"github.com/DataDog/datadog-operator/internal/controller/datadogagent/object/volume"
 	"github.com/DataDog/datadog-operator/pkg/constants"
 	"github.com/DataDog/datadog-operator/pkg/controller/utils/comparison"
-	"github.com/DataDog/datadog-operator/pkg/images"
 	"github.com/DataDog/datadog-operator/pkg/kubernetes"
 	"github.com/DataDog/datadog-operator/pkg/utils"
 )
@@ -96,8 +95,8 @@ func (f *ksmFeature) Configure(dda metav1.Object, ddaSpec *v2alpha1.DatadogAgent
 		f.serviceAccountName = constants.GetClusterAgentServiceAccount(dda.GetName(), ddaSpec)
 
 		// Determine CollectControllerRevisions setting
-		// Default to false, then check version requirements
-		f.collectControllerRevisions = false
+		// Default to true, then check version requirements
+		f.collectControllerRevisions = true
 
 		// This check will only run in the Cluster Checks Runners or Cluster Agent (not the Node Agent)
 		if ddaSpec.Features.ClusterChecks != nil && apiutils.BoolValue(ddaSpec.Features.ClusterChecks.Enabled) && apiutils.BoolValue(ddaSpec.Features.ClusterChecks.UseClusterChecksRunners) {
@@ -119,8 +118,8 @@ func (f *ksmFeature) Configure(dda metav1.Object, ddaSpec *v2alpha1.DatadogAgent
 
 					// ControllerRevisions version check - enable if version supports it
 					fallback := false
-					if utils.IsAboveMinVersion(agentVersion, controllerRevisionsCollectionMinVersion, &fallback) {
-						f.collectControllerRevisions = true
+					if !utils.IsAboveMinVersion(agentVersion, controllerRevisionsCollectionMinVersion, &fallback) {
+						f.collectControllerRevisions = false
 					}
 				}
 			}
@@ -137,27 +136,10 @@ func (f *ksmFeature) Configure(dda metav1.Object, ddaSpec *v2alpha1.DatadogAgent
 
 					// ControllerRevisions version check - enable if version supports it
 					fallback := false
-					if utils.IsAboveMinVersion(agentVersion, controllerRevisionsCollectionMinVersion, &fallback) {
-						f.collectControllerRevisions = true
+					if !utils.IsAboveMinVersion(agentVersion, controllerRevisionsCollectionMinVersion, &fallback) {
+						f.collectControllerRevisions = false
 					}
 				}
-			}
-		}
-
-		// If no override was found, check default version based on deployment mode
-		if !f.collectControllerRevisions {
-			// Determine which default version to check based on deployment mode
-			var defaultVersion string
-			if f.runInClusterChecksRunner {
-				defaultVersion = images.AgentLatestVersion
-			} else {
-				defaultVersion = images.ClusterAgentLatestVersion
-			}
-
-			// Check if default version supports controllerrevisions
-			fallback := false
-			if utils.IsAboveMinVersion(defaultVersion, controllerRevisionsCollectionMinVersion, &fallback) {
-				f.collectControllerRevisions = true
 			}
 		}
 
--- a/pkg/utils/version_test.go
+++ b/pkg/utils/version_test.go
@@ -9,188 +9,190 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+
+	apiutils "github.com/DataDog/datadog-operator/api/utils"
 )
 
 func TestIsAboveMinVersion(t *testing.T) {
 	testCases := []struct {
-		name                  string
-		version               string
-		minVersion            string
-		defaultValue          *bool
-		expected              bool
-		description           string
+		name         string
+		version      string
+		minVersion   string
+		defaultValue *bool
+		expected     bool
+		description  string
 	}{
 		{
-			name:        "7.27.0 >= 7.26.0-rc.5",
-			version:     "7.27.0",
-			minVersion:  "7.26.0-rc.5",
+			name:         "7.27.0 >= 7.26.0-rc.5",
+			version:      "7.27.0",
+			minVersion:   "7.26.0-rc.5",
 			defaultValue: nil,
-			expected:    true,
+			expected:     true,
 		},
 		{
-			name:        "7.29.0-rc.5 >= 7.27.0-0",
-			version:     "7.29.0-rc.5",
-			minVersion:  "7.27.0-0",
+			name:         "7.29.0-rc.5 >= 7.27.0-0",
+			version:      "7.29.0-rc.5",
+			minVersion:   "7.27.0-0",
 			defaultValue: nil,
-			expected:    true,
+			expected:     true,
 		},
 		{
-			name:        "7.30.0 >= 7.27.1-0",
-			version:     "7.30.0",
-			minVersion:  "7.27.1-0",
+			name:         "7.30.0 >= 7.27.1-0",
+			version:      "7.30.0",
+			minVersion:   "7.27.1-0",
 			defaultValue: nil,
-			expected:    true,
+			expected:     true,
 		},
 		{
-			name:        "7.25.0 < 7.27.1-0",
-			version:     "7.25.0",
-			minVersion:  "7.27.1-0",
+			name:         "7.25.0 < 7.27.1-0",
+			version:      "7.25.0",
+			minVersion:   "7.27.1-0",
 			defaultValue: nil,
-			expected:    false,
+			expected:     false,
 		},
 		{
-			name:        "6.27.0 < 6.28.0-0",
-			version:     "6.27.0",
-			minVersion:  "6.28.0-0",
+			name:         "6.27.0 < 6.28.0-0",
+			version:      "6.27.0",
+			minVersion:   "6.28.0-0",
 			defaultValue: nil,
-			expected:    false,
+			expected:     false,
 		},
 		{
-			name:        "6.28.1 >= 6.28.0-0",
-			version:     "6.28.1",
-			minVersion:  "6.28.0-0",
+			name:         "6.28.1 >= 6.28.0-0",
+			version:      "6.28.1",
+			minVersion:   "6.28.0-0",
 			defaultValue: nil,
-			expected:    true,
+			expected:     true,
 		},
 		{
-			name:        "6.27.0 < 6.28.0",
-			version:     "6.27.0",
-			minVersion:  "6.28.0",
+			name:         "6.27.0 < 6.28.0",
+			version:      "6.27.0",
+			minVersion:   "6.28.0",
 			defaultValue: nil,
-			expected:    false,
+			expected:     false,
 		},
 		{
-			name:        "foobar unparseable with nil default",
-			version:     "foobar",
-			minVersion:  "7.28.0",
+			name:         "foobar unparseable with nil default",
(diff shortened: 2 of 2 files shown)


hestonhoffman (approved):

tbavelier (changes requested):
Haven't reviewed it (code wise), but tests should be fixed first: https://gitlab.ddbuild.io/DataDog/datadog-operator/-/jobs/1185963278#L641
```shell
        	Error:      	Should be empty, but was resource controllerrevisions does not exist. Available resources: endpoints,persistentvolumeclaims,apiextensions.k8s.io/v1, Resource=customresourcedefinitions,core/v1, Resource=nodes_extended,configmaps,namespaces,poddisruptionbudgets,rolebindings,secrets,volumeattachments,datadoghq.com/v2alpha1, Resource=datadogagents,clusterrolebindings,mutatingwebhookconfigurations,replicasets,roles,storageclasses,certificatesigningrequests,endpointslices,leases,nodes,services,apiregistration.k8s.io/v1, Resource=apiservices,apps/v1, Resource=deployments_extended,horizontalpodautoscalers,ingresses,pods,replicationcontrollers,statefulsets,batch/v1, Resource=jobs_extended,resourcequotas,validatingwebhookconfigurations,core/v1, Resource=pods_extended,clusterroles,cronjobs,daemonsets,jobs,limitranges,ingressclasses,networkpolicies,persistentvolumes,serviceaccounts,autoscaling.k8s.io/v1, Resource=verticalpodautoscalers,apps/v1, Resource=replicasets_extended,deployments
```

Moreover, QA instructions should be added on top of the unit/e2e tests if those cover the addition, thanks!

api/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -833,6 +833,12 @@ type KubeStateMetricsCoreFeatureConfig struct {
 	// +optional
 	// +listType=atomic
 	CollectCrMetrics []Resource `json:"collectCrMetrics,omitempty"`
+
+	// CollectControllerRevisions enables collection of ControllerRevision metrics.
+	// This requires agent version 7.72.0 or later.
+	// Default: false
+	// +optional
+	CollectControllerRevisions *bool `json:"collectControllerRevisions,omitempty"`
```
clamoriniere: **Open question**: Do we want to expose a configuration option that enables the collection of ControllerRevision resources — or should we instead use these resources internally to derive higher-level insights?

Based on our previous discussion, this seems related to exposing rollout duration for `DaemonSets` and `StatefulSets`.

If that’s the case, I wonder whether the new setting should be framed around the benefit or outcome (e.g., enabling rollout duration metrics) rather than the low-level implementation detail (collecting ControllerRevisions).

What do you think? Open to discussion.

frank-spano: @clamoriniere So the collection of DaemonSets and StatefulSets will still occur by default, it's just that the "start time" of the deployment would be less accurate without controllerrevisions access. However, this flag was meant as more of a killswitch than anything else, because enabling this on agent < 7.72.0 breaks the KSM check, or if there was a custom version that was 7.72 but tagged as something non-parsable, we wanted to be able to turn it on manually. I'm open to removing this option and just strictly having it version gated in the feature.go code. 

Also - and this is something I'd probably rather not do (hence being extra careful with the operator and helm-chart deployments) - there's the option of modifying the agent code to NOT fail the ksm check if it's configured to collect controllerrevisions but doesn't have access. We just remove it from the list and continue normal function.

Happy to hear your thoughts

clamoriniere: Thanks for the explanation, that makes sense.

I think a good approach would be to always add the RBAC for `controllerrevisions` when the KSM check is enabled. Then, starting from the next Agent/Cluster-Agent version, we could detect at runtime whether the permission is available. If it isn’t, we simply skip generating the metrics that rely on `controllerrevisions`.

This way, the check can still run successfully without introducing a new configuration parameter, which should provide a smoother experience for users and reduce long-term config maintenance.

Additionally, we could emit a check configuration warning when the RBAC is missing — the check would continue running, but the user would see a warning in the Agent status suggesting they add the missing permission or upgrade their Operator/Helm chart version.

I realize this involves a bit more work, but it would likely result in a better UX and a cleaner setup overall.
let me know what is your opinion?

frank-spano: So based on this PR the rbac will always be enabled, regardless of agent version. As far as the agent code changes, I may have misspoke - newer agents that don't have controllerrevisions enabled will still function, it's just that the start times for statefulsets will be less accurate. The case where the ksm check would break is in older agents < 7.72.0, if we enable collection of controllerrevisions (adding the rbac is fine). Unfortunately we can't fix that in the code, so we just have to make sure it's not enabled for those agents. 

I'm not against the runtime check for resources in the agent, but it won't help us with this release, so I think it can be something we implement in the future?

khewonc: Is there a particular reason for having a config option in the CRD as opposed to automatically enabling controller revisions metric collection based on the agent version? I'd prefer to keep the feature config options simple and limited since a user can always provide a custom config if they want more fine grained control over what's collected

internal/controller/datadogagent/feature/kubernetesstatecore/feature.go
```diff
@@ -67,6 +69,9 @@ type ksmFeature struct {
 // Add "-0" so that prerelase versions are considered sufficient. https://github.com/Masterminds/semver#working-with-prerelease-versions
 const crdAPIServiceCollectionMinVersion = "7.46.0-0"
 
+// Minimum agent version that supports collection of controllerrevisions
+const controllerRevisionsCollectionMinVersion = "7.72.0-0"
```
khewonc: Since this and crdAPIServiceCollectionMinVersion are both about minimum agent versions, I think it makes sense to group them together as constants
```go
const (
  first = "aaa"
  second = "bbb"
)
```

internal/controller/datadogagent/feature/kubernetesstatecore/feature.go
```diff
@@ -78,6 +83,7 @@ func (f *ksmFeature) Configure(dda metav1.Object, ddaSpec *v2alpha1.DatadogAgent
 	var output feature.RequiredComponents
 
 	if ddaSpec.Features != nil && ddaSpec.Features.KubeStateMetricsCore != nil && apiutils.BoolValue(ddaSpec.Features.KubeStateMetricsCore.Enabled) {
+		f.logger.Info("KubeStateMetricsCore feature enabled")
```
khewonc: We already log out enabled features elsewhere. Could you also remove the other log lines? We don't generally log whether a config is set or not

internal/controller/datadogagent/feature/kubernetesstatecore/feature.go
```diff
@@ -88,6 +94,21 @@ func (f *ksmFeature) Configure(dda metav1.Object, ddaSpec *v2alpha1.DatadogAgent
 		f.collectCrMetrics = ddaSpec.Features.KubeStateMetricsCore.CollectCrMetrics
 		f.serviceAccountName = constants.GetClusterAgentServiceAccount(dda.GetName(), ddaSpec)
 
+		// Determine CollectControllerRevisions setting
+		// Priority: 1) Explicit spec setting, 2) Image override version check, 3) Default image version check
```
khewonc: I don't think we need to have complicated logic for this and the behavior is not consistent with other config fields in the CRD. I think we should auto enable based on agent version

pkg/utils/version.go
```diff
@@ -34,6 +35,55 @@ func IsAboveMinVersion(version, minVersion string) bool {
 	return c.Check(v)
 }
 
+// IsAboveMinVersionWithFallback uses semver to check if `version` is >= minVersion.
+// If semver parsing fails, it falls back to parsing major.minor versions only.
+// For versions that can't be parsed at all, it returns false (conservative approach).
+func IsAboveMinVersionWithFallback(version, minVersion string) bool {
```
khewonc: Do we create `X.Y` formatted agent image tags or are there other formats that you're looking to cover with this function that we don't with `IsAboveMinVersion`? We should keep version checking behavior consistent and use a single function since we use version checking multiple times in the operator code. It isn't intuitive if we have two similar functions where one defaults to true and the other false for the same image tag. If there are standard agent image tag formats we're not parsing correctly I think it's better to change `IsAboveMinVersion` directly

khewonc (changes requested):

internal/controller/datadogagent/feature/kubernetesstatecore/feature.go
frank-spano: will group

internal/controller/datadogagent/feature/kubernetesstatecore/feature.go
frank-spano: removing

internal/controller/datadogagent/feature/kubernetesstatecore/feature.go
frank-spano: The reason for this logic, and for having the variable in the config rather than just having it enabled by version is that having the wrong value breaks the ksm check in older agents, and reduces the accuracy in new versions. So we added it as an override if there was a custom image tag or some situation where the version wasn't parsing properly. @sblumenthal Given that we simplified the helm chart to just use the agent version, should we remove this logic and just do a simple version check here as well?

pkg/utils/version.go
frank-spano: This was added for extra sanity, but can be removed if we decide to enable based on version only.

internal/controller/datadogagent/feature/kubernetesstatecore/feature.go
sblumenthal: I am personally okay with removing it... it would remove our ability to explicitly have this part of the check be enabled when deploying an image built off of a branch from the datadog-agent repo, but in reality I don't think that is a scenario we currently find ourselves using

pkg/utils/version.go
frank-spano: Actually on second thought I'd like to leave this function. IsAboveMinVersion(..) returns true if the version is non-semver. In our case, we want to default to false if we can't parse the version, as enabling controllerrevisions in agents < 7.72.0 will break the check, where as the opposite is not as damaging.

frank-spano:
@khewonc pushed some changes if you wouldn't mind having a look

internal/controller/datadogagent/feature/kubernetesstatecore/rbac_test.go
```diff
@@ -288,4 +288,3 @@ func TestGetRBACPolicyRules(t *testing.T) {
 		})
 	}
 }
-
```
lavigne958: [nit]: you're removing the trailing white space here

lavigne958 (approved):

internal/controller/datadogagent/feature/kubernetesstatecore/feature.go
```diff
@@ -97,17 +108,53 @@ func (f *ksmFeature) Configure(dda metav1.Object, ddaSpec *v2alpha1.DatadogAgent
...
+
+					// ControllerRevisions version check - enable if version supports it
+					if utils.IsAboveMinVersionWithFallback(agentVersion, controllerRevisionsCollectionMinVersion) {
+						f.collectControllerRevisions = true
+					}
+				}
+			}
+		}
+
+		// If no override was found, check default version based on deployment mode
```
khewonc: I don't think we need to check for latest version since we can time it so that the operator will release with agent 7.72+ by default

pkg/utils/version.go
khewonc: If the default value of `IsAboveMinVersion` is the issue, then what about adding an arg (maybe a bool pointer?) to `IsAboveMinVersion` that determines the default value? The behavior could be `nil` uses the default value (true), `true` forces the default value to be `true`, and `false` forces the default value to be false

frank-spano: They're different functions outside of the default return value, the new function attempts to parse the major/minor version if it fails the semver check - to cover custom builds where someone appends a suffix that isn't a proper semver.

khewonc: We can't reasonably cover all cases of custom tag formats. I think it's enough that we cover the publicly released tag formats. Other formats would be out of scope for the operator

frank-spano:
@khewonc implemented your suggested change to IsAboveMinVersion

internal/controller/datadogagent/feature/kubernetesstatecore/feature.go
```diff
@@ -97,17 +108,56 @@ func (f *ksmFeature) Configure(dda metav1.Object, ddaSpec *v2alpha1.DatadogAgent
...
 				}
 			}
-		} else if clusterAgentOverride, ok := ddaSpec.Override[v2alpha1.ClusterAgentComponentName]; ok {
-			if clusterAgentOverride.Image != nil && !utils.IsAboveMinVersion(common.GetAgentVersionFromImage(*clusterAgentOverride.Image), crdAPIServiceCollectionMinVersion) {
-				// Disable if image is overridden to an unsupported version
-				f.collectAPIServiceMetrics = false
-				f.collectCRDMetrics = false
+		}
+
+		// If no override was found, check default version based on deployment mode
```
khewonc: @frank-spano The default latest version of the agent is [7.72.1](https://github.com/DataDog/datadog-operator/blob/15325dbc0464fb162c303d942a5cbcd0a7f8e4ec/pkg/images/images.go#L18) as of writing so this check is redundant. Can we delete this section and set `f.collectControllerRevisions = true` in line 100?

PR #2256, opened 2025-10-23
[AGENTONB-2604] (chore) Bump Go version to 1.25.3

tbavelier:
### What does this PR do?

* Bump Go version to 1.25.3
    * Requires bumping openapi-gen: (https://github.com/kubernetes/kube-openapi/commit/7fc278399c7f4ea5c23254a18ee3581d6e20e4e6)
    * Requires bumping controller-gen -> used the closest minor
        * This updated manifests in a no-op manner (remove in the role specific verbs while we're using `*`): https://github.com/kubernetes-sigs/controller-tools/pull/1081 (0.17.0 release)
    * (dev container) Disable `moby` in docker-in-docker feature as it's not compatible with Debian Trixie on which `dev-1.25` is based (An alternative would be to use `dev-1.25-bookworm` tag which is available https://mcr.microsoft.com/v2/devcontainers/go/tags/list): https://github.com/devcontainers/features/blob/main/src/docker-in-docker/NOTES.md#os-support
    > Debian Trixie (13) does not include moby-cli and related system packages, so the feature cannot install with "moby": "true". To use this feature on Trixie, please set "moby": "false" or choose a different base image (for example, Ubuntu 24.04).
    * Requires the workaround from https://github.com/golang/go/issues/75031#issuecomment-3195256688 for e2e-tests otherwise, we run into https://gitlab.ddbuild.io/DataDog/datadog-operator/-/jobs/1193560724#L838
    ```shell
    PASS
    coverage: [no statements]
    ok  	github.com/DataDog/datadog-operator/test/e2e/tests/k8s_suite	1289.267s	coverage: [no statements]
    # github.com/DataDog/datadog-operator/test/e2e/tests/utils
    go: no such tool "covdata"
    make: *** [Makefile:207: e2e-tests] Error 1
    ```

### Motivation

Go 1.25 release is incompatible with a lot of x/tools versions (`0.8.0` to `0.25.0`)/projects using these versions https://github.com/golang/go/issues/74462#issuecomment-3197068769:
```shell
/Users/timothee/go/pkg/mod/golang.org/x/tools@v0.24.0/internal/tokeninternal/tokeninternal.go:64:9: invalid array length -delta * delta (constant -256 of type int64)
```

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -36,7 +36,7 @@ stages:
   - deploy
 
 .on_build_images:
-  - if: '$CI_COMMIT_BRANCH =~ /^mq-working-branch-/'
+  - if: "$CI_COMMIT_BRANCH =~ /^mq-working-branch-/"
     when: never
   - when: always
 
@@ -178,7 +178,6 @@ build_operator_check_image_arm64:
     - IMG_CHECK=$TARGET_IMAGE make docker-build-push-check-ci
     - if [ -n "$CI_COMMIT_TAG" ]; then docker buildx imagetools create $TARGET_IMAGE --tag $RELEASE_IMAGE; fi
 
-
 build_bundle_image:
   stage: image
   tags:
@@ -200,7 +199,7 @@ build_bundle_image:
       compare_to: "refs/heads/main"
     when: never
   # Allow mergequeue branches to merge even if tests fail
-  - if: '$CI_COMMIT_BRANCH =~ /^mq-working-branch-/'
+  - if: "$CI_COMMIT_BRANCH =~ /^mq-working-branch-/"
     allow_failure: true
   - if: $CI_COMMIT_TAG
     allow_failure: true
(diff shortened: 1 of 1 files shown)


config/rbac/role.yaml
```diff
@@ -97,9 +97,6 @@ rules:
   - apiservices
   verbs:
   - '*'
-  - deletecollection
-  - list
-  - watch
```
khewonc, line 102: For my reference, this behavior was changed in https://github.com/kubernetes-sigs/controller-tools/releases/tag/v0.17.0 (I think)

tbavelier: yes this was added in https://github.com/kubernetes-sigs/controller-tools/pull/1081, added to the description

PR #2264, opened 2025-10-27
[AGENTONB-2642] Add checksum on default KSM configmap to restart DCA deployment if updated

tbavelier:
### What does this PR do?

* Computes a checksum annotation for the default KSM ConfigMap based on the user fields

### Motivation

* https://github.com/DataDog/datadog-operator/pull/1927 introduced `collectCrMetrics`. This new field uses the default KSM ConfigMap which doesn’t have a checksum annotation. So when applying a new DDA spec with this field updated, while the ConfigMap is updated, the DCA deploy is not restarted, so it’s not being used until a manual rollout.

### Additional Notes

### Minimum Agent Versions

Are there minimum versions of the Datadog Agent and/or Cluster Agent required? -> not really, while the feature itself needs .63+, this change simply adds the checksum annotation

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

1. Deploy a minimal DDA spec, ensures there's a checksum annotation on the KSM Core ConfigMap
2. Update the DDA spec to include `collectCrMetrics` (use example from https://github.com/DataDog/datadog-operator/pull/1927)
3. Ensure the checksum is updated, and DCA is rolled out

### Checklist

### DIFF
--- a/internal/controller/datadogagent/feature/kubernetesstatecore/feature.go
+++ b/internal/controller/datadogagent/feature/kubernetesstatecore/feature.go
@@ -134,9 +134,9 @@ func (f *ksmFeature) Configure(dda metav1.Object, ddaSpec *v2alpha1.DatadogAgent
 				f.logger.Error(err, "couldn't generate hash for default ksm core config")
 			} else {
 				f.logger.V(2).Info("generated default ksm core config hash", "hash", hash, "config", defaultConfigData)
-				f.customConfigAnnotationValue = hash
-				f.customConfigAnnotationKey = object.GetChecksumAnnotationKey(feature.KubernetesStateCoreIDType)
 			}
+			f.customConfigAnnotationValue = hash
+			f.customConfigAnnotationKey = object.GetChecksumAnnotationKey(feature.KubernetesStateCoreIDType)
 		}
 
 		f.configConfigMapName = constants.GetConfName(dda, f.customConfig, defaultKubeStateMetricsCoreConf)


internal/controller/datadogagent/feature/kubernetesstatecore/feature.go
```diff
@@ -121,6 +121,22 @@ func (f *ksmFeature) Configure(dda metav1.Object, ddaSpec *v2alpha1.DatadogAgent
...
+				"collect_cr_metrics":  f.collectCrMetrics,
+			}
+
+			hash, err := comparison.GenerateMD5ForSpec(defaultConfigData)
+			if err != nil {
+				f.logger.Error(err, "couldn't generate hash for default ksm core config")
+			} else {
+				f.logger.V(2).Info("generated default ksm core config hash", "hash", hash, "config", defaultConfigData)
+				f.customConfigAnnotationValue = hash
+				f.customConfigAnnotationKey = object.GetChecksumAnnotationKey(feature.KubernetesStateCoreIDType)
```
L3n41c: There’s a small difference between this and what happens when `ddaSpec.Features.KubeStateMetricsCore.Conf != nil`:
* Here, we create the hash annotation only if the `comparison.GenerateMD5ForSpec(…)` call was successful whereas
* if the `if` block, when `comparison.GenerateMD5ForSpec(…)` returns an error, we log the error, but we still generate the hash annotation.

Is it on purpose ?

L3n41c (approved):

internal/controller/datadogagent/feature/kubernetesstatecore/feature.go
tbavelier: Thanks, good catch, adressed in 0cbb7d85 to keep the same behaviour

PR #2272, opened 2025-10-28
fix: remove autoMultiLineDetection feature from operator docs

Rookie0031:
### What does this PR do?
Removes references to the `spec.features.logCollection.autoMultiLineDetection` option from the Datadog Operator documentation.  

image

In testing with Datadog Agent **7.70**, this field is not recognized by the Operator CRD and causes a strict decoding error.

### Motivation
I’m running Datadog Agent **7.70** in a test cluster. The Operator documentation currently suggests:

```
spec:
  features:
    logCollection:
      autoMultiLineDetection: true
```

However, applying a DatadogAgent resource with that field fails with:

strict decoding error: unknown field "spec.features.logCollection.autoMultiLineDetection"

image

#### Maybe It still follows the legacy setting

image

But the legacy docs says it applies to version earlier than v7.65

image

### Minimum Agent Versions
Not identified

### Describe your test plan
Describe your test plan

1. Create a DatadogAgent manifest using spec.features.logCollection.autoMultiLineDetection: true.

2. Apply it and observe the Operator/webhook validation error:
`strict decoding error: unknown field "spec.features.logCollection.autoMultiLineDetection"`

### DIFF
--- a/docs/configuration.v2alpha1.md
+++ b/docs/configuration.v2alpha1.md
@@ -133,7 +133,6 @@ spec:
 | features.liveProcessCollection.enabled | Enables Process monitoring. Default: false |
 | features.liveProcessCollection.scrubProcessArguments | ScrubProcessArguments enables scrubbing of sensitive data in process command-lines (passwords, tokens, etc. ). Default: true |
 | features.liveProcessCollection.stripProcessArguments | StripProcessArguments enables stripping of all process arguments. Default: false |
-| features.logCollection.autoMultiLineDetection | AutoMultiLineDetection allows the Agent to detect and aggregate common multi-line logs automatically. See also: https://docs.datadoghq.com/agent/logs/auto_multiline_detection/ |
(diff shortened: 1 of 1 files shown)


tbavelier:
Hello @Rookie0031 ,
On the operator side, this was added in https://github.com/DataDog/datadog-operator/pull/2147 released with 1.19. This is present in the updated CRD as you can see from https://github.com/DataDog/helm-charts/blob/9889fc6432745aabf801cc94985271715e915b18/charts/datadog-crds/templates/datadoghq.com_datadogagents_v1.yaml#L989-L992, so make sure to update the operator and CRD and you should be able to use it.

shloch:
Hello, i have a Cx having this same error on agent **v7.71.2**


`
""2025-11-27 18:27:09 UTC | CORE | WARN | (comp/logs/agent/config/integration_config.go:412 in AutoMultiLineEnabled) | config key "logs_config.experimental_auto_multi_line_detection" is unknown""`

PR #2273, opened 2025-10-28
Add helm metadata forwarder

swang392:
### What does this PR do?

* Adds helm metadata forwader to `pkg/util/metadata` which extends from `SharedMetadata`
* Adds unit tests for helm metadata forwarder methods
* Uses agent's [scrubber](https://github.com/DataDog/datadog-agent/tree/main/pkg/util/scrubber) pkg to scrub sensitive keys, secret backends, etc. from the helm values payload so they will not be surfaced. 

### Motivation

https://datadoghq.atlassian.net/browse/AGENTONB-2594
We want to extend the current metadata forwarding schema in the Operator to also collect metadata on helm values configurations. This is necessary for future work in Agent Onboarding and Fleet Automation. 
[RFC](https://docs.google.com/document/d/1nG_bxOsOVG83o_0_VjI9hTFsuDxzokSu-1uv_-dySWQ/edit?usp=sharing)

### Additional Notes

Implementation is modeled off of https://github.com/DataDog/datadog-agent/pull/37860/files

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan
**Local Testing**
* Installed multiple helm charts (`datadog` and `datadog-operator`) Make sure to install the datadog-operator chart with the `debug` loglevel to verify the payload, which is logged at the debug level. 
```
helm install datadog ~/dd/helm-charts/charts/datadog
helm upgrade datadog-operator datadog/datadog-operator --set image.tag=<custom-image> --set image.doNotCheckTag=true --set image.repository=operator --set logLevel="debug" --set apiKey=xxxxxxx --set datadog.secretBackend.command="/readsecret.sh"
```
* verified that helm metadata forwarder started, and payload logs are visible for all helm releases. The payload contains both the provided configuration and full values.yaml. 
     * Since Operator metadata payloads are logged, look specifically for the helm release names for the charts you installed, and the name of the helm payload: `datadog_operator_helm_metadata\`
* check that secrets are scrubbed: 
```
"helm_provided_configuration\":\"apiKey: \\\"***************************xxxxx\\\"\\ndatadog:\\n  secretBackend:\\n    command: /readsecret.sh\\nimage:\\n  doNotCheckTag: true\\n  repository: operator\\n  tag: metadata-16\\nlogLevel: debug\"
```
Example logs: 

**E2E Testing**
I also tested on a staging cluster after rolling out my dd-go feature flag. Internally collecting the `datadog-operator` and `datadog-agent` chart

### Checklist

### DIFF
--- a/pkg/controller/utils/metadata/helm_metadata_test.go
+++ b/pkg/controller/utils/metadata/helm_metadata_test.go
@@ -12,6 +12,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/DataDog/datadog-operator/pkg/config"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
@@ -27,7 +28,7 @@ func Test_HelmMetadataForwarder_getPayload(t *testing.T) {
 	expectedChartVersion := "3.10.0"
 	expectedAppVersion := "7.50.0"
 
-	hmf := NewHelmMetadataForwarder(zap.New(zap.UseDevMode(true)), nil, expectedKubernetesVersion, expectedOperatorVersion)
+	hmf := NewHelmMetadataForwarder(zap.New(zap.UseDevMode(true)), nil, expectedKubernetesVersion, expectedOperatorVersion, config.NewCredentialManager())
 
 	// Set required fields
 	hmf.hostName = expectedHostname
@@ -131,7 +132,7 @@ func Test_HelmMetadataForwarder_getPayload(t *testing.T) {
 }
 
 func Test_parseHelmResource(t *testing.T) {
-	hmf := NewHelmMetadataForwarder(zap.New(zap.UseDevMode(true)), nil, "v1.28.0", "v1.19.0")
+	hmf := NewHelmMetadataForwarder(zap.New(zap.UseDevMode(true)), nil, "v1.28.0", "v1.19.0", config.NewCredentialManager())
 
 	// Create a minimal valid Helm release JSON
 	releaseData := HelmReleaseMinimal{
@@ -277,7 +278,7 @@ func Test_allHelmReleasesCache(t *testing.T) {
 }
 
 func Test_mergeValues(t *testing.T) {
-	hmf := NewHelmMetadataForwarder(zap.New(zap.UseDevMode(true)), nil, "v1.28.0", "v1.19.0")
+	hmf := NewHelmMetadataForwarder(zap.New(zap.UseDevMode(true)), nil, "v1.28.0", "v1.19.0", config.NewCredentialManager())
 
 	tests := []struct {
 		name      string


pkg/controller/utils/metadata/helm_metadata.go
```diff
@@ -0,0 +1,521 @@
...
+				"namespace", release.Namespace)
+			sendErrors = append(sendErrors, err)
+		} else {
+			hmf.logger.V(1).Info("Successfully sent Helm metadata",
+				"release", release.ReleaseName,
+				"namespace", release.Namespace)
+		}
+	}
+
+	if len(sendErrors) > 0 {
```
fanny-jiang, line 190: Not a blocker, but maybe we could consider sending telemetry on sendErrors. I.e. if the errors are not network-related, it would be nice to understand why metadata send requests fail

pkg/controller/utils/metadata/helm_metadata.go
```diff
@@ -0,0 +1,521 @@
...
+		if err != nil {
+			hmf.logger.V(1).Info("Failed to scrub provided values, using unscrubbed", "release", data.release.Name, "error", err)
+			scrubbedProvidedYAML = providedValuesYAML
+		}
+
+		scrubbedFullYAML, err := scrubber.ScrubBytes(fullValuesYAML)
+		if err != nil {
+			hmf.logger.V(1).Info("Failed to scrub full values, using unscrubbed", "release", data.release.Name, "error", err)
+			scrubbedFullYAML = fullValuesYAML
+		}
```
fanny-jiang, line 433: If scrubbing fails, are we still logging the unscrubbed values in debug mode?

swang392: Ah yeah, I made a fix so if the scrubber fails, we won't log the payload. For example:
```
{"level":"DEBUG","ts":"2025-11-03T22:58:21.794Z","logger":"metadata","msg":"Built Helm metadata payload (not scrubbed)","release":"datadog-operator","namespace":"default","chart":"datadog-operator","payload_size":2564}
```

pkg/controller/utils/metadata/helm_metadata.go
```diff
@@ -0,0 +1,534 @@
...
+// Start starts the helm metadata forwarder
+func (hmf *HelmMetadataForwarder) Start() {
+	err := hmf.setCredentials()
+	if err != nil {
+		hmf.logger.Error(err, "Could not set credentials; not starting helm metadata forwarder")
+		return
+	}
+
+	if hmf.hostName == "" {
+		hmf.logger.Error(ErrEmptyHostName, "Could not set host name; not starting helm metadata forwarder")
```
fanny-jiang, line 136: Can we add a suggestion here to set the `DD_CLUSTER_NAME` env var or `clusterName` if using helm chart? 

On that same note, we should update the public docs to include that in the installation instructions.

pkg/controller/utils/metadata/helm_metadata.go
```diff
@@ -0,0 +1,534 @@
...
+
+// Start starts the helm metadata forwarder
+func (hmf *HelmMetadataForwarder) Start() {
+	err := hmf.setCredentials()
+	if err != nil {
+		hmf.logger.Error(err, "Could not set credentials; not starting helm metadata forwarder")
+		return
+	}
+
+	if hmf.hostName == "" {
```
fanny-jiang, line 135: Should this be `hmf.clusterName`?

cmd/main.go
```diff
@@ -521,3 +522,8 @@ func setupAndStartOperatorMetadataForwarder(logger logr.Logger, client client.Re
 
 	omf.Start()
 }
+
+func setupAndStartHelmMetadataForwarder(logger logr.Logger, client client.Reader, kubernetesVersion string, options *options, credsManager *config.CredentialManager) {
+	hmf := metadata.NewHelmMetadataForwarder(logger, client, kubernetesVersion, version.GetVersion(), credsManager)
+	hmf.Start()
+}
```
fanny-jiang, line 529: Is it possible to call this function again when a DDA is created? 

When the operator is first deployed without credentials or a clusterName, the hmf doesn't get started and it doesn't get started when a valid DDA is created unless the operator pod is deleted and restarted. Not a blocker for this PR, but it would be nicer UX

pkg/controller/utils/metadata/helm_metadata.go
```diff
@@ -0,0 +1,534 @@
...
+		}
+	}
+
+	// If both Secrets and ConfigMaps failed to list, return error
+	if secretListErr != nil && cmListErr != nil {
+		return nil, fmt.Errorf("failed to discover Helm releases: secrets error: %w, configmaps error: %w", secretListErr, cmListErr), notScrubbed
+	}
+
+	releases := make([]HelmReleaseData, 0, len(latestReleases))
+	for _, data := range latestReleases {
```
fanny-jiang, line 406: Just curious, is it possible for a single helm release to appear in both the secret and configMap?

pkg/controller/utils/metadata/helm_metadata.go
fanny-jiang: discussed offline, will be refactored in a future PR

fanny-jiang (approved):

pkg/controller/utils/metadata/helm_metadata.go
swang392: No, it depends on the helm version (3+ is secret, older is configmap)

pkg/controller/utils/metadata/helm_metadata.go
swang392: gonna make separate PR to update doc

cmd/main.go
swang392: I think this [PR](https://github.com/DataDog/datadog-operator/pull/2295) will help resolve -- instead of setting credentials only once, moving `setCredentials` to `sendMetadata` will check if credentials can be set more frequently, allowing time for DDA to be instantiated

PR #2292, opened 2025-11-05
Build requirements once per profile, not for every node

khewonc:
### What does this PR do?

Refactor requirement generation from profile affinity and node matching so that we generate requirements once per profile instead of generating them for each node. Also separates out requirement generation and node matching into separate functions to help with a future refactor

### Motivation

### Additional Notes

Performance impact is minimal since the previous behavior didn't use a significant amount of resources

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

There should be no functional change for existing profiles. Can be confirmed in staging by ensuring the number of nodes with profiles on them doesn't change after this change

### Checklist

### DIFF
--- a/pkg/agentprofile/agent_profile.go
+++ b/pkg/agentprofile/agent_profile.go
@@ -75,7 +75,7 @@ func ApplyProfile(logger logr.Logger, profile *v1alpha1.DatadogAgentProfile, nod
 	toLabelNodeCount := 0
 
 	// Parse profile requirements once to avoid recreating them for each node
-	profileRequirements, err := ParseProfileRequirements(profile)
+	profileRequirements, err := parseProfileRequirements(profile)
 	if err != nil {
 		logger.Error(err, "profile selector is invalid, skipping", "datadogagentprofile", profile.Name, "datadogagentprofile_namespace", profile.Namespace)
 		metrics.DAPValid.With(prometheus.Labels{"datadogagentprofile": profile.Name}).Set(metrics.FalseValue)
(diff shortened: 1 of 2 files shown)


pkg/agentprofile/agent_profile.go
```diff
@@ -73,16 +73,19 @@ func ApplyProfile(logger logr.Logger, profile *v1alpha1.DatadogAgentProfile, nod
 
 	toLabelNodeCount := 0
 
+	// Parse profile requirements once to avoid recreating them for each node
+	profileRequirements, err := ParseProfileRequirements(profile)
```
levan-m: nit - why are we exporting this? don't see it used elsewhere

pkg/agentprofile/agent_profile.go
```diff
@@ -369,27 +372,40 @@ func SortProfiles(profiles []v1alpha1.DatadogAgentProfile) []v1alpha1.DatadogAge
...
+	}
+	return requirements, nil
+}
 
-		if !selector.Matches(labels.Set(nodeLabels)) {
-			return false, nil
-		}
+// profileMatchesNodeWithRequirements checks if a node matches the given requirements
+func profileMatchesNodeWithRequirements(requirements []*labels.Requirement, nodeLabels map[string]string) bool {
+	if requirements == nil || len(requirements) == 0 {
```
levan-m:
```suggestion
	if len(requirements) == 0 {
```
nit - no need for nil check

levan-m (approved):

PR #2317, opened 2025-11-17
Use otel agent standalone image

mackjmr:
### What does this PR do?

This PR updates the OTel agent to use the standalone image. If the user overrides the version tag to a tag incompatible with the standalone image, the otel agent feature emits an error log and errors out.
### Motivation

### Additional Notes

When testing this PR, OTel Agent will crash. Make changes in #2326 to fix.

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

1. Default

Run the otel agent with no overrides.

Ensure all container images are `gcr.io/datadoghq/agent:7.72.1`, but otel-agent image is `gcr.io/datadoghq/ddot-collector:7.72.1`

2. Tag override to supported version

Run the otel agent with override tag:

```
spec:
  override:
    nodeAgent:
      image:
        tag: 7.71.0
```

Ensure all container images are `gcr.io/datadoghq/agent:7.71.0`, but otel-agent image is `gcr.io/datadoghq/ddot-collector:7.71.0`

3. Tag override to unsupported version

Run the otel agent with override tag:

```
spec:
  override:
    nodeAgent:
      image:
        tag: 7.66.0
```

Ensure otel agent container is not run and operator contains error log:
`"msg":"OTel Agent Standalone image requires agent version 7.67.0 or higher. Update the Agent version or use the agent image with -full tag instead."`

4. Tag override to unsupported full version

Run the otel agent with override tag:

```
spec:
  override:
    nodeAgent:
      image:
        tag: 7.71.1-full
```

Ensure otel agent container is not run and operator contains error log:
`"msg":"OTel Agent Standalone image does not support full tag. Update the Agent to version without -full tag, or use the agent image with -full tag instead."`

5. JMX override

Run the otel agent with override jmx:

```
spec:
  override:
    nodeAgent:
      image:
        jmxEnabled: true
```

Ensure otel agent container container image is: `gcr.io/datadoghq/ddot-collector:7.72.1` (no `-jmx` tag).

### Checklist


internal/controller/datadogagent/component/agent/default.go
```diff
@@ -337,9 +337,8 @@ func agentImage() string {
 	return images.GetLatestAgentImage()
 }
 
-func fullAgentImage() string {
-	return images.GetLatestAgentImageWithSuffix(false, false, true)
-
+func ddotCollectorImage() string {
```
swang392, line 340: I think this only works when there’s no image-name override. `override/podtemplatespec.go` applies `.spec.override["nodeAgent"].image` to the otel container, so if a user sets that field, the otel-agent container gets rewritten to the agent image and  `ddot-collector` is never used. I tested with the snippet below and the otel container fell back to the agent image/CLBO. Could reapply the collector image afterward in the override?

```
override:
    nodeAgent:
    image:
        name: agent
        tag: 7.71.0
```

internal/controller/datadogagent/feature/otelcollector/feature.go
```diff
@@ -68,6 +74,22 @@ func (o *otelCollectorFeature) ID() feature.IDType {
...
+	var agentImageName string
+	agentVersion := images.AgentLatestVersion
+	if nodeAgent, ok := ddaSpec.Override[v2alpha1.NodeAgentComponentName]; ok {
+		if nodeAgent.Image != nil {
+			agentImageName = nodeAgent.Image.Name
+			agentVersion = common.GetAgentVersionFromImage(*nodeAgent.Image)
+		}
+	}
+	supportedVersion := utils.IsAboveMinVersion(agentVersion, "7.67.0-0", apiutils.NewBoolPointer(true))
+	if !supportedVersion && agentImageName == "" {
```
swang392, line 96: Why does agentImageName need to be empty to gate this feature? If a user sets the image name the error will never get reached.

swang392:
`Test_otelImageTags` in `internal/controller/datadogagent/controller_v2_test.go` still expects the previous nodeAgent image name, not the standalone image. Could you update it to check for the new standalone image and also add coverage for the override error with unsupported versions?

internal/controller/datadogagent/component/agent/default.go
mackjmr: That is correct. It was my understanding that the overrides apply post feature logic, so any updates to the image in the feature will be overriden by the override statements. 

I think if the user chooses to override the image name, they should understand what they are overriding it to (and I plan on adding documentation to make this easier to understand which image to use for otel agent). 
If they update just the tag, because we are changing the image name, we can error out and let the user know that this is an incorrect tag for our new standalone image.

Also, I think it can create confusion if the user sets an override, and then we don't take it into account and re-override it with our own logic.

internal/controller/datadogagent/feature/otelcollector/feature.go
mackjmr: Answered in https://github.com/DataDog/datadog-operator/pull/2317#discussion_r2549516349. We help user if they set incompatible version, but if they override name, they should understand what they are overriding it to. Let me know if you disagree.

mackjmr:
@swang392 Updated the test, thanks. I also added some logic to error out on full image tag for standalone image.

pkg/images/images.go
tbavelier, line 1: Please remove `GetLatestAgentImageWithSuffix` since it's not used anymore following the changes in `default.go`.
Also remove the reference to full, e.g. `WithFull` function, in the struct, etc.

internal/controller/datadogagent/feature/otelcollector/feature.go
```diff
@@ -68,6 +74,29 @@ func (o *otelCollectorFeature) ID() feature.IDType {
...
+		o.incompatibleImage = true
+		o.logger.Error(errIncompatibleImage, "OTel Agent Standalone image requires agent version 7.67.0 or higher. Update the Agent version or use the agent image with -full tag instead.",
+			"current_version", agentVersion)
+		return feature.RequiredComponents{}
+	}
+
+	if strings.HasSuffix(agentVersion, "-full") && agentImageName == "" {
+		o.incompatibleImage = true
+		o.logger.Error(errIncompatibleImage, "OTel Agent Standalone image does not support full tag. Update the Agent to version without -full tag, or use the agent image with -full tag instead.",
```
tbavelier: I don't think this error message is really helpful/clear, I think the intent is to ask users to use a FQDN override, e.g. `override.nodeAgent.image.name: registry.domain/agent:version-full` so all containers in the Agent pod use the `-full` image ?

internal/controller/datadogagent/feature/otelcollector/feature.go
```diff
@@ -68,6 +74,29 @@ func (o *otelCollectorFeature) ID() feature.IDType {
...
+	if nodeAgent, ok := ddaSpec.Override[v2alpha1.NodeAgentComponentName]; ok {
+		if nodeAgent.Image != nil {
+			agentImageName = nodeAgent.Image.Name
+			agentVersion = common.GetAgentVersionFromImage(*nodeAgent.Image)
+		}
+	}
+	supportedVersion := utils.IsAboveMinVersion(agentVersion, "7.67.0-0", apiutils.NewBoolPointer(true))
+	if !supportedVersion && agentImageName == "" {
+		o.incompatibleImage = true
+		o.logger.Error(errIncompatibleImage, "OTel Agent Standalone image requires agent version 7.67.0 or higher. Update the Agent version or use the agent image with -full tag instead.",
```
tbavelier: Same as below "Update the Agent version or use the agent image with -full tag instead." is not clear/how it should be used

tbavelier (changes requested):
This will break users using `jmxEnabled` in override (https://docs.datadoghq.com/containers/guide/autodiscovery-with-jmx/?tab=operator#use-a-jmx-enabled-agent), as we'll try to pull `ddot-collector:7.72.1-jmx` while previoulsy, the `-full` suffix would take precedence and include `jmx`. Some logic for this should be added in `PodTemplateSpec` in overrides, e.g. something like
image
The above change would still not fix users that specify a FQDN image `gcr.io/datadoghq/agent:7.72.2`, they'll still need to use the `-full` in their FQDN but that's expected / we are aligned on this

pkg/images/images.go
mackjmr: I removed `GetLatestAgentImageWithSuffix`. I don't think I can remove the rest, as its used when the agent is overriden with full tag. For example: [This](https://github.com/DataDog/datadog-operator/blob/b524f0dbd5732d20af1950a8f29f7ab285e31376/pkg/images/images.go#L170) calls [fromImageConfig](https://github.com/DataDog/datadog-operator/blob/b524f0dbd5732d20af1950a8f29f7ab285e31376/pkg/images/images.go#L251) which uses the [WithFull](https://github.com/DataDog/datadog-operator/blob/b524f0dbd5732d20af1950a8f29f7ab285e31376/pkg/images/images.go#L293-L298) function / struct

internal/controller/datadogagent/feature/otelcollector/feature.go
mackjmr: The intent is to tell the customer to:
- use a tag without `-full`
- OR use a FQDN override indeed.

I've updated the wording, lmk what you think

internal/controller/datadogagent/feature/otelcollector/feature.go
mackjmr: I've updated the wording, lmk what you think

pkg/images/images.go
tbavelier: Indeed, let's keep it so we don't end up with `full-jmx` as removing it would break these users, thanks!

tbavelier (approved):

PR #2323, opened 2025-11-18
Add missing system-probe cws volume mount points (bis)

spikat:
### What does this PR do?

This is the reaply of [this PR](https://github.com/DataDog/datadog-operator/pull/2308), which has be reverted [here](https://github.com/DataDog/datadog-operator/pull/2316) because it caused issues related with this log:

> Error: failed to create containerd task: failed to create shim task: OCI runtime
> create failed: runc create failed: unable to start container process: error during
> container init: error mounting "/var/run/containerd/containerd.sock" to rootfs at
> "/host/var/run": mount src=/var/run/containerd/containerd.sock, dst=/host/var/run,
> dstFd=/proc/thread-self/fd/62, flags=MS_RDONLY|MS_BIND|MS_REC: not a directory

The issue was that he commit adds both a HostRoot mount (/ → /host/root) and a RuntimeSocket mount (/var/run → /host/var/run), which conflict because /host/var/run already exists within the /host/root mount. Removing GetVolumeMountForRuntimeSocket(true) (should) fix the issue.

The QA was done by deploying the operator version to empoleon-a: https://gitlab.ddbuild.io/DataDog/k8s-datadog-agent-ops/-/pipelines/83181276 

Initial PR description:

> Many customers complained about system-probe log errors when enabling only workload protection (formerly CWS). This PR adds the missing mounts for CWS (it should have already been the case, I'm not sure when the regression happened):
> 
> /host/root
> /host/sys/fs/cgroup
> Also, remove /sys/fs/security mount which can cause issues on systems where it's not present
> 
> Changes has been QA using minikube, enabling only CWS and double checking that everything started fine and looks good with no error logs

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist


lebauce:
@spikat GetVolumeMountForRuntimeSocket seems to be still there. Am I missing something ?

PR #2336, opened 2025-11-20
Add support of vsock for intra agents communications

lebauce:
### What does this PR do?

Kata containers makes use of microVM to run pods. In these environments, network may not be available at all. vsock allows communication between the host and the guest through the hypervisor, without using TCP sockets.

https://github.com/DataDog/datadog-agent/pull/39478 added support for vsock sockets in the agent.

This PR modifies the Helm Charts by:
- setting the right configs (DD_VSOCK_ADDR, ...) to enable the feature
- set the /etc/datadog-agent/auth as a hostpath so that it can be shared with the microVM through virtiofs

### Motivation

Some environments do not allow workload to have network interfaces. Using `vsock` sockets
allow these workloads to communicate with the agent.

### Additional Notes

### Minimum Agent Versions

* Agent: v7.74.0
* Cluster Agent:  v7.74.0

### Describe your test plan

### Checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -724,7 +724,8 @@ check_preflight_redhat_image:
     - when: never
   needs:
     - "publish_redhat_public_tag"
-  tags: ["runner:docker", "size:large"]
+  tags:
+    - "docker-in-docker:amd64"
   image: $JOB_DOCKER_IMAGE
   script:
     - apt-get update && apt-get -y install --no-install-recommends build-essential git awscli && apt-get -y clean && rm -rf /var/lib/apt/lists/*
@@ -743,7 +744,8 @@ submit_preflight_redhat_image:
     - when: never
   needs:
     - "publish_redhat_public_tag"
-  tags: ["runner:docker", "size:large"]
+  tags:
+    - "docker-in-docker:amd64"
   image: $JOB_DOCKER_IMAGE
   script:
     - apt-get update && apt-get -y install --no-install-recommends build-essential git awscli && apt-get -y clean && rm -rf /var/lib/apt/lists/*
@@ -762,7 +764,8 @@ publish_community_operators:
   # preflight submission can be retried locally followed by this manual job trigger in the pipeline
   # needs:
   #   - "submit_preflight_redhat_image"
-  tags: ["runner:docker", "size:large"]
+  tags:
+    - "docker-in-docker:amd64"
   image: $OCTO_JOB_DOCKER_IMAGE
   id_tokens:
     DDOCTOSTS_ID_TOKEN:
--- a/api/datadoghq/common/datadogpodautoscaler_types.go
+++ b/api/datadoghq/common/datadogpodautoscaler_types.go
@@ -24,7 +24,7 @@ const (
 )
 
 // DatadogPodAutoscalerUpdateStrategy defines the mode of the update policy.
-// +kubebuilder:validation:Enum:=Auto;Disabled
+// +kubebuilder:validation:Enum:=Auto;Disabled;TriggerRollout
 type DatadogPodAutoscalerUpdateStrategy string
 
 const (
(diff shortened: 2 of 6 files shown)


internal/controller/datadogagent/component/agent/default.go
```diff
@@ -32,22 +32,22 @@ import (
 
 // NewDefaultAgentDaemonset return a new default agent DaemonSet
 // TODO: remove instanceName once v2 reconcile is removed
-func NewDefaultAgentDaemonset(dda metav1.Object, edsOptions *ExtendedDaemonsetOptions, agentComponent feature.RequiredComponent, instanceName string) *appsv1.DaemonSet {
+func NewDefaultAgentDaemonset(dda metav1.Object, ddaSpec *v2alpha1.DatadogAgentSpec, edsOptions *ExtendedDaemonsetOptions, agentComponent feature.RequiredComponent, instanceName string) *appsv1.DaemonSet {
```
clamoriniere: why do we need to provide now the ddaSpec when we already provide the `dda metav1.Object` as first parameter of the function?

clamoriniere: if you look at https://github.com/DataDog/datadog-operator/blob/a312b889282f7c4b87d1b8f817a4ec7c765d75c0/internal/controller/datadogagent/controller_reconcile_agent.go#L142
the dda parameter type is  `*datadoghqv2alpha1.DatadogAgent`  https://github.com/DataDog/datadog-operator/blob/a312b889282f7c4b87d1b8f817a4ec7c765d75c0/internal/controller/datadogagent/controller_reconcile_agent.go#L43

lebauce: For this one https://github.com/DataDog/datadog-operator/blob/main/internal/controller/datadogagentinternal/controller_reconcile_agent.go#L35, dda is `ddai *datadoghqv1alpha1.DatadogAgentInternal` - which also has a `Spec v2alpha1.DatadogAgentSpec` field. What we could do is change it to an interface that would have a function GetSpec that would return a Spec v2alpha1.DatadogAgentSpec but I'm not sure it's worth it. WDYT ?

tbavelier: We'll eventually drop support for the legacy DDA without DDAI path in 1.27, so we'll be able to reconcile a single dda/ddai without having to have to split between the meta and the spec, as it was before. This "pattern" (meta, spec) is used everywhere temporarily while we have the 2 paths, so it should be outside the scope of this PR

internal/controller/datadogagent/common/volumes.go
```diff
@@ -57,7 +57,18 @@ func GetVolumeForRmCorechecks() corev1.Volume {
 }
 
 // GetVolumeForAuth return the Volume container authentication information
-func GetVolumeForAuth() corev1.Volume {
+func GetVolumeForAuth(hostPath bool) corev1.Volume {
```
tbavelier, line 60: See be6c54095279313ee66b41f24cd6265029057fe0 : unlike the other hostPath from this file, we have no guarantee `/etc/datadog-agent/auth` is present on the host, so we should ensure it's `DirectoryOrCreate`: https://kubernetes.io/docs/concepts/storage/volumes/#hostpath

internal/controller/datadogagent/global/global.go
```diff
@@ -290,6 +290,20 @@ func applyGlobalSettings(logger logr.Logger, manager feature.PodTemplateManagers
...
+			Name:  DDVSockAddr,
+			Value: "host",
+		})
+
+		// Remote agent doesn't work with vsock yet
+		manager.EnvVar().AddEnvVar(&corev1.EnvVar{
+			Name:  DDRemoteAgentRegistryEnabled,
+			Value: "false",
+		})
+	}
```
tbavelier: Since this applies solely from my understanding to node Agent, it's better suited to be moved in `applyNodeAgentResources` in `internal/controller/datadogagent/global/agent.go`. This also means we don't have to touch the signature of the default EDS/DS functions. See 6d158f72df3f70554870934a419c12cdc6c2d1c6

tbavelier (changes requested):
Wondering if this communication mode is CWS/some specific features only, in which case, all the code could be moved to said features instead of `spec.global` ? Features are applied after global, so they would take priority over the default for the auth volume ensuring it's hostPath instead of emptyDir. This would make it even cleaner/scoped.
If it's meant to be a real global for all sub-Agents in a supported communication mode (they'll all communicate to core Agent using this socket), then it's fine to keep in `spec.global` but it should be moved as per the comments below: see https://github.com/DataDog/datadog-operator/pull/2786

internal/controller/datadogagent/global/global.go
lebauce: :+1: Done

internal/controller/datadogagent/common/volumes.go
lebauce: :+1: Done

lebauce:
> Wondering if this communication mode is CWS/some specific features only, in which case, all the code could be moved to said features instead of `spec.global` ? Features are applied after global, so they would take priority over the default for the auth volume ensuring it's hostPath instead of emptyDir. This would make it even cleaner/scoped. If it's meant to be a real global for all sub-Agents in a supported communication mode (they'll all communicate to core Agent using this socket), then it's fine to keep in `spec.global` but it should be moved as per the comments below: see #2786

This is not specific to CWS (you may want to use the vsock mode without CWS at all). So I left it in `spec.global` as suggested

internal/controller/datadogagent/global/global.go
tbavelier: If you see the draft PR I made, we don't need the signature changes to the defaultpodtempaltespec, the default shouldn't know what's configured in the global. We simply need to move the volume to global:
```
authVol := common.GetVolumeForAuth(true)
manager.Volume().AddVolume(&authVol)
```
My PR compared to your head now looks weird, but this simple change allows to avoid changing all the defaults

PR #2351, opened 2025-11-24
Metadata forwarder creds refactor

levan-m:
### What does this PR do?

First four commits are relatively small refactor preparing for refactor in last commit.

* [7c0e134c](https://github.com/DataDog/datadog-operator/commit/b0f375a0647960bf76d0cfb0ad288ea9597793cd) functional change, moves credentials resolution from `Start` to `sendMetadata` and adds tests to assert current behavior.
* [e603fd0d](https://github.com/DataDog/datadog-operator/commit/08ec6cbec30c409eb62a7be4fb9d74fde7d19083) should be minimal functional change moving DDA creds/url resolution to `creds.go`.
* [f9da087c](https://github.com/DataDog/datadog-operator/commit/f91ee3759b2ea51c338b9a2b10274cde25c5f9fa) Add basically non-functional change, making `creds.go` easier to test and adding test for DDA code path.
* [35e12685](https://github.com/DataDog/datadog-operator/commit/f4fc1e566970dac7a0b5bd5b1c2891eb1dce93e3) dropped `apiKey` in `shared_metadata.go`, getting it on-demand vs setting in the field.
* [e3d47079](https://github.com/DataDog/datadog-operator/commit/d17b446ea370aa6780997c7660b0c8551b56c609) consolidate http request logic; reduce shared state/duplicate cache; adjusted some of the tests.

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: Build
 env:
-  GO_VERSION: 1.25.3
+  GO_VERSION: 1.25.5
 on:
   push:
 # Permission forced by repo-level setting; only elevate on job-level
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,7 +6,7 @@ permissions:
   # packages: read
 env:
   PROJECTNAME: "datadog-operator"
-  GO_VERSION: 1.25.3
+  GO_VERSION: 1.25.5
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -11,7 +11,7 @@ permissions:
   contents: read
   # packages: read
 env:
-  GO_VERSION: 1.25.3
+  GO_VERSION: 1.25.5
 jobs:
   build-linux-binary:
     if: startsWith(github.ref, 'refs/tags/v')
--- a/.github/workflows/yaml-mapper.yml
+++ b/.github/workflows/yaml-mapper.yml
@@ -12,7 +12,7 @@ permissions:
   # packages: read
 env:
   PROJECTNAME: "datadog-operator"
-  GO_VERSION: 1.25.3
+  GO_VERSION: 1.25.5
 jobs:
   yaml-mapper-test:
     runs-on: ubuntu-latest
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,4 +1,4 @@
-image: registry.ddbuild.io/images/mirror/library/golang:1.25.3
+image: registry.ddbuild.io/images/mirror/library/golang:1.25.5
 variables:
   PROJECTNAME: "datadog-operator"
   PROJECTNAME_CHECK: "datadog-operator-check"
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 ARG FIPS_ENABLED=false
 
 # Build the manager binary
-FROM golang:1.25.3 AS builder
+FROM golang:1.25.5 AS builder
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
--- a/api/datadoghq/v1alpha1/datadoggenericresource_types.go
+++ b/api/datadoghq/v1alpha1/datadoggenericresource_types.go
@@ -13,6 +13,7 @@ type SupportedResourcesType string
 
 // When adding a new type, make sure to update the kubebuilder validation enum marker
 const (
+	Downtime              SupportedResourcesType = "downtime"
 	Monitor               SupportedResourcesType = "monitor"
 	Notebook              SupportedResourcesType = "notebook"
 	SyntheticsAPITest     SupportedResourcesType = "synthetics_api_test"
@@ -23,7 +24,7 @@ const (
 // +k8s:openapi-gen=true
 type DatadogGenericResourceSpec struct {
 	// Type is the type of the API object
-	// +kubebuilder:validation:Enum=monitor;notebook;synthetics_api_test;synthetics_browser_test
+	// +kubebuilder:validation:Enum=downtime;monitor;notebook;synthetics_api_test;synthetics_browser_test
 	Type SupportedResourcesType `json:"type"`
 	// JsonSpec is the specification of the API object
 	JsonSpec string `json:"jsonSpec"`
--- a/api/datadoghq/v1alpha1/datadoggenericresource_validation.go
+++ b/api/datadoghq/v1alpha1/datadoggenericresource_validation.go
@@ -12,6 +12,7 @@ import (
 )
 
 var allowedCustomResourcesEnumMap = map[SupportedResourcesType]string{
+	Downtime:              "",
 	Monitor:               "",
 	Notebook:              "",
 	SyntheticsAPITest:     "",
--- a/api/go.mod
+++ b/api/go.mod
@@ -2,7 +2,7 @@ module github.com/DataDog/datadog-operator/api
 
 go 1.25
 
-toolchain go1.25.3
+toolchain go1.25.5
 
 require (
 	github.com/DataDog/datadog-api-client-go/v2 v2.34.0
--- a/bundle/manifests/datadoghq.com_datadoggenericresources.yaml
+++ b/bundle/manifests/datadoghq.com_datadoggenericresources.yaml
@@ -58,6 +58,7 @@ spec:
               type:
                 description: Type is the type of the API object
                 enum:
+                - downtime
                 - monitor
                 - notebook
                 - synthetics_api_test
--- a/check-operator.Dockerfile
+++ b/check-operator.Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.25.3 AS builder
+FROM golang:1.25.5 AS builder
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
--- a/config/crd/bases/v1/datadoghq.com_datadoggenericresources.yaml
+++ b/config/crd/bases/v1/datadoghq.com_datadoggenericresources.yaml
@@ -57,6 +57,7 @@ spec:
                 type:
                   description: Type is the type of the API object
                   enum:
+                    - downtime
                     - monitor
                     - notebook
                     - synthetics_api_test
--- a/config/crd/bases/v1/datadoghq.com_datadoggenericresources_v1alpha1.json
+++ b/config/crd/bases/v1/datadoghq.com_datadoggenericresources_v1alpha1.json
@@ -24,6 +24,7 @@
         "type": {
           "description": "Type is the type of the API object",
           "enum": [
+            "downtime",
             "monitor",
             "notebook",
             "synthetics_api_test",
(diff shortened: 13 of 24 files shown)


pkg/controller/utils/metadata/shared_metadata.go
```diff
@@ -75,16 +67,41 @@ func NewSharedMetadata(logger logr.Logger, k8sClient client.Reader, kubernetesVe
...
+	}
+
+	apiKey, requestURL, err := sm.getApiKeyAndURL()
+	if err != nil {
+		sm.logger.Error(err, "Could not get credentials")
+		return nil, err
+	}
+	payloadHeader := sm.GetHeaders(*apiKey)
+
+	sm.logger.Info("Operator metadata payload", "payload", string(payload))
```
swang392: Should this be at the debug level to avoid bloating info-level logs? This is called by all forwarders (helm and operator) but the log message only mentions operator, which could be misleading when debugging. 

Nit: Previously for helm, in case scrubbing failed I also only logged the payload if `notScrubbed=false`. Eventually I think the payload log should be removed entirely, but we lose that safeguard for sensitive yaml configurations here.

pkg/controller/utils/metadata/operator_metadata.go
```diff
@@ -79,21 +74,11 @@ func NewOperatorMetadataForwarder(logger logr.Logger, k8sClient client.Reader, k
...
-
 	if omf.hostName == "" {
-		omf.logger.Error(ErrEmptyHostName, "Could not set host name; not starting operator metadata forwarder")
+		omf.logger.Error(ErrEmptyHostName, "Could not set host name; not starting helm metadata forwarder")
 		return
 	}
 
-	omf.payloadHeader = omf.getHeaders()
-
-	omf.updateResourceCounts()
```
swang392, line 95:
```suggestion
	omf.updateResourceCounts()
```
add this back, so any payloads sent at the start of the metadata forwarder include the latest resource counts

pkg/controller/utils/metadata/shared_metadata.go
levan-m: Removed logging, this isn't necessary in even in debug mode.

pkg/controller/utils/metadata/operator_metadata.go
levan-m: Added back.

pkg/controller/utils/metadata/shared_metadata.go
```diff
@@ -75,16 +67,39 @@ func NewSharedMetadata(logger logr.Logger, k8sClient client.Reader, kubernetesVe
...
 		hostName:          os.Getenv(constants.DDHostName),
 		httpClient: &http.Client{
 			Timeout: 10 * time.Second,
 		},
 		credsManager: credsManager,
-		decryptor:    secrets.NewSecretBackend(),
 	}
 }
 
+func (sm *SharedMetadata) createRequest(payload []byte) (*http.Request, error) {
```
swang392, line 78: This method is used by all metadata forwarders and though its functionally doing the same thing for all, there's no way to differentiate between the forwarders. Adding some string type to this method and adding that to all the error and info logs would make it easier to figure out which specific forwarder fails at a given step
```suggestion
func (sm *SharedMetadata) createRequest(payload []byte) (*http.Request, error, forwarderType) {
```

levan-m: If it's just logging, we can create the forwarder with extended logger name
```go
	forwarderLogger := logger.WithName("operator")
	return &OperatorMetadataForwarder{
		SharedMetadata:   NewSharedMetadata(forwarderLogger, k8sClient, kubernetesVersion, operatorVersion, credsManager),
		OperatorMetadata: OperatorMetadata{},
	}
```
to get this output
```log
{"level":"DEBUG","ts":"2025-12-03T20:18:18.382Z","logger":"metadata.helm","msg":"Successfully sent all Helm release metadata","count":1}
{"level":"DEBUG","ts":"2025-12-03T20:18:18.383Z","logger":"metadata.operator","msg":"Read operator metadata response","status code":202,"body":"Accepted"}
```

pkg/controller/utils/metadata/shared_metadata.go
```diff
@@ -101,84 +116,59 @@ func (sm *SharedMetadata) GetOrCreateClusterUID() (string, error) {
...
-// it falls back to setting up from DatadogAgent to ensure we have a valid cluster name.
-func (sm *SharedMetadata) setCredentials() error {
-	err := sm.setupFromOperator()
-	if err == nil && sm.clusterName != "" {
-		return nil
+// getApiKeyAndURL retrieves the API key and request URL from the operator or DDA
+// and sets the cluster name from the operator or DDA in the SharedMetadata struct
+func (sm *SharedMetadata) getApiKeyAndURL() (*string, *string, error) {
+	// Get credentials (operator → DDA fallback handled internally), dda can be nil
+	dda, _ := sm.getDatadogAgent()
```
swang392: If there's an issue with DDA retrieval here, should we log the error and note that creds will only be set via the operator? Since the error is discarded, we don't have visibility into any errors related to k8s api call or failure to list DDAs.

pkg/controller/utils/metadata/shared_metadata.go
```diff
@@ -101,84 +116,59 @@ func (sm *SharedMetadata) GetOrCreateClusterUID() (string, error) {
...
+		Path:   defaultURLPath,
+	}
+	if creds.Site != nil {
+		mdfURL.Host = defaultURLHostPrefix + *creds.Site
 	}
 
-	return sm.setupFromDDA(dda)
+	if creds.URL != nil {
+		tempURL, err := url.Parse(*creds.URL)
+		if err == nil {
```
swang392, line 156: Right now this silently falls back to default, so if a user is looking for metadata at their custom URL and parsing failed, there is no logging to explain why the metadata was sent to the default url.

pkg/config/creds.go
```diff
@@ -179,3 +237,109 @@ func (cm *CredentialManager) StartCredentialRefreshRoutine(interval time.Duratio
...
+
+	if dda.Spec.Global == nil || dda.Spec.Global.Credentials == nil {
+		return creds, fmt.Errorf("credentials not configured in the DatadogAgent")
+	}
+
+	defaultSecretName := secrets.GetDefaultCredentialsSecretName(dda)
+
+	apiKey := ""
+	err := error(nil)
+	if dda.Spec.Global != nil && dda.Spec.Global.Credentials != nil && dda.Spec.Global.Credentials.APIKey != nil && *dda.Spec.Global.Credentials.APIKey != "" {
```
swang392, line 262: nit
```suggestion
	if dda.Spec.Global.Credentials.APIKey != nil && *dda.Spec.Global.Credentials.APIKey != "" {
```

pkg/controller/utils/metadata/shared_metadata.go
levan-m: DDA is just fallback for resolving api/app key and URL so it's optional and for situations where DDA isn't present they'd get the recurring errors unnecessarily. If Operator api/app key can't be resolved and DDA is nil, generic [error](https://github.com/DataDog/datadog-operator/pull/2351/files#diff-a9a24878b2fd6dbb1cc87e7dcedf9b6bfc8f2497057cd4f73e539a0eda882751R246) will be returned from creds.go.

levan-m: ..ideally we can improve `CredentialManager` to handle DDA internally.

pkg/controller/utils/metadata/shared_metadata.go
levan-m: This code initializes `mdfURL` with defaults and then overrides it with one returned with `creds` which would be either from Operator env or DDA [here](https://github.com/DataDog/datadog-operator/pull/2351/files#diff-a9a24878b2fd6dbb1cc87e7dcedf9b6bfc8f2497057cd4f73e539a0eda882751R148-R158) for Operator and [here](https://github.com/DataDog/datadog-operator/pull/2351/files#diff-a9a24878b2fd6dbb1cc87e7dcedf9b6bfc8f2497057cd4f73e539a0eda882751R148-R158) for DDA. I _think_ that aligns with earlier behavior below. let me know if I'm missing something.

1. Initialize `requestURL: getURL()` to defaults, which is default overriden by Operator env:
```go
func getURL() string {
	mdfURL := url.URL{
		Scheme: defaultURLScheme,
		Host:   defaultURLHost,
		Path:   defaultURLPath,
	}

	// check site env var
	// example: datadoghq.com
	if siteFromEnvVar := os.Getenv("DD_SITE"); siteFromEnvVar != "" {
		mdfURL.Host = defaultURLHostPrefix + siteFromEnvVar
	}
	// check url env var
	// example: https://app.datadoghq.com
	if urlFromEnvVar := os.Getenv("DD_URL"); urlFromEnvVar != "" {
		tempURL, err := url.Parse(urlFromEnvVar)
		if err == nil {
			mdfURL.Host = tempURL.Host
			mdfURL.Scheme = tempURL.Scheme
		}
	}

	return mdfURL.String()
```
2. Try Operator API key, no url handling here
```go
func (sm *SharedMetadata) setupFromOperator() error {
// no URL handling, just API key
```

3.  Fallback to DDA override site
```go
func (sm *SharedMetadata) setupFromDDA(dda *v2alpha1.DatadogAgent) error {
...
	if sm.apiKey == "" {
		apiKey, err := sm.getCredentialsFromDDA(dda)
		if err != nil {
			return err
		}
		sm.apiKey = apiKey

		// if API key is set from DDA, also update request URL if needed
		if dda.Spec.Global != nil && dda.Spec.Global.Site != nil {
			mdfURL := url.URL{
				Scheme: defaultURLScheme,
				Host:   defaultURLHostPrefix + *dda.Spec.Global.Site,
				Path:   defaultURLPath,
			}
			sm.requestURL = mdfURL.String()
		}
	}
```

swang392: Ah yeah, that makes sense. Nvm, I think a log isn't needed because the request URL will already be logged when the metadata payload is sent. And the fallback behavior is the same as before

glasser:
Out of curiosity — I just installed Datadog Operator for the first time today, to use it for DatadogMetrics only (and not set up a DatadogAgent resource as we already have a way of installing our agents). While everything works fine, we are seeing this error log on startup

```
{
    "ts": "2025-12-10T20:12:32.554Z",
    "level": "ERROR",
    "stacktrace": "github.com/DataDog/datadog-operator/pkg/controller/utils/metadata.(*MetadataForwarder).Start\n\t/workspace/pkg/controller/utils/metadata/metadata_forwarder.go:122\nmain.setupAndStartMetadataForwarder\n\t/workspace/cmd/main.go:511\nmain.run\n\t/workspace/cmd/main.go:381\nmain.main\n\t/workspace/cmd/main.go:202\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:285",
    "error": "DatadogAgent not found",
    "msg": "Could not set credentials; not starting metadata forwarder",
    "logger": "metadata"
}
```

I went to look at the relevant code and I saw that it had been rewritten literally today!  I am curious — does this PR help with my issue (surprising error when DatadogAgent is disabled)?

levan-m:
Hi @glasser, thanks for submitting your question! You don't need to worry about this error log. I assume you are using 1.21.0-rc.2, this PR which is released in rc.3, changes how credentials are handled and errors logged. In any case, errors logged by metadata forwarder shouldn't affect how Operator manages cluster resources.

I'm curious why you need Datadog Operator if you don't use it for managing Agents and only need it for DatadogMetrics CRD which can be installed via `datadog-crds` [chart](https://github.com/DataDog/helm-charts/tree/main/charts/datadog-crds)?

PR #2354, opened 2025-11-25
Enable linters: `ineffassign`, `perfsprint` and `prealloc`

tbavelier:
### What does this PR do?

* Enables some linters that were disabled

### Motivation

* Best practices: ~~while not relevant here performance wise, `perfsprint` on a hot path can lead to drastic improvements: https://github.com/DataDog/datadog-agent/pull/43407~~
* Catch potential bugs

### Additional Notes

* ~~`perfsprint` makes string building harder to "understand"/follow compared to `fmt.Sprintf`~~
* Next step is to enable `staticcheck` and `errcheck`

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist


internal/controller/datadogagent/global/dependencies.go
```diff
@@ -296,7 +297,7 @@ func clusterAgentDependencies(ddaMeta metav1.Object, ddaSpec *v2alpha1.DatadogAg
 		}
 	}
 
-	return nil
+	return errors.Join(errs...)
```
AlexanderYastrebov, line 300: Looks like a bugfix

AlexanderYastrebov (approved):
perfsprint changes feel dubious - there likely will be no performance benefits but linter might be annoying to deal with.

internal/controller/datadogagent/global/dependencies.go
tbavelier: yes, we indeed were not returning these errors previously, another benefit of these linters, thanks!

docs/configuration_public.md
```diff
@@ -35,7 +35,7 @@ spec:
         keyName: app-key
 {{< /highlight >}}
 
-{{% collapse-content title="Parameters" level="h4" expanded=true id="global-options-list" %}}
+{{%% collapse-content title="Parameters" level="h4" expanded=true id="global-options-list" %%}}
```
tbavelier: Not sure about this one, @DataDog/documentation could you confirm whether or not this change makes sense and same below ?

internal/controller/datadogagent/global/dependencies.go
tbavelier, line 300: Behaviour change: we would not previously return errors from any of these steps, would always be nil. Now if such an error is added, we'll return it (as we should considering not being able to add service/RBACs is critical)

pkg/plugin/common/const.go
```diff
@@ -11,7 +11,7 @@ const (
 	// ADPrefix prefix used for AD annotations
 	ADPrefix = "ad.datadoghq.com/"
 	// ADPrefixRegex used for matching AD annotations
-	ADPrefixRegex = "ad\\.datadoghq\\.com/"
+	ADPrefixRegex = "^ad\\.datadoghq\\.com/"
```
tbavelier, line 14: Flagged by graphql bot security review

docs/configuration_public.md
buraizu:
```suggestion
{{% collapse-content title="Parameters" level="h4" expanded=true id="global-options-list" %}}
```

docs/configuration_public.md
```diff
@@ -609,7 +609,7 @@ spec:
 {{< /highlight >}}
 In the table, `spec.override.nodeAgent.image.name` and `spec.override.nodeAgent.containers.system-probe.resources.limits` appear as `[component].image.name` and `[component].containers.[container].resources.limits`, respectively.
 
-{{% collapse-content title="Parameters" level="h4" expanded=true id="override-options-list" %}}
+{{%% collapse-content title="Parameters" level="h4" expanded=true id="override-options-list" %%}}
```
buraizu:
```suggestion
{{% collapse-content title="Parameters" level="h4" expanded=true id="override-options-list" %}}
```

docs/configuration_public.md
buraizu: Thanks for flagging this @tbavelier. This caused errors in my local build, so I've suggested to revert this change.

buraizu (changes requested):

tbavelier:
We decided not to go forward with `perfsprint` for now/in the short term, so closing in favour of https://github.com/DataDog/datadog-operator/pull/2359 for commit history to cleaner/easier to review

PR #2341, opened 2025-11-21
feat: add downtime resource to operator

neelsvanwyk-wk:
## What does this PR do?

* Adds support for `downtime` as a `DatadogGenericResource`

## Motivation

* Supporting scheduled downtime management through the operator, enabling users to create, update, and delete downtimes using Kubernetes manifests:

```yaml
apiVersion: datadoghq.com/v1alpha1
kind: DatadogGenericResource
metadata:
  name: ddgr-downtime-sample
spec:
  type: downtime
  jsonSpec: |-
    {
      "data": {
        "type": "downtime",
        "attributes": {
          "message": "Scheduled maintenance window for example-service",
          "monitor_identifier": {
            "monitor_tags": [
              "service:example-service"
            ]
          },
          "scope": "env:production",
          "display_timezone": "UTC",
          "notify_end_types": [
            "expired",
            "canceled"
          ],
          "notify_end_states": [
            "alert",
            "warn"
          ],
          "schedule": {
            "start": null
          }
        }
      }
    }
```

## Additional Notes

* This implementation uses the Datadog v2 API since that supports downtimes, the previous implementations used the v1 API, I'm unsure if that facilitates additional changes? 
* This PR was based off the addition of monitor support ([#1635](https://github.com/DataDog/datadog-operator/pull/1635))
* It seems like it was necessary in `UpdateDatadogClient` to properly update `NotebooksClient` and `DowntimesClient` fields during credential rotation. Was it intentional that `NotebooksClient` was omitted previously?

## Minimum Agent Versions

Are there minimum versions of the Datadog Agent and/or Cluster Agent required?

* Agent: N/A (operator-only feature)
* Cluster Agent: N/A (operator-only feature)

## Describe your test plan

Same as [#1534](https://github.com/DataDog/datadog-operator/pull/1534) and [#1635](https://github.com/DataDog/datadog-operator/pull/1635): test the whole lifecycle of the downtime object:
* Create downtime resource
* Update downtime attributes (message, scope, schedule)
* (Optionally: update from Datadog UI and verify forcesync)
* Delete downtime resource

Still to be tested in local kind cluster with successful CRUD operations, will take the PR out of draft state afterwards

## Checklist

- [ ] PR has at least one valid label: `bug`, `enhancement`, `refactoring`, `documentation`, `tooling`, and/or `dependencies`
- [ ] PR has a milestone or the `qa/skip-qa` label

PR #2363, opened 2025-12-01
(feat) Add downtime GenericResource

tbavelier:
The below is taken nearly verbatim from #2341 , on a local branch to have CI run properly

---

## What does this PR do?

* Adds support for `downtime` as a `DatadogGenericResource`

## Motivation

* Supporting scheduled downtime management through the operator, enabling users to create, update, and delete downtimes using Kubernetes manifests:

```yaml
kind: DatadogGenericResource
metadata:
  name: ddgr-downtime-sample
spec:
  type: downtime
  jsonSpec: |-
    {
      "data": {
        "type": "downtime",
        "attributes": {
          "message": "Scheduled maintenance window for example-service",
          "monitor_identifier": {
            "monitor_tags": [
              "service:example-service"
            ]
          },
          "scope": "foo:bar",
          "display_timezone": "UTC",
          "notify_end_types": [
            "expired",
            "canceled"
          ],
          "notify_end_states": [
            "alert",
            "warn"
          ],
          "schedule": {
            "start": null
          }
        }
      }
    }
```

## Minimum Agent Versions

* Agent: N/A (operator-only feature)
* Cluster Agent: N/A (operator-only feature)

## Describe your test plan

Same as [#1534](https://github.com/DataDog/datadog-operator/pull/1534) and [#1635](https://github.com/DataDog/datadog-operator/pull/1635): test the whole lifecycle of the downtime object:
1. Create downtime resource from `k apply -f examples/datadoggenericresource/downtime-sample.yaml`
    1. It will error out (from the API) when using the example as is, due to the date being in the past. You can do a `k edit ddgr ddgr-downtime-sample` to fix it by changing the start to something in the future
3. Update downtime attributes, e.g. scope or w/e and verify it's reflected in the UI
4. Modify the downtime from the UI, e.g. scope/recurrence, anything
5. Verify forceSync to put it back in the original state (`k edit ddgr ddgr-downtime-sample --subresource=status` and remove the `status.lastForceSyncTime` field completely): after the edit, you will need to wait up to a minute (default requeue time)
7. Delete downtime resource

## Checklist


tbavelier:
Hello @neelsvanwyk-wk ,

Thank you again for the contribution, especially the fix for credentials refresh that was missed! After reviewing and playing around with the build, I made a few modifications:
* 7b41fb089ec638c77808fde00ad01d4e213a05fb: this will come with 1.22 (next release), not 1.14
* 4a5d3db8bdab1ea851713c801d5db5f22d75e77e: I modified the example to use a recurring downtime and few fields to make it it clear it's an example with a dummy scope instead of possibly targetting a real environment. The idea about the recurring downtime compared to a one-off is that it makes more sense to me to manage those with a Terraform-like behaviour a bit more, but both use cases are valid and work anyway
* cba1479ba9a6394981dea308739b389c72f26788: I was having a hard time with the update method about some missing / invalid field, so reworked slightly to use the public constructor from `datadogV2` which I think is a net benefit in any case
* 2d383762d92c5f75b638c9390bb40fc80075ba77: not related directly to this PR, but while inspecting the object after a flow from error update to successful sync, I noticed we weren't clearing the previous error condition

I also modified the QA step in the PR description.

Could you please take a look and let us know if that works for you ?

Regards,

neelsvanwyk-wk:
This works excellently, thanks for the additional improvements @tbavelier!

internal/controller/datadoggenericresource/controller.go
```diff
@@ -207,6 +214,8 @@ func (r *Reconciler) create(logger logr.Logger, instance *v1alpha1.DatadogGeneri
 
 	// Set condition and status
 	condition.UpdateStatusConditions(&status.Conditions, now, condition.DatadogConditionTypeCreated, metav1.ConditionTrue, "CreatingGenericResource", "DatadogGenericResource Created")
+	// Clear error condition on successful creation
+	condition.UpdateStatusConditions(&status.Conditions, now, condition.DatadogConditionTypeError, metav1.ConditionFalse, "", "")
```
levan-m: I noticed errors in operator log, it seems reason can't be empty string here, see [this](https://pkg.go.dev/k8s.io/apimachinery@v0.33.3/pkg/apis/meta/v1#Condition). 
Not sure why was this added but I guess either we leave as is (an rely on lastTransition) or we have to remove the condition.
```
 ...is invalid: status.conditions[1].reason: Invalid value: \"\": conditions[1].reason in body should be at least 1 chars long","stacktrace":"sigs.k8s.io/controller..
```

tbavelier: Good catch! It was added as we'd leave the error condition to true despite a successful sync so it could be a bit confusing even if lasttransitiontime showed it was before a successful sync. Fixed by 7e3d4996f111db744e485a6f33686f4d0d11dd40

PR #2373, opened 2025-12-02
[e2e] Fix flaky log collection and ksm tests

swang392:
### What does this PR do?

- 1199a7de31b4ec472584013e953350831210372d and 3ca547b1a8a73ef9ea5d91d1d7713b1943b9d806: fix flakiness in the logs collection check, caused by trying to parse logs agent status before it is ready. Added gating to prevent errors
- 48ef15e8f31bca43ad29d9ff8f57c535cdf358ad naming issue in ksm check. Previously searching for labels that did not correspond to the dda name 
- 352275fe1b7ec4eff2e8c34cf9756132c100252c: same as for logs, let's early exit instead of panicking when everything is not ready
- f2687817b95652d4862f996e2a025f4c99835247: delete old manifest not used and causing confusion
- 64f3e13bd8168bebdf54c6c480294c11f6a157cc: fix a rare race condition where token does not match between DCA and CCR leading to the latter unable to connect to DCA, so checks are not distributed

previously e2e tests took ~40 min and sometimes restarted up to 2x due to flakiness. 
Now this [example pipeline](https://gitlab.ddbuild.io/DataDog/datadog-operator/-/jobs/1266765452) has no retries and tests take ~30 min. 

### Motivation

hacking !!!!

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

tested via running e2e tests in CI

### Checklist

### DIFF
--- a/internal/controller/datadogagent/feature/cws/feature.go
+++ b/internal/controller/datadogagent/feature/cws/feature.go
@@ -182,6 +182,9 @@ func (f *cwsFeature) ManageSingleContainerNodeAgent(managers feature.PodTemplate
 // ManageNodeAgent allows a feature to configure the Node Agent's corev1.PodTemplateSpec
 // It should do nothing if the feature doesn't need to configure it.
 func (f *cwsFeature) ManageNodeAgent(managers feature.PodTemplateManagers, provider string) error {
+	// enable HostPID for system-probe
+	managers.PodTemplateSpec().Spec.HostPID = true
+
 	// annotations
 	managers.Annotation().AddAnnotation(common.SystemProbeAppArmorAnnotationKey, common.SystemProbeAppArmorAnnotationValue)
 
--- a/internal/controller/datadogagent/feature/npm/feature.go
+++ b/internal/controller/datadogagent/feature/npm/feature.go
@@ -88,6 +88,9 @@ func (f *npmFeature) ManageSingleContainerNodeAgent(managers feature.PodTemplate
 // ManageNodeAgent allows a feature to configure the Node Agent's corev1.PodTemplateSpec
 // It should do nothing if the feature doesn't need to configure it.
 func (f *npmFeature) ManageNodeAgent(managers feature.PodTemplateManagers, provider string) error {
+	// enable HostPID for system-probe
+	managers.PodTemplateSpec().Spec.HostPID = true
+
 	// annotations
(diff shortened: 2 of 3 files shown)


test/e2e/tests/utils/utils.go
```diff
@@ -47,32 +47,45 @@ func VerifyCheck(c *assert.CollectT, collectorOutput string, checkName string) {
...
 				assert.Zero(c, totalErrors)
 
-				totalMetricSamples, exists := instance.(map[string]interface{})["TotalMetricSamples"].(float64)
+				totalMetricSamples, exists := instanceMap["TotalMetricSamples"].(float64)
 				assert.True(c, exists)
 				assert.Greater(c, totalMetricSamples, float64(0))
 			}
 		} else {
-			assert.True(c, found, "Check %s not found or not yet running.", checkName)
+			assert.Failf(c, "Check not found", "Check %s not found or not yet running", checkName)
```
tbavelier, line 88: In the `else` branch here, `found` is false, so we'd guarantee the assertion fails here. Let's explicitly fail for clarity

test/e2e/manifests/datadog-agent-ccr-enabled.yaml
tbavelier, line 1: This is an original manifest that is not used, everything is in `new_manifests` now (for a while). We should probably in a follow-up PR move everything back to the root `manifests`.

test/e2e/manifests/new_manifests/datadog-agent-ccr-enabled.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
     agent.datadoghq.com/e2e-test: datadog-agent-ccr-enabled
 spec:
   global:
+    clusterAgentToken: thisisa32characterstoken00000000
```
tbavelier, line 10: There's rarely a race condition I'm investigating where the DCA token is set to an older value (possibly from previous reconciliation or something) while the CCR has the new / correct value (from the secret). This causes the CCR to be unable to connect to DCA and the test fails eventually. By forcing a token, we ensure we don't end up in this situation.

tbavelier (approved):

test/e2e/manifests/datadog-agent-ccr-enabled.yaml
swang392: yep, agreed. i started on that in a follow-up branch with more refactoring (still testing that one)

PR #2408, opened 2025-12-16
Do not mount tracefs for CWS

lebauce:
### What does this PR do?

Do not mount tracefs on /sys/kernel/tracing

### Motivation

On RHEL 7, /sys/kernel/tracing is not available. Besides, /sys/kernel/debug is
already mounted so mounting both is of little use.

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist


lebauce:
@tbavelier When do you think we can have this can available to customers ? I'm afraid 1.22 won't be released before long right ?

PR #2411, opened 2025-12-16
DCA status cleanup

levan-m:
### What does this PR do?

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist

### DIFF
--- a/internal/controller/datadogagent/controller_v2_test.go
+++ b/internal/controller/datadogagent/controller_v2_test.go
@@ -421,8 +421,8 @@ func TestReconcileDatadogAgentV2_Reconcile(t *testing.T) {
 				if err := c.Get(context.TODO(), types.NamespacedName{Namespace: resourcesNamespace, Name: resourcesName}, dda); client.IgnoreNotFound(err) != nil {
 					return err
 				}
-				// assert.Equal(t, "token", dda.Status.ClusterAgent.GeneratedToken)
 				assert.NotNil(t, dda.Status.ClusterAgent, "DCA status should be set")
+				assert.Equal(t, "token", dda.Status.ClusterAgent.GeneratedToken)
 				dcaCondition := condition.GetCondition(&dda.Status, common.ClusterAgentReconcileConditionType)
 				assert.True(t, dcaCondition.Status == metav1.ConditionTrue && dcaCondition.Reason == "reconcile_succeed", "DCA status condition should be set")
 				return nil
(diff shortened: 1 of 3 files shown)


pkg/condition/condition.go
```diff
@@ -104,6 +104,21 @@ func GetMetav1ConditionStatus(status bool) metav1.ConditionStatus {
...
 
+func GetCondition(status *v2alpha1.DatadogAgentStatus, t string) *metav1.Condition {
+	idConditionComplete := getIndexForConditionType(status, t)
+	if idConditionComplete >= 0 {
+		return &status.Conditions[idConditionComplete]
+	}
+	return nil
+}
+
+func GetConditionDDAI(status *v1alpha1.DatadogAgentInternalStatus, t string) *metav1.Condition {
```
tbavelier: This function is not used: is a test supposed to use it or it's added for parity + see below comment ?

pkg/condition/condition.go
```diff
@@ -104,6 +104,21 @@ func GetMetav1ConditionStatus(status bool) metav1.ConditionStatus {
 	return metav1.ConditionFalse
 }
 
+func GetCondition(status *v2alpha1.DatadogAgentStatus, t string) *metav1.Condition {
```
tbavelier, line 107: GetCondition is solely used in `controller_v2_test.go`, it's a bit unfortunate to expose it there. Could we maybe move it to a helper not public util function ? (Tho, it will require to change `getIndexForConditionType` to a public function) ? Or we can keep it as is if we see value in using it elsewhere in the future

internal/controller/datadogagent/controller_v2_test.go
```diff
@@ -395,6 +396,72 @@ func TestReconcileDatadogAgentV2_Reconcile(t *testing.T) {
...
+				return dda
+			},
+			want:    reconcile.Result{RequeueAfter: defaultRequeueDuration},
+			wantErr: false,
+			wantFunc: func(c client.Client) error {
+				dda := &v2alpha1.DatadogAgent{}
+				if err := c.Get(context.TODO(), types.NamespacedName{Namespace: resourcesNamespace, Name: resourcesName}, dda); client.IgnoreNotFound(err) != nil {
+					return err
+				}
+				// assert.Equal(t, "token", dda.Status.ClusterAgent.GeneratedToken)
```
tbavelier:
```suggestion
```
To be un-commented and moved after checking DCA status is not nil (line after) ?

internal/controller/datadogagent/component_clusteragent.go
```diff
@@ -126,6 +125,7 @@ func (r *Reconciler) cleanupV2ClusterAgent(logger logr.Logger, dda *datadoghqv2a
 	clusterAgentDeployment := &appsv1.Deployment{}
 	if err := r.client.Get(context.TODO(), nsName, clusterAgentDeployment); err != nil {
 		if errors.IsNotFound(err) {
+			deleteStatusV2WithClusterAgent(newStatus)
```
tbavelier, line 128: Should this also be added in component_clustercheckrunner.go for parity ?

pkg/condition/condition.go
levan-m: agree with reluctance with exposing functions just for the tests but this is a simple read-only getter which imo is a good tradeoff vs exposing getIndex or reimplementing it in the test..

pkg/condition/condition.go
levan-m: Will delete this. I thought I'd use it in the DDAI flow but we don't have a test where we wouldn't need this. I missed updating DDAI component_clusteragent.go and made the change now.

internal/controller/datadogagent/component_clusteragent.go
levan-m: CLC variant is a bit different implementation but should be similar behavior as this one. It exits only if error is _not_ `notFound`, otherwise proceeds with cleanup.
```go
	if err := r.client.Get(context.TODO(), nsName, ClusterChecksRunnerDeployment); err != nil {
		if !errors.IsNotFound(err) {
		}
```

levan-m: ..added test for CLC.

pkg/condition/condition.go
tbavelier: Agreed!

PR #2412, opened 2025-12-16
Add new options field with OOMBumpRatio as the first option 

nickbirnberg:
### What does this PR do?

  Adds a new optional options section to the DatadogPodAutoscalerSpec in the v1alpha2 API. This section initially contains an outOfMemory configuration with a bumpUpRatio field that defines the ratio to multiply memory by when OOM events are detected.

  Example usage:
```
  spec:
    options:
      outOfMemory:
        bumpUpRatio: "1.2"  # Multiply memory by 1.2 (20% increase) on OOM
```
### Motivation

  Support memory scale-up configuration when out-of-memory events are detected, allowing users to specify how much to increase memory allocation in response to OOM kills.

### Additional Notes

  - The bumpUpRatio field uses resource.Quantity type to avoid floating point precision issues in CRDs
  - A value of "1.2" means increase memory by 20% (multiply current by 1.2)
  - The new types are defined in the common package (api/datadoghq/common/datadogpodautoscaler_types.go) for potential reuse

### Minimum Agent Versions

  N/A - This is an API-only change for the operator CRD.

  - Agent: N/A
  - Cluster Agent: N/A

### Describe your test plan

  1. Apply a DatadogPodAutoscaler with the new options field:
  ```
  apiVersion: datadoghq.com/v1alpha2
  kind: DatadogPodAutoscaler
  metadata:
    name: test-dpa
  spec:
    targetRef:
      apiVersion: apps/v1
      kind: Deployment
      name: my-app
    owner: Local
    options:
      outOfMemory:
        bumpUpRatio: "1.2"
   ```
  2. Verify the resource is accepted by the API server
  3. Verify kubectl get dpa test-dpa -o yaml shows the options field correctly

### Checklist

  - PR has at least one valid label: bug, enhancement, refactoring, documentation, tooling, and/or dependencies
  - PR has a milestone or the qa/skip-qa label

### DIFF
--- a/api/datadoghq/v1alpha2/datadogpodautoscaler_types.go
+++ b/api/datadoghq/v1alpha2/datadogpodautoscaler_types.go
@@ -12,7 +12,6 @@ import (
 	"github.com/DataDog/datadog-operator/api/datadoghq/common"
 )
 
-
 // spec:
 //   targetRef:
 //     apiVersion: apps/v1


api/datadoghq/v1alpha2/datadogpodautoscaler_types.go
```diff
@@ -89,6 +89,8 @@ import (
 //         requests:
 //           minAllowed:
 //           maxAllowed:
+//   recommendationOptions:
```
vboulineau: I'm thinking if we should include `recommendations` because it's not _really_ tied to recommendations, technically it could applied in-cluster on top of recommendations.

What do you think about
```
options | settings:
  outOfMemory:
    increaseFactor: 20% (same syntax as `Deployment.MaxUnavailable`?)
```

nickbirnberg: that's good for me, I updated with the hierarchy. However, I kept the `bumpUpRatio` name and use of a multiplier for familiarity (with VPA) and to reduce confusion. but can go to percentages.

vboulineau: Makes sense to align with VPA then.

tbavelier (approved):

PR #2301, opened 2025-11-07
[CASCL-647] Add a subcommand to kubectl plugin to install Karpenter

L3n41c:
### What does this PR do?

Add a subcommand to the `kubectl` plugin to install Karpenter on a cluster:
```
kubectl datadog autoscaling cluster install
```

### Motivation

Installing Karpenter is a pre-requisite for the cluster auto-scaling product.

### Additional Notes

[asciicast](https://asciinema.org/a/z6XU8uSY4WvnfFP0sh0vjalU5)

### Minimum Agent Versions

Are there minimum versions of the Datadog Agent and/or Cluster Agent required?
No

### Describe your test plan

Create an EKS cluster with several node pools and deploy on it a Deployment that cannot fit on the cluster nodes.
We can use the following script that creates a deployment with 5 replicas and an anti-affinity rule that ensures that they cannot fit on the 2 not-tainted nodes.

* `common.sh`
```bash
# AWS_ACCOUNT=agent-sandbox
# export AWS_REGION=us-east-1
AWS_ACCOUNT=container-integrations
export AWS_REGION=eu-west-3

export CLUSTER_NAME=${USER}-karpenter-test

export AWS_PROFILE=exec-sso-${AWS_ACCOUNT}-account-admin

export VPC=$(aws ec2 describe-vpcs \
                 --filters "Name=tag:Name,Values=$AWS_ACCOUNT" \
                 --query 'Vpcs[0].VpcId' \
                 --output text)
```

* `create_cluster.sh`
```bash
#!/usr/bin/env bash
set -euo pipefail

cd "$(dirname "${BASH_SOURCE[0]}")"
source common.sh

SG=$(aws ec2 create-security-group \
         --description "Security group for ${CLUSTER_NAME} EKS cluster" \
         --group-name "${CLUSTER_NAME}" \
         --vpc-id "${VPC}" \
         --output text)

PL_APPGATE=$(aws ec2 describe-managed-prefix-lists \
                 --filters 'Name=prefix-list-name,Values=vpn-services-commercial-appgate' \
                 --query 'PrefixLists[*].PrefixListId' \
                 --output text)

for tcp_port in 22 443; do
    aws ec2 authorize-security-group-ingress \
        --group-id "${SG}" \
        --ip-permissions "IpProtocol=tcp,FromPort=${tcp_port},ToPort=${tcp_port},PrefixListIds=[{PrefixListId=${PL_APPGATE}}]"
done

[...]

PR #2424, opened 2025-12-22
[CASCL-647] Add a subcommand to kubectl plugin to uninstall Karpenter

L3n41c:
### What does this PR do?

Add a subcommand to the kubectl plugin to uninstall Karpenter on a cluster:
```
kubectl datadog autoscaling cluster uninstall
```

### Motivation

#2301 added a subcommand to install Karpenter.
This PR introduce a subcommend to do the reverse operation.

### Minimum Agent Versions

No

### Describe your test plan

* Create an EKS cluster;
* Install Karpenter on it with `kubectl datadog autoscaling cluster install`;
* Then validate that it can be uninstalled with `kubectl datadog autoscaling cluster uninstall`.

### Checklist

### DIFF
--- a/cmd/kubectl-datadog/autoscaling/cluster/uninstall/uninstall.go
+++ b/cmd/kubectl-datadog/autoscaling/cluster/uninstall/uninstall.go
@@ -135,10 +135,7 @@ func (o *options) run(cmd *cobra.Command) error {
 		return fmt.Errorf("failed to build clients: %w", err)
 	}
 
-	nodePoolNames, nodes, err := displayResourceSummary(ctx, cmd, cli, clusterName)
-	if err != nil {
-		return err
-	}
+	nodePoolNames, nodes := displayResourceSummary(ctx, cmd, cli, clusterName)
 	if len(nodes) > 0 && !yes {
 		if promptConfirmation(cmd) != nil {
 			return nil // User cancelled
@@ -191,14 +188,15 @@ func (o *options) run(cmd *cobra.Command) error {
 	return nil
 }
 
-func displayResourceSummary(ctx context.Context, cmd *cobra.Command, cli *clients.Clients, clusterName string) (nodePools []string, nodes []string, err error) {
+func displayResourceSummary(ctx context.Context, cmd *cobra.Command, cli *clients.Clients, clusterName string) (nodePools []string, nodes []string) {
 	cmd.Println("\nThis will delete:")
 
-	if nodePools, err = listKarpenterNodePools(ctx, cli); err != nil {
-		return nil, nil, fmt.Errorf("failed to list NodePools: %w", err)
-	} else if len(nodePools) == 0 {
+	if n, err := listKarpenterNodePools(ctx, cli); err != nil {
+		cmd.Printf("  - NodePools: (unable to list: %v)\n", err)
+	} else if len(n) == 0 {
 		cmd.Println("  - NodePools: none found")
 	} else {
+		nodePools = n
 		cmd.Printf("  - %d NodePool(s):\n", len(nodePools))
 		for _, np := range nodePools {
 			cmd.Printf("      • %s\n", np)
@@ -207,7 +205,7 @@ func displayResourceSummary(ctx context.Context, cmd *cobra.Command, cli *client
 
 	nodeClasses, err := listKarpenterEC2NodeClasses(ctx, cli)
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to list EC2NodeClasses: %w", err)
+		cmd.Printf("  - EC2NodeClasses: (unable to list: %v)\n", err)
 	} else if len(nodeClasses) == 0 {
 		cmd.Println("  - EC2NodeClasses: none found")
 	} else {
@@ -217,11 +215,14 @@ func displayResourceSummary(ctx context.Context, cmd *cobra.Command, cli *client
 		}
 	}
 
-	if nodes, err = listKarpenterNodes(ctx, cli, nodeClasses); err != nil {
-		return nil, nil, fmt.Errorf("failed to list Karpenter nodes: %w", err)
-	} else if len(nodes) == 0 {
+	if err != nil {
+		cmd.Println("  - Karpenter nodes: (unable to list - depends on EC2NodeClasses)")
+	} else if n, err := listKarpenterNodes(ctx, cli, nodeClasses); err != nil {
+		cmd.Printf("  - Karpenter nodes: (unable to list: %v)\n", err)
+	} else if len(n) == 0 {
 		cmd.Println("  - Karpenter nodes: none found")
 	} else {
+		nodes = n
 		cmd.Printf("  - %d Karpenter-managed node(s):\n", len(nodes))
 		for _, node := range nodes {
 			cmd.Printf("      • %s\n", node)
@@ -231,7 +232,7 @@ func displayResourceSummary(ctx context.Context, cmd *cobra.Command, cli *client
 	cmd.Println("  - The Karpenter Helm release")
 
 	if stacks, err := listCloudFormationStacks(ctx, cli, clusterName); err != nil {
-		return nil, nil, fmt.Errorf("failed to list CloudFormation stacks: %w", err)
+		cmd.Printf("  - CloudFormation stacks: (unable to list: %v)\n", err)
 	} else if len(stacks) == 0 {
 		cmd.Println("  - CloudFormation stacks: none found")
 	} else {
@@ -248,7 +249,7 @@ func displayResourceSummary(ctx context.Context, cmd *cobra.Command, cli *client
 		cmd.Println(color.YellowString("⚠ WARNING: %d Karpenter node(s) will be drained and terminated.", len(nodes)))
 	}
 
-	return nodePools, nodes, nil
+	return nodePools, nodes
 }
 
 func promptConfirmation(cmd *cobra.Command) error {


L3n41c:
@codex review

L3n41c:
@codex review

L3n41c:
@codex review

L3n41c:
@codex review

cmd/kubectl-datadog/autoscaling/cluster/common/k8s/object.go
```diff
@@ -58,3 +58,59 @@ func update(ctx context.Context, cli client.Client, object client.Object) error
...
+
+	return nil
+}
+
+func extractListItems(list client.ObjectList) ([]client.Object, error) {
+	switch v := list.(type) {
+	case interface{ GetItems() []client.Object }:
+		return v.GetItems(), nil
+	default:
+		return nil, fmt.Errorf("unsupported list type: %T", list)
```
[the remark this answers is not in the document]
L3n41c: This actually works for `NodePoolList` and `EC2NodeClassList`.

L3n41c:
@codex review

L3n41c:
@codex review

cmd/kubectl-datadog/autoscaling/cluster/common/display/display.go
```diff
@@ -0,0 +1,44 @@
...
+// characters (such as emojis and East Asian characters).
+func PrintBox(w io.Writer, lines ...string) {
+	maxWidth := lo.Max(lo.Map(lines, func(line string, _ int) int {
+		return visualWidth(line)
+	}))
+
+	fmt.Fprintln(w, "╭─"+strings.Repeat("─", maxWidth)+"─╮")
+	for _, line := range lines {
+		padding := maxWidth - visualWidth(line)
+		fmt.Fprintf(w, "│ %s%s │\n", line, strings.Repeat(" ", padding))
```
AlexanderYastrebov, line 41: nit: could use `%-*s`
```go
func main() {
	line := "hello"
	padding := 20
	fmt.Printf("│ %-*s │\n", padding, line)
}
```

cmd/kubectl-datadog/autoscaling/cluster/common/aws/aws-auth.go
```diff
@@ -55,3 +55,50 @@ func EnsureAwsAuthRole(ctx context.Context, clientset kubernetes.Interface, role
...
+
+	found := false
+	updatedRoles := make([]RoleMapping, 0, len(roles))
+	for _, role := range roles {
+		if role.RoleArn == roleArn {
+			found = true
+			continue
+		}
+		updatedRoles = append(updatedRoles, role)
+	}
```
AlexanderYastrebov: nit: could use [slices.DeleteFunc](https://pkg.go.dev/slices#DeleteFunc):
```go
oldLen := len(roles)
roles = slices.DeleteFunc(roles, func(role RoleMapping) bool { return role.RoleArn == roleArn })
found := oldLen != len(roles)
```

cmd/kubectl-datadog/autoscaling/cluster/common/display/display_test.go
```diff
@@ -0,0 +1,167 @@
...
+func TestPrintBox(t *testing.T) {
+	tests := []struct {
+		name     string
+		lines    []string
+		expected string
+	}{
+		{
+			name:  "single line",
+			lines: []string{"Hello"},
+			expected: "╭───────╮\n" +
```
AlexanderYastrebov: nit: could line them up nicely via empty string prefix:
```go
			expected: "" +
				"╭──────────╮\n" +
				"│ Hello 🎉 │\n" +
				"│ World    │\n" +
				"╰──────────╯\n",
```

cmd/kubectl-datadog/autoscaling/cluster/common/k8s/object.go
```diff
@@ -58,3 +58,19 @@ func update(ctx context.Context, cli client.Client, object client.Object) error
...
 }
+
+func Delete(ctx context.Context, cli client.Client, object client.Object) error {
+	log.Printf("Deleting %s %s…", object.GetObjectKind().GroupVersionKind().Kind, object.GetName())
+
+	if err := cli.Delete(ctx, object); err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Printf("%s %s not found, skipping deletion.", object.GetObjectKind().GroupVersionKind().Kind, object.GetName())
+			return nil
+		}
```
AlexanderYastrebov, line 69: Is it a good idea to mask 404? E.g. update bubbles it up.

cmd/kubectl-datadog/autoscaling/cluster/uninstall/uninstall.go
```diff
@@ -0,0 +1,529 @@
...
+	if len(nodes) > 0 && !yes {
+		if promptConfirmation(cmd) != nil {
+			return nil // User cancelled
+		}
+	}
+
+	// Accumulate errors from cleanup steps - continue on failure to clean up as much as possible
+	var errs []error
+
+	if err = deleteKarpenterNodePools(ctx, cli); err != nil {
```
AlexanderYastrebov, line 148: Does it make sense to proceed in case of error? Maybe its better to bail out an let user resolve some steps manually.

cmd/kubectl-datadog/autoscaling/cluster/uninstall/uninstall.go
```diff
@@ -0,0 +1,529 @@
...
+
+	display.PrintBox(cmd.OutOrStdout(), "✅ Karpenter uninstalled from cluster "+clusterName+".")
+
+	return nil
+}
+
+func displayResourceSummary(ctx context.Context, cmd *cobra.Command, cli *clients.Clients, clusterName string) []string {
+	cmd.Println("\nThis will delete:")
+
+	if nodePools, err := listKarpenterNodePools(ctx, cli); err != nil {
```
AlexanderYastrebov: Isn't this a good sign that something is off? Maybe its better to bail out if we can not list stuff to delete.

cmd/kubectl-datadog/autoscaling/cluster/common/display/display.go
L3n41c: `Printf` width specifier cannot be used because it counts the number of runes whereas I need the display width here.

Ex. of program to show the issue:
```go
package main

import (
        "fmt"
)

func main() {
        for _, line := range []string{
                "I want all vertical",
                "bars to be aligned",
                "on both sides.",
                "The problem is that fmt.Printf",
                "counts width in runes.",
                "And some characters have",
                "a display width greater than 1.",
                "Example: 🚀🚀🚀🚀🚀",
                "It also doesn’t work",
                "when there are \033[1;31mcolors\033[0m",
        } {
                fmt.Printf("│ %-*s │\n", 40, line)
        }
}
```

image

AlexanderYastrebov: FWIW aligning emojis might not work either way as they are not fixed width glyphs:

[playground](https://go.dev/play/p/SUKobMtbT0f)

https://go.dev/play/p/SUKobMtbT0f

cmd/kubectl-datadog/autoscaling/cluster/common/helm/helm.go
```diff
@@ -125,3 +128,51 @@ func upgrade(ctx context.Context, ac *action.Configuration, releaseName, namespa
...
+	kubeConfig := ""
+	if configFlags.KubeConfig != nil {
+		kubeConfig = *configFlags.KubeConfig
+	}
+	kubeContext := ""
+	if configFlags.Context != nil {
+		kubeContext = *configFlags.Context
+	}
+
+	restClientGetter := kube.GetConfig(kubeConfig, kubeContext, namespace)
```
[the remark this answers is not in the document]
L3n41c: Fixed in https://github.com/DataDog/datadog-operator/pull/2424/changes/f8616f7585906d00ed9d7831547236b20927a49e#diff-c8b943f4b048b7d6e8f1e96bcba9e124e2e3a038c50b1e49595074b72d3c2b16

cmd/kubectl-datadog/autoscaling/cluster/uninstall/uninstall.go
```diff
@@ -0,0 +1,529 @@
...
+			result, err := cli.EC2.DescribeInstances(ctx, &ec2.DescribeInstancesInput{
+				Filters: []ec2types.Filter{
+					{
+						Name:   awssdk.String("tag:kubernetes.io/cluster/" + clusterName),
+						Values: []string{"owned"},
+					},
+					{
+						Name:   awssdk.String("tag-key"),
+						Values: []string{"karpenter.sh/nodepool"},
+					},
```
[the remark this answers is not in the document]
L3n41c: Fixed in https://github.com/DataDog/datadog-operator/commit/f8616f7585906d00ed9d7831547236b20927a49e#diff-a42350cb037bad7adef8e1396310575c39180e73ea1181e182fa3179a7464f32R428-R429

cmd/kubectl-datadog/autoscaling/cluster/common/aws/aws-auth.go
L3n41c: That’s much better! Thanks!
Implemented in https://github.com/DataDog/datadog-operator/commit/f8616f7585906d00ed9d7831547236b20927a49e#diff-75e34b5bb02db7ae8e1635b717b48c7090c87a7f1a2d2a63b955847f3db9acd3

cmd/kubectl-datadog/autoscaling/cluster/common/display/display_test.go
L3n41c: Good idea!
Implemented in https://github.com/DataDog/datadog-operator/commit/f8616f7585906d00ed9d7831547236b20927a49e#diff-774caa6680aa940302346b336b4cb70b818f61b82b1cab7ecd107ff245fc540a

L3n41c:
@codex review

cmd/kubectl-datadog/autoscaling/cluster/common/k8s/object.go
L3n41c: This is inside the `Delete` function the goal of which is to ensure that the given object is eventually not existing anymore.
If the given object cannot be found, it is already not existing, then the system is already in the expected state.
The promise of the `Delete` function, which is: “The object will eventually be deleted” is already kept.
Such a situation shouldn’t be considered as an error.

The question might then be: why would we try to delete an object that doesn’t exist?
Well, this might happen if a user tries to manually delete some objects concurrently to the execution of this script.
One could argue that it isn’t a good idea to do manual actions that might conflict with a running script.
But if we can handle this case gracefully, that’s still better.

cmd/kubectl-datadog/autoscaling/cluster/uninstall/uninstall.go
L3n41c: As far as cleanup is concerned, I preferred to have a best-effort approach, trying to clean up as much things as possible.
The errors won’t be lost and they will eventually be logged at the end.

I wanted to avoid that a failure on one step prevents other steps.

Let’s imagine a situation where the user has broken their Kubernetes cluster so that the API server isn’t reachable anymore and they want to destroy their broken cluster and to cleanup everything.

If the Kubernetes cluster is so broken that the API server isn’t reachable, we won’t be able to list and delete the Karpenter node pools.
This isn’t a big deal because the user will destroy the whole cluster anyway.
Yet, we still need to destroy the CloudFormation stacks that we created.
If our “uninstall” command exits at the first error, it means that it won’t be helpful to clean AWS objects when the K8s cluster isn’t reachable.

We can also imagine a situation where a user first deletes their Kubernetes cluster and then realizes afterwards that `kubectl datadog autoscaling cluster uninstall` still needs to be executed to clean AWS objects up.
Then, this command should be able to delete the AWS objects even if the K8s cluster isn’t reachable anymore.

That’s why I think that a failure at one cleanup step shouldn’t prevent the other steps from running.
Otherwise, there will be some situations where the user won’t be able to use that script to clean things up.

cmd/kubectl-datadog/autoscaling/cluster/uninstall/uninstall.go
L3n41c: I initially changed the behavior in https://github.com/DataDog/datadog-operator/commit/f8616f7585906d00ed9d7831547236b20927a49e#diff-a42350cb037bad7adef8e1396310575c39180e73ea1181e182fa3179a7464f32 to make the script abort if it’s unable to list stuff to delete.

But the more I think about it, the more I think it’s a mistake.
As explained in my above comment, I want this `uninstall` command to still be useful for users to clean up the AWS objects even if they deleted their Kubernetes cluster first.

L3n41c:
@codex review

fanny-jiang (approved):

PR #2380, opened 2025-12-05
Levan m/dca ccr reconciler refactor

levan-m:
### What does this PR do?

* [6c855b1](https://github.com/DataDog/datadog-operator/commit/6c855b18e78e579e1cd5f9c654f4c72d9683cb2e) resolves some seemingly accidental differences: feature error handling; passing ctx; introspection provider; error handling in cleanupV2ClusterChecksRunner. Mostly following DCA pattern in most cases. Also adds a test to cover these changes up front. 
* [80db14a](https://github.com/DataDog/datadog-operator/commit/80db14a030363cc4edf0d7a73d209336c0f832b7) moved `cleanupOldCCR**` function to file where it's used. Changed some log lines, variable names.
* [619d4e0](https://github.com/DataDog/datadog-operator/commit/619d4e0fcc1e4152aabbbe2693ce238b53ef16ab) pulled some of the functions, variable so they can be parametrized later.
* [bb24086](https://github.com/DataDog/datadog-operator/pull/2380/commits/bb24086cec21fc75df3d75aeacf1cac9e9ce920d) moved status cleanup inside component cleanup which was case with CCR, not DCA. This simplifies code but seems reasonable as we want to clear status only when component is cleaned up not before that.
* [043a3cc](https://github.com/DataDog/datadog-operator/pull/2380/commits/043a3cc606eac5d44b27c1d6728b7a259b5a0b9e) move custom logic outside and added no-op function where necessary. reconcile and cleanup should be same, only using different functions which probably can be parametrized or extracted in interface - which would be next step.
* [77e27a5](https://github.com/DataDog/datadog-operator/pull/2380/changes/77e27a5554b53ccae3f8daeecafd554271567a0d) fixed function name and uncommented tests.

### Motivation

What inspired you to submit this pull request?

### Additional Notes

Anything else we should know when reviewing?

### Minimum Agent Versions

Are there minimum versions of the Datadog Agent and/or Cluster Agent required?

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Write there any instructions and details you may have to test your PR.

### Checklist

- [ ] PR has at least one valid label: `bug`, `enhancement`, `refactoring`, `documentation`, `tooling`, and/or `dependencies`
- [ ] PR has a milestone or the `qa/skip-qa` label

PR #2434, opened 2025-12-26
Component reconcile refactor

levan-m:
### What does this PR do?

Ultimate goal of this change is to move `Reconcile` and `Cleanup` to ComponentRegistry and remove those from the `ComponentReconciler` interface. Assumption is that these two should be _almost_ same for all components (except maybe Agent). Two hooks are still necessary: 1) for cleaning up DCA RBAC 2) deleting CLC if DCA is disabled. These are added to the interface.

ComponentReconciler/Registry refactor following approach in https://github.com/DataDog/datadog-operator/pull/2380. 

* https://github.com/DataDog/datadog-operator/pull/2434/commits/eb5b1774128cdfce3c2a8ea78f96de6390f984c5 carries over changes from #2380. Mostly affect DCA, CLC components; also adds `controller_reconcile_deployment_test.go` to assert on existing behavior.
* https://github.com/DataDog/datadog-operator/pull/2434/commits/e66e910decc31cc9121aab0d4ee31cf4b615306d adds DDAI support in `controller_reconcile_deployment_test.go` but disables as tests fail. There seems some variation between DDA/DDAI components, hence this change can't be drop-in replacement of DDAI components. 
* https://github.com/DataDog/datadog-operator/pull/2434/commits/3ea55d3d15221478346491da9d51bd4bf97ef32f function signature changes to align DCA and CLC components.
* https://github.com/DataDog/datadog-operator/pull/2434/commits/b96dc886f6e4e222a4d67cce80e864a02e618925 **This should be only functional change in the PR.** splits `cleanupV2***` functions in two parts: 1) deleting deployment 2) updating status.  Regular reconcile flow executes both, `cleanupOld***` only deployment deletion. Reasoning is that deleting old/stale deployments after rename shouldn't cleanup status on DDA.
* https://github.com/DataDog/datadog-operator/pull/2434/commits/1f6f9e68f8a6820956df587e3f647ebad1030a47 move deployment deletion part to `controller_reconcile_v2_helpers.go` as this is not component specific any more.
* https://github.com/DataDog/datadog-operator/pull/2434/commits/04a98b68084717909464a7599e6e5ae2c19fafc5 adds function getters to interface so they can be used in `ComponentRegistry`, some minor cleanup.
* https://github.com/DataDog/datadog-operator/pull/2434/commits/f9eb06463ab95c63db64da6fbcd302df9d688a5b moves `ComponentReconcilerReconcile()` from components to a common one in `ComponentRegistry`, removes interface function.
* https://github.com/DataDog/datadog-operator/pull/2434/commits/17063f831549c5bc161c0fd36fae976bf73f7a6f does same for `Cleanup`.

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist

### DIFF
--- a/internal/controller/datadogagent/component_reconciler.go
+++ b/internal/controller/datadogagent/component_reconciler.go
@@ -131,7 +131,7 @@ func (r *ComponentRegistry) Register(component ComponentReconciler) {
 // ReconcileComponents reconciles all registered components in order
 func (r *ComponentRegistry) ReconcileComponents(ctx context.Context, params *ReconcileComponentParams) (reconcile.Result, error) {
 	var result reconcile.Result
-
+	now := metav1.NewTime(time.Now())
 	for _, comp := range r.components {
 		// Check if component is enabled and if there's a conflict
 		enabled, conflict := comp.IsEnabled(params.RequiredComponents, params.DDA.Spec.Override)
@@ -145,7 +145,7 @@ func (r *ComponentRegistry) ReconcileComponents(ctx context.Context, params *Rec
 				// Set conflict status condition
 				condition.UpdateDatadogAgentStatusConditions(
 					params.Status,
-					metav1.NewTime(time.Now()),
+					now,
 					common.OverrideReconcileConflictConditionType,
 					metav1.ConditionTrue,
 					"OverrideConflict",


internal/controller/datadogagent/component_reconciler.go
```diff
@@ -72,70 +93,118 @@ func (r *ComponentRegistry) Register(component ComponentReconciler) {
...
+
+	// If Override is defined for the component, apply the override on the PodTemplateSpec, it will cascade to container.
+	if componentOverride, ok := params.DDA.Spec.Override[component.Name()]; ok {
+		if apiutils.BoolValue(componentOverride.Disabled) {
+			if componentEnabled {
 				// The override supersedes what's set in requiredComponents; update status to reflect the conflict
 				condition.UpdateDatadogAgentStatusConditions(
 					params.Status,
-					now,
+					metav1.NewTime(time.Now()),
```
tbavelier: Is this intended instead of using the function-level `now` ?

internal/controller/datadogagent/component_reconciler.go
tbavelier, line 1: We have an issue here: if the **component override struct is not empty but not explicitly disabled**, we reconcile a component that should not be possibly (CCR is not enabled by default).
```yaml
  features:
    clusterChecks:
      enabled: true
      useClusterChecksRunners: false
  override:
    clusterChecksRunner:
      disabled: true
      containers:
        agent:
          resources:
            requests: 256m
```
-> we get a CCR deployment (no pods scheduled cuz missing dependencies but the component shouldnt be reconciled at all)

We could early exit instead of further reconciling:
```go
	// Explicit override disable always wins (and may create a conflict condition if the component is otherwise enabled).
	if ok && apiutils.BoolValue(componentOverride.Disabled) {
		if componentEnabled {
			// The override supersedes what's set in requiredComponents; update status to reflect the conflict
			condition.UpdateDatadogAgentStatusConditions(
				params.Status,
				now,
				common.OverrideReconcileConflictConditionType,
				metav1.ConditionTrue,
				"OverrideConflict",
				fmt.Sprintf("%s component is set to disabled", component.Name()),
				true,
			)
		}
		return r.Cleanup(ctx, params, component)
	}

	// If the component isn't enabled, we should cleanup regardless of whether an override struct exists.
	// (Overrides should not implicitly enable a component.)
	if !componentEnabled {
		return r.Cleanup(ctx, params, component)
	}

	// Apply non-disable overrides.
	if ok {
		override.PodTemplateSpec(params.Logger, podManagers, componentOverride, component.Name(), params.DDA.Name)
		override.Deployment(deployment, componentOverride)
	}
```

internal/controller/datadogagent/component_reconciler.go
```diff
@@ -72,70 +93,118 @@ func (r *ComponentRegistry) Register(component ComponentReconciler) {
...
+	var result reconcile.Result
+	now := metav1.NewTime(time.Now())
+	deploymentLogger := params.Logger.WithValues("component", component.Name())
+
+	// Start by creating the Default Cluster-Agent deployment
+	deployment := component.GetNewDeploymentFunc()(params.DDA.GetObjectMeta(), &params.DDA.Spec)
+	podManagers := feature.NewPodTemplateManagers(&deployment.Spec.Template)
+
+	// Set Global setting on the default deployment
+	component.GetGlobalSettingsFunc()(params.Logger, podManagers, params.DDA.GetObjectMeta(), &params.DDA.Spec, params.ResourceManagers, params.RequiredComponents)
```
tbavelier:
```suggestion
	component.GetGlobalSettingsFunc()(deploymentLogger, podManagers, params.DDA.GetObjectMeta(), &params.DDA.Spec, params.ResourceManagers, params.RequiredComponents)
```

internal/controller/datadogagent/component_reconciler.go
```diff
@@ -72,70 +93,118 @@ func (r *ComponentRegistry) Register(component ComponentReconciler) {
...
-
-			componentLogger.V(1).Info("Component disabled, cleaning up")
-			res, err := comp.Cleanup(ctx, &compParams)
-			if utils.ShouldReturn(res, err) {
-				return res, err
-			}
-			continue
+			return r.Cleanup(ctx, params, component)
 		}
+		override.PodTemplateSpec(params.Logger, podManagers, componentOverride, component.Name(), params.DDA.Name)
```
tbavelier:
```suggestion
		override.PodTemplateSpec(deploymentLogger, podManagers, componentOverride, component.Name(), params.DDA.Name)
```

internal/controller/datadogagent/component_reconciler.go
levan-m: Good find. Yea implemented similar logic but in a bit different way. I separated cleanup (except forceCleanup) from creation/update and moved in `ReconcileComponent`  in f9072feaca239e8a5a0b263ce87371896c30f927. Also added test to cover the case.

internal/controller/datadogagent/component_reconciler.go
levan-m: Since this is moved to a different function I left as is, don't see reason to have same in all functions and cleanup is entirely separate code branch anyway.

internal/controller/datadogagent/component_reconciler.go
levan-m: fixed in f9072feaca239e8a5a0b263ce87371896c30f927

internal/controller/datadogagent/component_reconciler.go
```diff
@@ -72,70 +131,124 @@ func (r *ComponentRegistry) Register(component ComponentReconciler) {
...
+		var err error
+
+		if !enabled {
+			// Component is disabled, clean it up
+			if conflict {
+				// Set conflict status condition
 				condition.UpdateDatadogAgentStatusConditions(
 					params.Status,
-					now,
+					metav1.NewTime(time.Now()),
```
tbavelier:
```suggestion
					now,
```
with `now := metav1.NewTime(time.Now())` before the loop ? Doesn't matter much, just happens in case of conflict, but just to be "consistent" with other places ?

PR #2455, opened 2026-01-05
Copy finalizer when updating DDAI

khewonc:
### What does this PR do?

When updating the DDAI, we remove all finalizers. This manually adds the finalizers back in if they are present
This happens when a DDAI is updated, then deleted before the finalizer is added back in

### Motivation

Error logs in metrics forwarder

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Difficult to reproduce the issue. Requires update + deletion of DDAI to happen very close together, which is uncommon. In previous operator versions (with the issue), in the logs, we should see:
* log about `updating DatadogAgentInternal`
* NO log about `Adding Finalizer for the DatadogAgentInternal`
* log about `Deleting unused DDAI`
* logs with the error `DatadogAgentInternal.datadoghq.com "<ddai-name>" not found` in the metrics forwarder periodically

With the fix, we shouldn't see the error log. On every reconcile requiring a DDAI update, this debug log should be present though: `copying finalizers from the current DDAI`

### Checklist


khewonc:
Closing in favor of https://github.com/DataDog/datadog-operator/pull/2463

PR #2468, opened 2026-01-07
[AGENTONB-2760] use UUID instead of HostName in metadata payload

swang392:
### What does this PR do?

* sobotka requires either `UUID` or `HostName` present to send metadata payloads through it to REDAPL. 
* This PR switches from `hostName` to UUID. This makes more sense because metadata is not sent per host, rather it is sent once per cluster. 
* `UUID` is set to `clusterUID`

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

* Deploy the operator
* Verify that metadata forwarders are starting and sending metadata as expected

Tested on a staging cluster (`gizmo`), and verified that the payload is being successfully sent -- the `modification_detected_at` field in DDSQL has been updated after the test image was deployed. 
### Checklist

### DIFF
--- a/pkg/controller/utils/metadata/credential_setup_test.go
+++ b/pkg/controller/utils/metadata/credential_setup_test.go
@@ -58,7 +58,6 @@ func TestSetupRequestPrerequisites(t *testing.T) {
 			setupEnv: func() {
 				os.Setenv(constants.DDAPIKey, "operator-api-key")
 				os.Setenv(constants.DDAppKey, "operator-app-key")
-				os.Setenv(constants.DDHostName, "test-hostname")
 				os.Setenv(constants.DDClusterName, "test-cluster")
 			},
 			setupDDA: func() []client.Object {
@@ -107,7 +106,6 @@ func TestSetupRequestPrerequisites(t *testing.T) {
 		{
 			name: "DDA with plaintext API key and default site",
 			setupEnv: func() {
-				os.Setenv(constants.DDHostName, "test-hostname")
 				// No operator credentials
 			},
 			setupDDA: func() []client.Object {
@@ -136,7 +134,6 @@ func TestSetupRequestPrerequisites(t *testing.T) {
 		{
 			name: "DDA with API key and custom site",
 			setupEnv: func() {
-				os.Setenv(constants.DDHostName, "test-hostname")
 				// No operator credentials
 			},
 			setupDDA: func() []client.Object {
@@ -166,7 +163,6 @@ func TestSetupRequestPrerequisites(t *testing.T) {
 		{
 			name: "DDA with secret reference",
 			setupEnv: func() {
-				os.Setenv(constants.DDHostName, "test-hostname")
 				// No operator credentials
 			},
 			setupDDA: func() []client.Object {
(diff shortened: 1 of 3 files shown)


pkg/controller/utils/metadata/helm_metadata.go
```diff
@@ -50,16 +50,18 @@ type HelmMetadataForwarder struct {
 }
 
 type HelmMetadataPayload struct {
-	Hostname  string       `json:"hostname"`
+	UUID      string       `json:"uuid"`
```
levan-m, line 53: Does this require schema or writer change?

pkg/controller/utils/metadata/operator_metadata.go
tbavelier, line 1: Not directly related to this PR, but we have a potential data race in `func (omf *OperatorMetadataForwarder) GetPayload(clusterUID string)`.
In `OperatorMetadataForwarder.GetPayload()`, we use `omf.mutex.RLock()` (read lock) and then mutate shared state `omf.OperatorMetadata.*`.
* Multiple goroutines can hold an RLock() at the same time, so two callers could write to `omf.OperatorMetadata` concurrently.
* `OperatorMetadata` contains `ResourceCounts map[string]int` with maps not being safe for concurrent access if anyone might write.

A safe approach could be: copy under lock, unlock, then mutate the local copy and marshal
```go
	omf.mutex.RLock()
	// Copy metadata while holding the lock, to avoid data races.
	operatorMetadata := omf.OperatorMetadata
	if omf.OperatorMetadata.ResourceCounts != nil {
		operatorMetadata.ResourceCounts = make(map[string]int, len(omf.OperatorMetadata.ResourceCounts))
		maps.Copy(operatorMetadata.ResourceCounts, omf.OperatorMetadata.ResourceCounts)
	}
	omf.mutex.RUnlock()
```

pkg/controller/utils/metadata/helm_metadata.go
swang392: No, the `dd-go` backend creates a schema resource based on the `HelmMetadataPayload` object `HelmMetadata` (`datadog_operator_helm_metadata`) and the schema matches that struct.

swang392:
> We have payload.clusterUID, payload.UUID and payload.metadata.clusterUID for the same value, can we possibly remove clusterID from `OperatorMetadata`, `CRDMetadata` and `HelmMetadata` structs ? Also, possibly remove it from payload if it's not necessary in the schema at intake to only keep UUID ? (since it can be derived from UUID ?)

I think these will need to be kept since each instance of clusterID is needed for something different. `payload.UUID` is needed by sobotka, `payload.clusterUID` is needed in `dd-go` to create the metadata resource from `payload.metadata`, and it can't be derived from `payload.UUID`. FA needs `payload.metadata.clusterUID` for the foreign key. 

One possible option would be to rely on `payload.metadata.clusterUID` to create the resource instead of `payload.clusterUID`, but that would introduce inconsistency between how cluster-level metadata payloads are created between the operator and cluster agent.

swang392:
> More generally (and outside the PR scope), we repeat a lot of OperatorVersion and KubernetesVersion keys in the structs, would it be possible to reduce duplication by using something like:

yep, I have a [card](https://datadoghq.atlassian.net/browse/AGENTONB-2764) here to refactor sharedMetadata and reduce duplication.

pkg/controller/utils/metadata/operator_metadata.go
swang392: addressed in a separate PR: https://github.com/DataDog/datadog-operator/pull/2481

PR #2481, opened 2026-01-12
Fix race condition in operator metadata

swang392:
### What does this PR do?

Fix race condition in the operator metadata forwarder. There are multiple goroutines (one to update resource count, one to send). Previously multiple goroutines could lock anbd modify data at once.

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

- Run test ` go test ./pkg/controller/utils/metadata -run Test_GetPayload_Concurrent -count=50 -race` from this branch
- Check out main and copy the `Test_GetPayload_Concurrent` to a branch without the fix, rerun. The test should fail with race conditions. 

### Checklist

### DIFF
--- a/pkg/controller/utils/metadata/operator_metadata_test.go
+++ b/pkg/controller/utils/metadata/operator_metadata_test.go
@@ -364,14 +364,17 @@ func Test_GetPayload_Concurrent(t *testing.T) {
 		}(i)
 	}
 
+	updateDone := make(chan bool, 1)
 	go func() {
-		for i := 0; i < 100; i++ {
+		for i := 0; i < 10; i++ {
 			omf.updateResourceCounts()
 		}
+		updateDone <- true
 	}()
 
 	// Wait for all goroutines to complete
 	for i := 0; i < numGoroutines; i++ {
 		<-done
 	}
+	<-updateDone
 }


pkg/controller/utils/metadata/operator_metadata.go
```diff
@@ -124,17 +124,25 @@ func (omf *OperatorMetadataForwarder) GetPayload(clusterUID string) []byte {
...
 	omf.mutex.RLock()
-	defer omf.mutex.RUnlock()
+	// Copy metadata while holding the lock to avoid data races
+	operatorMetadata := omf.OperatorMetadata
+	if omf.OperatorMetadata.ResourceCounts != nil {
+		operatorMetadata.ResourceCounts = make(map[string]int, len(omf.OperatorMetadata.ResourceCounts))
+		for k, v := range omf.OperatorMetadata.ResourceCounts {
+			operatorMetadata.ResourceCounts[k] = v
+		}
+	}
```
tbavelier, line 133: Think we can use `maps.Copy` there to avoid the loop

pkg/controller/utils/metadata/operator_metadata_test.go
```diff
@@ -321,3 +322,59 @@ func Test_GetPayload(t *testing.T) {
...
+	go func() {
+		for i := 0; i < 1000; i++ {
+			omf.mutex.Lock()
+			if omf.OperatorMetadata.ResourceCounts == nil {
+				omf.OperatorMetadata.ResourceCounts = make(map[string]int)
+			}
+			omf.OperatorMetadata.ResourceCounts["test-resource"] = i
+			omf.mutex.Unlock()
+		}
+	}()
```
tbavelier, line 373: Is there a way to use the "real" goroutine that does this update or it's too much work to mock/not worth it ?

swang392: updated in [8b31683](https://github.com/DataDog/datadog-operator/pull/2481/commits/8b316839ea5e8e3c85b6d89410553093a90c1935)

pkg/controller/utils/metadata/operator_metadata.go
swang392: updated in [8b31683](https://github.com/DataDog/datadog-operator/pull/2481/commits/8b316839ea5e8e3c85b6d89410553093a90c1935)

pkg/controller/utils/metadata/operator_metadata_test.go
```diff
@@ -321,3 +322,59 @@ func Test_GetPayload(t *testing.T) {
 		t.Errorf("GetPayload() resource_count is not a map, got: %T", metadata["resource_count"])
 	}
 }
+
+// Test that GetPayload is safe for concurrent access (no data races)
+func Test_GetPayload_Concurrent(t *testing.T) {
```
levan-m, line 327: nice!

PR #2483, opened 2026-01-13
[OTAGENT-512] Add OTel Agent Gateway feature implementation

songy23:
### What does this PR do?

Add OTel Agent Gateway feature implementation. Reference implementation in helm charts: http://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/otel-agent-gateway-deployment.yaml

### Motivation

Support OTel Agent Gateway in operator. See how it currently works in helm charts: https://docs.datadoghq.com/opentelemetry/setup/ddot_collector/install/kubernetes_gateway/

### Additional Notes

HPA is currently not natively supported in operator, we may require users to define HPA as a separate resource

Replicas, node selector, affinity and other configs will be supported in the next PR.

### Minimum Agent Versions

* Agent: v7.74.0

### Describe your test plan

Built the operator locally and deployed to a kind cluster with OTel Agent Gateway enabled.

For QA, use the following config:
```
kind: DatadogAgent
metadata:
  name: datadog-otel-test
  namespace: datadog
spec:
  global:
    clusterName: kind-cluster
    credentials: ...
    site: datadoghq.com
  features:
    # Enable OTel Agent Gateway feature (standalone gateway deployment)
    otelAgentGateway:
      enabled: true
      # Use default ports: 4317 (gRPC) and 4318 (HTTP)

    # Enable OTel Collector feature (runs in each node agent pod)
    otelCollector:
      enabled: true
      # Collector will send data to the gateway service
```

Verify:
1. k8s deployment `datadog-otel-test-otel-agent-gateway` is created
2. k8s service `datadog-otel-test-otel-agent-gateway` is created and listening on 4317/TCP, 4318/TCP 
3. pod `datadog-otel-test-otel-agent-gateway-<pod>` is created and running
4. k8s endpoint `datadog-otel-test-otel-agent-gateway` binds to the pod above
5. the `otel-agent` container in the node agent pod has a config that sends to the k8s service above
6. cluster role `datadog-otel-test-otel-agent-gateway` and its binding are created

Additionally, deploy the OTel test client below. Ideally you have a multi-node cluster and deploy the client to a different node than the one that gateway runs on, to test the cross-node routing.
```
apiVersion: v1
kind: Pod
metadata:
  name: otel-test-client
  namespace: datadog
spec:
  nodeSelector:
    kubernetes.io/hostname: kind-worker # replace if needed to make sure test client runs on a different node
  containers:
  - name: telemetrygen
    image: ghcr.io/open-telemetry/opentelemetry-collector-contrib/telemetrygen:latest
    command:
    - /telemetrygen
    args:
    - traces
    - --otlp-endpoint=datadog-otel-test-agent.datadog.svc.cluster.local:4317
    - --otlp-insecure
    - --duration=60s
    - --rate=1
    - --service=test-service
  restartPolicy: Never
```

It sends OTLP data to the otel-agent in the node agent pod (NOT the otel agent gateway): datadog-otel-test-agent 4317. Then the otel-agent in the node agent pod forwards the data to the otel agent gateway, and gateway sends the data eventually to DD.

### Checklist

### DIFF
--- a/internal/controller/datadogagent/component/otelagentgateway/default.go
+++ b/internal/controller/datadogagent/component/otelagentgateway/default.go
@@ -64,8 +64,8 @@ func defaultPodSpec(dda metav1.Object) corev1.PodSpec {
 			{
 				Name:    "init-volume",
 				Image:   images.GetLatestDdotCollectorImage(),
-				Command: []string{"cp", "-r"},
-				Args:    []string{"/etc/datadog-agent", "/opt"},
+				Command: []string{"bash", "-c"},
+				Args:    []string{"cp -vnr /etc/datadog-agent /opt"},
 				VolumeMounts: []corev1.VolumeMount{
 					{
 						Name:      common.ConfigVolumeName,
--- a/internal/controller/datadogagent/component/otelagentgateway/rbac.go
+++ b/internal/controller/datadogagent/component/otelagentgateway/rbac.go
@@ -8,6 +8,8 @@ package otelagentgateway
 import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/DataDog/datadog-operator/pkg/kubernetes/rbac"
 )
 
 // RBAC for OTel Agent Gateway
@@ -17,37 +19,26 @@ import (
 func GetDefaultOtelAgentGatewayClusterRolePolicyRules(dda metav1.Object, excludeNonResourceRules bool) []rbacv1.PolicyRule {
 	policyRule := []rbacv1.PolicyRule{
 		{
-			APIGroups: []string{""},
+			APIGroups: []string{rbac.CoreAPIGroup},
 			Resources: []string{
-				"pods",
-				"namespaces",
+				rbac.PodsResource,
+				rbac.NamespaceResource,
 			},
 			Verbs: []string{
-				"get",
-				"watch",
-				"list",
+				rbac.GetVerb,
+				rbac.WatchVerb,
+				rbac.ListVerb,
 			},
 		},
 		{
-			APIGroups: []string{"apps"},
+			APIGroups: []string{rbac.AppsAPIGroup},
 			Resources: []string{
-				"replicasets",
+				rbac.ReplicasetsResource,
 			},
 			Verbs: []string{
-				"get",
-				"list",
-				"watch",
-			},
-		},
-		{
-			APIGroups: []string{"extensions"},
-			Resources: []string{
-				"replicasets",
-			},
-			Verbs: []string{
-				"get",
-				"list",
-				"watch",
+				rbac.GetVerb,
+				rbac.ListVerb,
+				rbac.WatchVerb,
 			},
 		},
 	}
--- a/internal/controller/datadogagent/feature/otelagentgateway/feature.go
+++ b/internal/controller/datadogagent/feature/otelagentgateway/feature.go
@@ -99,22 +99,22 @@ func (f *otelAgentGatewayFeature) Configure(dda metav1.Object, ddaSpec *v2alpha1
 	return reqComp
 }
 
-func (o *otelAgentGatewayFeature) buildOTelAgentCoreConfigMap() (*corev1.ConfigMap, error) {
-	if o.customConfig != nil && o.customConfig.ConfigData != nil {
-		cm, err := configmap.BuildConfigMapConfigData(o.owner.GetNamespace(), o.customConfig.ConfigData, o.configMapName, otelConfigFileName)
+func (f *otelAgentGatewayFeature) buildOTelAgentCoreConfigMap() (*corev1.ConfigMap, error) {
+	if f.customConfig != nil && f.customConfig.ConfigData != nil {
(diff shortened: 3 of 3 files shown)


pkg/images/images.go
```diff
@@ -125,7 +125,7 @@ func GetLatestAgentImage() string {
 	return image.ToString()
 }
 
-// GetLatestAgentImage returns the latest ddot collector image
+// GetLatestDdotCollectorImage returns the latest ddot collector image
```
tbavelier, line 130: Thanks!

internal/controller/datadogagent/component/otelagentgateway/rbac.go
tbavelier, line 1: Please use the constants from `pkg/kubernetes/rbac/const.go` instead of hard-coding apigroups, resources, verbs. etc.: e.g. `apps` is `AppsAPIGroup`, "" is `CoreAPIGroup`

internal/controller/datadogagent/feature/otelagentgateway/feature.go
```diff
@@ -38,10 +60,132 @@ func (f *otelAgentGatewayFeature) ID() feature.IDType {
...
+	internalTrafficPolicy := corev1.ServiceInternalTrafficPolicyLocal
+	if err := managers.ServiceManager().AddService(
+		f.localServiceName,
+		f.owner.GetNamespace(),
+		common.GetOtelAgentGatewayServiceSelector(f.owner),
+		[]corev1.ServicePort{*otlpGrpcPort, *otlpHttpPort},
+		&internalTrafficPolicy,
+	); err != nil {
+		return err
+	}
```
tbavelier: Pretty sure you do NOT want to use an internal traffic policy, that's intended for daemonset componnents, as that would cause pods to not see any endpoints if they're not on the same node as the gateway deployment. You wouldn't notice on a 1-node kind cluster since everything is on the same node : https://kubernetes.io/docs/concepts/services-networking/service-traffic-policy/#how-it-works
So it should remain as nil, and you'd need to update QA instructions to test this scenario properly (2 nodes cluster, deploy client on non gateway node, etc.)

internal/controller/datadogagent/feature/otelagentgateway/feature.go
```diff
@@ -38,10 +60,132 @@ func (f *otelAgentGatewayFeature) ID() feature.IDType {
...
+	if ddaSpec.Features.OtelAgentGateway.Conf != nil {
+		f.customConfig = ddaSpec.Features.OtelAgentGateway.Conf
+	}
+	f.configMapName = constants.GetConfName(dda, f.customConfig, defaultOTelAgentGatewayConf)
+
 	return reqComp
 }
 
-func (f *otelAgentGatewayFeature) ManageDependencies(feature.ResourceManagers, string) error {
+func (o *otelAgentGatewayFeature) buildOTelAgentCoreConfigMap() (*corev1.ConfigMap, error) {
```
tbavelier: No biggy but can we use f as the receiver for consistency ? Also the function name is the same as for otel collector, I assume due to copy-paste, is that expected ?

internal/controller/datadogagent/component/otelagentgateway/default.go
```diff
@@ -60,13 +60,33 @@ func NewDefaultOtelAgentGatewayPodTemplateSpec(dda metav1.Object) *corev1.PodTem
 func defaultPodSpec(dda metav1.Object) corev1.PodSpec {
 	return corev1.PodSpec{
 		ServiceAccountName: GetOtelAgentGatewayRbacResourcesName(dda),
+		InitContainers: []corev1.Container{
+			{
+				Name:    "init-volume",
+				Image:   images.GetLatestDdotCollectorImage(),
+				Command: []string{"cp", "-r"},
```
tbavelier: Not blocking:
❓ This matches the Helm chart, but the "pattern" in Agent for init containers is to bash -c + cp. No strong opinion tho there, not sure about the subtleties it implies.
Also we use -vnr in Agent for verbosity.

internal/controller/datadogagent/component/otelagentgateway/rbac.go
songy23: Sure, replaced all

internal/controller/datadogagent/feature/otelagentgateway/feature.go
songy23: Good call - yes the gateway should use cluster traffic (the default). I've fixed this and tested it's working in a 3-node kind cluster

internal/controller/datadogagent/feature/otelagentgateway/feature.go
songy23: Sure, and yes this function (otel config) works the same way in otel collector (node agent) and otel gateway

internal/controller/datadogagent/component/otelagentgateway/default.go
songy23: I've changed it to match the other components (agent, CCR). The only difference I see is it won't overwrite the same file so a small improvement

internal/controller/datadogagent/feature/otelagentgateway/feature.go
songy23: Updated QA too

PR #2492, opened 2026-01-14
Refactor controller_v2_test, test DDA+DDAI path

levan-m:
### What does this PR do?

Mostly non-functional change:
* Create common test case `struct`.
* Extracts test execution and some common code (like client creation) into separate function.
* Change helpers to assert instead of returning error.
* Otel test assertion fixes.
* Added `focus` field to test case to ease debugging of a single test. 

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist

### DIFF
--- a/internal/controller/datadogagent/controller_v2_test.go
+++ b/internal/controller/datadogagent/controller_v2_test.go
@@ -20,6 +20,7 @@ import (
 	common "github.com/DataDog/datadog-operator/internal/controller/datadogagent/common"
 	"github.com/DataDog/datadog-operator/internal/controller/datadogagent/experimental"
 	agenttestutils "github.com/DataDog/datadog-operator/internal/controller/datadogagent/testutils"
+	"github.com/DataDog/datadog-operator/internal/controller/datadogagentinternal"
 	"github.com/DataDog/datadog-operator/pkg/condition"
 	"github.com/DataDog/datadog-operator/pkg/constants"
 	"github.com/DataDog/datadog-operator/pkg/images"
@@ -31,6 +32,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/tools/record"
@@ -59,7 +61,7 @@ type testCase struct {
 
 // runTestCases runs test cases, respecting the focus field for debugging.
 // If any test has focus=true, only focused tests run. Otherwise all tests run.
-func runTestCases(t *testing.T, tests []testCase) {
+func runTestCases(t *testing.T, tests []testCase, testFunc func(t *testing.T, tt testCase, opts ReconcilerOptions)) {
 	// Check if any test is focused
 	hasFocused := false
 	// for _, tt := range tests {
@@ -84,13 +86,13 @@ func runTestCases(t *testing.T, tests []testCase) {
 				IntrospectionEnabled:        tt.introspectionEnabled,
 			}
 
-			runReconcilerTest(t, tt, testOpts)
+			testFunc(t, tt, testOpts)
 		})
 	}
 }
 
-// runReconcilerTest is a common test runner that executes reconciliation tests
-func runReconcilerTest(t *testing.T, tt testCase, opts ReconcilerOptions) {
+// runDDAReconcilerTest runs test case using only the DDA reconciler
+func runDDAReconcilerTest(t *testing.T, tt testCase, opts ReconcilerOptions) {
 	t.Helper()
 
 	logf.SetLogger(zap.New(zap.UseDevMode(true)))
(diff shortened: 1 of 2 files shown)


internal/controller/datadogagent/controller_v2_test.go
```diff
@@ -45,44 +41,141 @@ import (
...
+	dda                  *v2alpha1.DatadogAgent
+	nodes                []client.Object
+	want                 reconcile.Result
+	wantErr              bool
+	wantFunc             func(t *testing.T, c client.Client)
+	profile              *v1alpha1.DatadogAgentProfile // For DDAI tests
+	profilesEnabled      bool                          // For DDAI tests
+	ddaiEnabled          bool                          // For DDAI tests
+	introspectionEnabled bool                          // For introspection tests
+	focus                bool                          // For debugging: run only focused tests if any are focused
```
tbavelier, line 59: Not a fan of having this focus flag and logic, could be "easy" to miss during a review when someone adds a new test with this flag, and then, the previous tests won't be run anymore. Debugging should be done by commenting the cases / using breakpoints / a debugger rather than these mechanics IMHO

internal/controller/datadogagent/controller_v2_test.go
```diff
@@ -951,16 +817,12 @@ func Test_AutopilotOverrides(t *testing.T) {
...
-				if err := verifyDaemonsetContainers(c, resourcesNamespace, dsName, expectedContainers); err != nil {
-					return err
-				}
+				verifyDaemonsetContainers(t, c, resourcesNamespace, dsName, expectedContainers)
 
 				ds := &appsv1.DaemonSet{}
-				if err := c.Get(context.TODO(), types.NamespacedName{Namespace: resourcesNamespace, Name: dsName}, ds); err != nil {
-					return err
-				}
+				c.Get(context.TODO(), types.NamespacedName{Namespace: resourcesNamespace, Name: dsName}, ds)
```
tbavelier:
```suggestion
				err: c.Get(context.TODO(), types.NamespacedName{Namespace: resourcesNamespace, Name: dsName}, ds)
				assert.NoError(t, err, "Failed to get DaemonSet %s/%s", resourcesNamespace, dsName)
```
Any reason we're not asserting error in getting the ds here unlike in the other places ?

internal/controller/datadogagent/controller_v2_test.go
```diff
@@ -1034,19 +896,15 @@ func Test_AutopilotOverrides(t *testing.T) {
...
```
tbavelier: Same as above on asserting the error

internal/controller/datadogagent/controller_v2_test.go
```diff
@@ -45,44 +41,141 @@ import (
...
 
-type fields struct {
-	client       client.Client
-	scheme       *runtime.Scheme
-	platformInfo kubernetes.PlatformInfo
-	recorder     record.EventRecorder
+type testCase struct {
+	name                 string
+	clientBuilder        *fake.ClientBuilder
+	loadFunc             func(c client.Client) *v2alpha1.DatadogAgent
```
tbavelier, line 49:
```suggestion
	loadFunc             func(t *testing.T, c client.Client) *v2alpha1.DatadogAgent
```
Same as wantFunc to assert we don't have an error when creating the object ? See further below. Tho we were not asserting errors on object loading before this refactor, but might as well add it ?
e.g.
```go
		{
			name: "DatadogAgent default, create Daemonset with core and trace agents",
			loadFunc: func(t *testing.T, c client.Client) *v2alpha1.DatadogAgent {
				dda := testutils.NewInitializedDatadogAgentBuilder(resourcesNamespace, resourcesName).
					Build()
				assert.NoError(t, c.Create(context.TODO(), dda))
				return dda
			},
```

internal/controller/datadogagent/controller_v2_test.go
```diff
@@ -45,44 +41,141 @@ import (
...
+	r.initializeComponentRegistry()
+
+	// Load or create DatadogAgent
+	var dda *v2alpha1.DatadogAgent
+	if tt.loadFunc != nil {
+		dda = tt.loadFunc(r.client)
+	} else if tt.dda != nil {
+		_ = r.client.Create(context.TODO(), tt.dda)
+		dda = tt.dda
+	}
```
tbavelier, line 128:
```suggestion
	if tt.loadFunc != nil {
		dda = tt.loadFunc(t, r.client)
	} else if tt.dda != nil {
		assert.NoError(t, r.client.Create(context.TODO(), tt.dda))
		dda = tt.dda
	}
```
Requires the loadFunc signature change

internal/controller/datadogagent/controller_v2_test.go
levan-m: From my experience, I have to debug individual tests often as they touch large part of code and not always same path so after doing comment/uncomment I decided to add it. I'll comment out code which actually makes `focus` evaluation so tests are always running to avoid scenario you mentioned. If someone uncomments it hopefully would be easier to spot.

internal/controller/datadogagent/controller_v2_test.go
levan-m: No this wasn't intentional, just oversight after switching from err handling to assertions.

internal/controller/datadogagent/controller_v2_test.go
levan-m: See comment ^^

internal/controller/datadogagent/controller_v2_test.go
levan-m: I don't see lack of assertions a big deal. `loadFunc` is tests setup not validating test outcomes so will as is.

PR #2508, opened 2026-01-20
[feature][autoscaling] Update RBAC allow watch for karpenter resources

jennchenn:
### What does this PR do?

Allow `watch` for `karpenter.sh` resources

### Motivation

The cluster agent controller requires this permission in order to live scale clusters.

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: v7.74.1

### Describe your test plan

1. Enable the cluster autoscaling feature
2. Check that the `datadog-cluster-agent-autoscaling` clusterrole has policy rule with `*.karpenter.sh` verb `watch`

### Checklist

### DIFF
--- a/internal/controller/datadogagent_controller.go
+++ b/internal/controller/datadogagent_controller.go
@@ -90,6 +90,8 @@ type DatadogAgentReconciler struct {
 // +kubebuilder:rbac:groups=*,resources=*/scale,verbs=get;update
 // +kubebuilder:rbac:groups=apps,resources=deployments,verbs=patch
 // +kubebuilder:rbac:groups=argoproj.io,resources=rollouts,verbs=patch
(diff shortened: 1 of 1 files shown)


clamoriniere (approved):

clamoriniere (commented):
You might also add the RBAC on the controller RBAC annotation to add automatically this RBAC to the generated manifest for Openshift and the OLM (Operator Lifecycle Manager) here: https://github.com/DataDog/datadog-operator/blob/0fc9a710f1431c576ff4e213331960d8a784a78e/internal/controller/datadogagent_controller.go#L86  

it is to generate the https://github.com/DataDog/datadog-operator/blob/main/config/rbac/role.yaml but it is already added by the k8s explorer it will not introduce change, however still better to properly define annotation RBAC

clamoriniere (approved):

PR #2551, opened 2026-01-29
[CXP-2639][datadog-operator] Remove envvar ovveride for controlling whether process checks run in core or process agent

daniel-taf:
### What does this PR do?
Remove envvar override for controlling whether process checks run in core or process agent. This behavior is automatically controlled via version checks. Related config is already deprecated: https://github.com/DataDog/datadog-operator/pull/2178

### Minimum Agent Versions

* Agent: v7.60

### Describe your test plan

Unit tests seem sufficient to me. However, for manual QA, one can deploy the operator with the envvar set to false on the relevant containers with an agent version > 7.60 and ensure the process agent comes up with the process check. Can be confirmed via logs and agent status.

### Checklist


go.work.sum
swang392, line 1: lgtm, but not sure if this file was supposed to be included

daniel-taf: not sure how that got in, removed

PR #2554, opened 2026-01-29
Use `monitorStateSyncStatus` for monitor k8s display column

loganjohnson-wk:
### What does this PR do?

Currently, the Datadog monitor object print column points to the `.status.syncStatus` which is no longer exists, changed in https://github.com/DataDog/datadog-operator/pull/1355. It should point to the `.status.monitorStateSyncStatus` so it can show up properly when viewing in k8s.

### Motivation

The sync status is just blank when doing a `kubectl get datadogmonitors` 😢 
image

It was causing some confusion when the "Monitor State" column would say "Alert" folks were assuming that the monitor was failing to sync vs. the monitor being in the alert state in Datadog. I think having both columns filled in properly will fix this.

### Additional Notes

When running `make generate-manifests` locally I get a vastly different result for the CRD output with different indentation and line wrapping, might be a local linting thing? Either way someone may want to double check the output, pretty sure it is correct though.

### Minimum Agent Versions

- No it's just changing k8s display columns

### Describe your test plan

Generate the CRD with `make generate-manifests` and deploy the CRD directly with `k apply -f config/crd/bases/v1/datadoghq.com_datadogmonitors.yaml` (we use helm to deploy our operator otherwise I would use the `make deploy`)

### Checklist

### DIFF
--- a/api/datadoghq/v1alpha1/datadogmonitor_types.go
+++ b/api/datadoghq/v1alpha1/datadogmonitor_types.go
@@ -376,7 +376,7 @@ type DatadogMonitorDowntimeStatus struct {
 // +kubebuilder:printcolumn:name="monitor state",type="string",JSONPath=".status.monitorState"
 // +kubebuilder:printcolumn:name="last state transition",type="string",JSONPath=".status.monitorStateLastTransitionTime"
 // +kubebuilder:printcolumn:name="last state sync",type="string",format="date",JSONPath=".status.monitorStateLastUpdateTime"
-// +kubebuilder:printcolumn:name="sync status",type="string",JSONPath=".status.syncStatus"
+// +kubebuilder:printcolumn:name="sync status",type="string",JSONPath=".status.monitorStateSyncStatus"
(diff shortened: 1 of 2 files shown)


loganjohnson-wk:
For the QA you can see the before and after when applying the updated CRD so the "SYNC STATUS" column is now populated.
image

tbavelier:
@loganjohnson-wk thank you for the contribution! Indeed, good catch. Verified locally as well.

> When running make generate-manifests locally I get a vastly different result for the CRD output with different indentation and line wrapping, might be a local linting thing? Either way someone may want to double check the output, pretty sure it is correct though.

You most likely have an outdated cached version of the binaries that generate the manifests. You can `rm -rf bin/` inside your copy of the repository, before running `make install-tools` and that should set you up correctly.

I don't remember if CI runs correctly on forks, so I created in parallel https://github.com/DataDog/datadog-operator/pull/2556 using your fork commit so your contribution is credited in any case.

PR #2558, opened 2026-02-02
Migrate the operator nightly build deployment to conductor and dynamic workflows

FlorentClarret:
### What does this PR do?

Migrate the operator nightly build deployment to conductor and dynamic workflows. This PR works in pair with https://github.com/DataDog/k8s-datadog-agent-ops/pull/7307 and must be merged the same day

- Remove the `service.datadog.yaml` from this repo and migrate it to the KDAO repository
- Remove variables from the KDAO downstream pipeline

### Motivation

- We can't use dynamic workflow targets if the config file is not in the same repository. We don't want our internal deployment config file in a public repository se we can move it there. This is what we have for the Agent
- Some variables that are no longer needed in the downstream pipeline for KDAO. The pipeline will only take care of building the bundle

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,4 +1,4 @@
-image: registry.ddbuild.io/images/mirror/library/golang:1.25.5
+image: registry.ddbuild.io/images/mirror/library/golang:1.25.6
 variables:
   PROJECTNAME: "datadog-operator"
   PROJECTNAME_CHECK: "datadog-operator-check"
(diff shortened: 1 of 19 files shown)


.gitlab-ci.yml
levan-m: I'm not sure how conductor works but the service definition which was deleted, defines this target and new service def in the other PR uses different one. Should this match `operator-nightly-build` in https://github.com/DataDog/k8s-datadog-agent-ops/pull/7307?

FlorentClarret: Yes, that's a good catch

PR #2572, opened 2026-02-04
enhancement: add feature-based support for Agent Data Plane

tobz:
### What does this PR do?

This PR adds first-class CRD support for the Agent Data Plane (ADP) feature in the Datadog Operator, transitioning it from an experimental annotation-based feature to a proper GA configuration option.

We've added a new `dataplane` feature, and updated the `DatadogAgent` CRD to expose a new `dataPlane` feature section for both enabling and configuring ADP:
```yaml
spec:
 features:
   dataPlane:
     enabled: true          # Enable ADP sidecar container
     dogstatsd:
       enabled: true        # Delegate DogStatsD to ADP
```

We still maintain support for the legacy annotation (`agent.datadoghq.com/adp-enabled`) with CRD taking precedence but this will be removed in the near future.

### Motivation

Agent Data Plane (ADP) is a high-performance sidecar container that handles data ingestion (e.g., DogStatsD metrics) alongside the Core Agent. This enables better resource utilization and performance for high-throughput workloads.

We're adding this as a new feature because while currently ADP only supports handling DogStatsD, it will eventually handle _more_ than just DogStatsD... so we don't want to add it as a setting of the DogStatsD feature, but as its own standalone feature.

### Additional Notes

We've updated the logic to determine when ADP is enabled, both to support the new feature-based configuration as well as allow the (now deprecated) annotation-based mechanism to continue working.

**Precedence Logic:**
1. `spec.features.dataPlane.enabled` is explicitly set: use CRD value
2. if annotation `agent.datadoghq.com/adp-enabled` is set: use annotation value (deprecated)
3. otherwise, disabled (default)

### Minimum Agent Versions

* Agent: v7.75.0-rc.5 or later (ADP support)
* Cluster Agent: N/A (ADP runs on node agent only)

### Describe your test plan

1. **Unit Tests**: New tests in `feature/dataplane/feature_test.go` covering:
   - ADP disabled (default)
   - ADP disabled via explicit CRD setting overriding annotation
   - ADP enabled via CRD
   - ADP with DogStatsD delegation enabled
   - Container name and environment variable verification

2. **Factory Tests**: Updated `feature/test/factory_test.go` to verify ADP container is included when enabled

3. **End-to-end Tests**: Added four new E2E tests exercising DogStatsD: ADP enabled/disabled, for both UDP and UDS. Ensures that the right environment variables are set, that the ADP container does/doesn't exist, that host port mappings move to the right container, etc.

4. **Manual Testing**:
   ```bash
   # Deploy with ADP enabled
   kubectl apply -f - <<EOF
   kind: DatadogAgent
   metadata:
     name: datadog
   spec:
     global:
       credentials:
         apiSecret:
           secretName: datadog-secret
           keyName: api-key
     features:
       dataPlane:
         enabled: true
         dogstatsd:
           enabled: true
   EOF

   # Verify ADP container is running
   kubectl get pods -l app.kubernetes.io/name=datadog-agent-deployment -o jsonpath='{.items[0].spec.containers[*].name}'
   # Should include: agent-data-plane

   # Verify environment variables
   kubectl get ds datadog-agent -o jsonpath='{.spec.template.spec.containers[?(@.name=="agent")].env}' | jq '.[] | select(.name | startswith("DD_DATA_PLANE"))'
   ```

### Checklist


git-thuerk-done (approved):
Good for docs

levan-m (approved):

PR #2583, opened 2026-02-10
[autoscaling] DPA Improvements: add vertical control + minor fixes

vboulineau:
### What does this PR do?

Add fields to CRD to control vertical resources (cpu, memory) and vertical values (requests, limits) per container.
Remove requirement of `MaxReplicas` as it's not needed in vertical-only mode.
Add a condition to highlight vertical scaling limited.

### Motivation

Improve UX and tuning capabilities.

### Minimum Agent Versions

Cluster Agent 7.77

### Describe your test plan

No QA

### Checklist

### DIFF
--- a/api/datadoghq/common/datadogpodautoscaler_types.go
+++ b/api/datadoghq/common/datadogpodautoscaler_types.go
@@ -152,7 +152,7 @@ type DatadogPodAutoscalerObjective struct {
 // +kubebuilder:object:generate=true
 type DatadogPodAutoscalerPodResourceObjective struct {
 	// Name is the name of the resource.
-	// +kubebuilder:validation:Enum:=cpu
+	// +kubebuilder:validation:Enum:=cpu;memory
 	Name corev1.ResourceName `json:"name"`
 
 	// Value is the value of the objective.
@@ -163,7 +163,7 @@ type DatadogPodAutoscalerPodResourceObjective struct {
 // +kubebuilder:object:generate=true
 type DatadogPodAutoscalerContainerResourceObjective struct {
 	// Name is the name of the resource.
-	// +kubebuilder:validation:Enum:=cpu
+	// +kubebuilder:validation:Enum:=cpu;memory
 	Name corev1.ResourceName `json:"name"`
 
 	// Value is the value of the objective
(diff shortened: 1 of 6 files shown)


api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -349,7 +361,27 @@ type DatadogPodAutoscalerContainerConstraints struct {
 	Enabled *bool `json:"enabled,omitempty"`
 
 	// Requests defines the constraints for the requests of the container.
+	// WARNING: Deprecated
+	// +deprecated: use MinAllowed and MaxAllowed instead
```
vboulineau, line 365: We used to have `Requests` and `Limits`, removed `Limits` as not used in `v1alpha2`.
Upstream VPA is using a common for both (or basically whatever is allowed), I think it makes sense, though 2 things to note:
- It's easier and avoid duplication
- It does not allow to restrict only requests anymore (except if only requests are set in ControlledValues)

HemeryJu: If I understood correctly the struct, it woud be impossible to set up different values for requests and limits (and still keeping both in controlled values). Could it be an issue?

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -332,12 +332,24 @@ type DatadogPodAutoscalerConstraints struct {
 	MinReplicas *int32 `json:"minReplicas,omitempty"`
 
 	// MaxReplicas is the upper limit for the number of POD replicas. Needs to be >= minReplicas.
-	MaxReplicas int32 `json:"maxReplicas"`
+	// +kubebuilder:validation:Minimum=1
+	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
```
HemeryJu, line 336: 🔥

config/crd/bases/v1/datadoghq.com_datadogpodautoscalers.yaml
```diff
@@ -88,14 +88,51 @@ spec:
                           DatadogPodAutoscalerContainerConstraints defines constraints that should always be respected for a container.
                           If no constraints are set, it enables resource scaling for all containers without any constraints.
                         properties:
+                          controlledResources:
+                            description: |-
+                              Specifies the resources for which recommendations will be computed.
+                              If not specified, it defaults to CPU and Memory.
+                              If an empty list is provided, no resource will be controlled (equivalent to Enabled=false).
+                            items:
+                              description: ResourceName is the name identifying various resources in a ResourceList.
```
NoelM, line 97: Should we consider adding examples, e.g., "cpu", "memory"?

NoelM (approved):

config/crd/bases/v1/datadoghq.com_datadogpodautoscalers.yaml
vboulineau: It's coming from upstream type, so not possible in the autogenerated

api/datadoghq/common/datadogpodautoscaler_types.go
vboulineau: I don't believe it would the case, I think bounding request and limits gives maximum flexibility but probably not required. If we see demand, we could revisit in next version of the CRD

ADKosm (approved):

PR #2605, opened 2026-02-13
Allow using FIPS variants of ddot-collector and agent -full images

agagniere:
Preserve useFIPSAgent setting when overriding the agent version tag

### What does this PR do?

`useFIPSAgent` setting now applies to agent `-full` and `ddot-collector` images.

### Motivation

We will start publishing FIPS variants of those images soon: https://github.com/DataDog/datadog-agent/pull/46053

[OTAGENT-845](https://datadoghq.atlassian.net/browse/OTAGENT-845?atlOrigin=eyJpIjoiY2E4YjMxZmZlYjg3NDk4NTg5NWY0ZjE1NDcxZjJhYmYiLCJwIjoiaiJ9)

### Minimum Agent Versions

Agent: v7.78.0

can be merged before as the check makes sure the combination is valid

### Describe your test plan

I added a few test cases

### Checklist

- [x] All commits are signed[^1]

[^1]: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

[OTAGENT-845]: https://datadoghq.atlassian.net/browse/OTAGENT-845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/datadoghq/v1alpha1/datadogagentprofile_validation_test.go
+++ b/api/datadoghq/v1alpha1/datadogagentprofile_validation_test.go
@@ -8,13 +8,14 @@ package v1alpha1
 import (
 	"testing"
 
+	"k8s.io/utils/ptr"
+
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 
 	"github.com/DataDog/datadog-operator/api/datadoghq/common"
 	"github.com/DataDog/datadog-operator/api/datadoghq/v2alpha1"
-	"github.com/DataDog/datadog-operator/api/utils"
 )
 
 func TestIsValidDatadogAgentProfile(t *testing.T) {
@@ -128,7 +129,7 @@ func TestIsValidDatadogAgentProfile(t *testing.T) {
 		Config: &v2alpha1.DatadogAgentSpec{
 			Features: &v2alpha1.DatadogFeatures{
 				GPU: &v2alpha1.GPUFeatureConfig{
-					Enabled: utils.NewBoolPointer(true),
+					Enabled: ptr.To(true),
 				},
 			},
 		},
@@ -138,8 +139,8 @@ func TestIsValidDatadogAgentProfile(t *testing.T) {
 		Config: &v2alpha1.DatadogAgentSpec{
 			Features: &v2alpha1.DatadogFeatures{
 				GPU: &v2alpha1.GPUFeatureConfig{
-					Enabled:        utils.NewBoolPointer(true),
-					PrivilegedMode: utils.NewBoolPointer(true),
+					Enabled:        ptr.To(true),
+					PrivilegedMode: ptr.To(true),
 				},
 			},
 		},
@@ -149,7 +150,7 @@ func TestIsValidDatadogAgentProfile(t *testing.T) {
 		Config: &v2alpha1.DatadogAgentSpec{
 			Features: &v2alpha1.DatadogFeatures{
 				NPM: &v2alpha1.NPMFeatureConfig{
-					Enabled: utils.NewBoolPointer(true),
+					Enabled: ptr.To(true),
 				},
 			},
 		},
@@ -159,7 +160,7 @@ func TestIsValidDatadogAgentProfile(t *testing.T) {
 		Config: &v2alpha1.DatadogAgentSpec{
 			Features: &v2alpha1.DatadogFeatures{
 				GPU: &v2alpha1.GPUFeatureConfig{
-					Enabled: utils.NewBoolPointer(true),
+					Enabled: ptr.To(true),
 				},
 			},
 		},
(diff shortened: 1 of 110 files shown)


pkg/images/images.go
```diff
@@ -173,8 +173,14 @@ func OverrideAgentImage(currentImage string, overrideImageSpec *v2alpha1.AgentIm
...
 	if overrideImage.tag != "" {
-		image.WithFIPS(overrideImage.isFIPS).
-			WithFull(overrideImage.isFull)
+		// Only override FIPS/Full if they're explicitly present in the override tag
+		// Otherwise preserve current settings (from global.useFIPSAgent or existing image)
+		if overrideImage.isFIPS {
+			image.WithFIPS(overrideImage.isFIPS)
+		}
+		if overrideImage.isFull {
+			image.WithFull(overrideImage.isFull)
```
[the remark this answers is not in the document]
agagniere: I don't think this is a valid use case. If a user sets a fully custom registry + name + tag in the `name` field, they aren't going to set `useFIPSAgent` to true at the same time if they don't expect the `-fips` variant to be used

pkg/images/images_test.go
```diff
@@ -328,7 +328,7 @@ func Test_ToString(t *testing.T) {
 				isFIPS:   true,
 				isFull:   true,
 			},
-			want: "gcr.io/datadoghq/agent:7.64.0-fips-jmx",
+			want: "gcr.io/datadoghq/agent:7.64.0-fips-full",
```
levan-m, line 331: Does `fips-full` include jmx?

pkg/images/images_test.go
```diff
@@ -552,7 +563,7 @@ func Test_OverrideAgentImage(t *testing.T) {
 			overrideImageSpec: &v2alpha1.AgentImageConfig{
 				Tag: "7.65.0",
 			},
-			want: "gcr.io/datadoghq/agent:7.65.0",
+			want: "gcr.io/datadoghq/agent:7.65.0-fips",
```
levan-m: I see you referenced helm in the description however `overrides` work differently in Operator and shouldn't change. Override has highest precedence.
> I changed the behavior to stop discarding the useFIPSAgent setting when overriding the version, in line with the helm chart.

pkg/images/images.go
```diff
@@ -121,6 +126,24 @@ func (i *Image) WithFull(isFull bool) *Image {
...
+// for the configured version. Specifically:
+//   - agent -fips-full (isFIPS && isFull) is only published from v7.78.0
+//   - ddot-collector -fips (isFIPS && name == DefaultDdotCollectorImageName) is only published from v7.78.0
+//
+// Returns nil if no incompatibility is detected.
+func (i *Image) FIPSVersionError() error {
+	if !i.isFIPS {
+		return nil
+	}
+	if i.isFull || i.name == DefaultDdotCollectorImageName {
```
levan-m: This is called from applyGlobalSettings which is based on internal default and global registry, so it can't have a `full` suffix or be any version other than one hardcoded in this file. So it will either always error without user being able to do anything about it or be a dead code once default is 7.78+. We can drop this and wait for 7.78 bump and then merge this change.

levan-m (changes requested):

pkg/images/images_test.go
agagniere: Yes, see https://github.com/DataDog/datadog-agent/blob/main/.gitlab/deploy/container_build/docker_linux.yml#L398
```
--build-arg WITH_JMX=true --build-arg WITH_JMX_FIPS=true
```

pkg/images/images.go
agagniere: Good catch, I had indeed placed my check at the wrong place, I now placed it after the override, which was my intent.

pkg/images/images.go
```diff
@@ -173,8 +196,14 @@ func OverrideAgentImage(currentImage string, overrideImageSpec *v2alpha1.AgentIm
 	// If an override tag (whether from the tag or name) is provided, then the FIPS and full suffixes present in this tag
 	// should take precedence over the current image suffixes. If it's not provided, then preserve the original
 	if overrideImage.tag != "" {
-		image.WithFIPS(overrideImage.isFIPS).
-			WithFull(overrideImage.isFull)
+		// Only override FIPS/Full if they're explicitly present in the override tag
+		// Otherwise preserve current settings (from global.useFIPSAgent or existing image)
+		if overrideImage.isFIPS {
```
mackjmr: I think user overrides should take precedence over useFips, so this can be:
```
image.WithFIPS(overrideImage.isFIPS).
image.WithFull(overrideImage.isFull)
```
IMO client override should take precedence over other settings, if they hardcode something in that use it. Middle ground would be to log an error/ error out when they override to a non fips image tag + add useFips, as it may be risky to run a non fips image if useFips was set

mackjmr: I'll defere to @levan-m, what do we do in core agent when useFips is true but user overrides image tag to a non fips image ?

agagniere: Reverted to previous behavior

PR #2610, opened 2026-02-18
discovery: add system-probe-lite support

Yumasi:
### What does this PR do?

Adds support for starting [system-probe-lite](https://datadoghq.atlassian.net/wiki/spaces/DSCVR/pages/6332384226/Starting+system-probe-lite+and+system-probe) (SPL) instead of the full system-probe in the operator when only discovery is enabled.

SPL is a privileged Rust binary that implements just the discovery module. Kubernetes needs a different approach than Docker/Linux since the operator is updated asynchronously to agent versions — we can't enable discovery by default and guarantee we won't start the full system-probe with default config.

#### How it works:

- When only discovery is explicitly enabled (no other system-probe features), the system-probe container runs SPL as the entry point
- If `discovery.enabled` is explicitly enabled and SPL doesn't exist, we fall back to the full system-probe binary (since the user opted in knowingly)
- If `discovery.enabled` is unset (nil), this means the user did not make an explicit choice and the operator chooses whether or not to enable the feature automatically, we fall back to `sleep infinity` to avoid starting system-probe unexpectedly when using an older agent image that doesn't have SPL. Note that the operator does not activate discovery by default in this PR; that is a separate, future step.
- If other system-probe features are also enabled (NPM, USM, etc.), the regular system-probe binary is used directly.

### Motivation

Match agent configuration.

### Minimum Agent Versions

* Agent: v7.78

*However*, like in the Helm chart change, it checks first for the `sd-agent` binary to exists in the container image. This means using an older agent image will simply use `system-probe`'s discovery implementation instead of sd-agent.

### Describe your test plan

- Tested on an AWS cluster that it behaves as follow:
	- Default config i.e `features.serviceDiscovery.enabled` set to false: no system-probe container.
	- `features.serviceDiscovery.enabled` set to true: `system-probe` container starts; execs `system-probe-lite`.
	- The above + another system-probe feature (e.g `features.npm.enabled`): `system-probe` container starts; `system-probe` runs directly.
	- Discovery enabled + agent version without sd-agent: `system-probe` container starts; `system-probe` runs directly.

### Checklist

- [X] All commits are signed (see: [signing commits][1])


docs/configuration.v2alpha1.md
```diff
@@ -200,6 +200,7 @@ spec:
 | features.sbom.host.analyzers | To use for SBOM collection. |
 | features.sbom.host.enabled | Enable this option to activate SBOM collection. Default: false |
 | features.serviceDiscovery.enabled | Enables the service discovery check. Default: false |
+| features.serviceDiscovery.enabledByDefault | EnabledByDefault is set by the operator when it enables this feature via default configuration, as opposed to an explicit user choice. When true, the system-probe binary is not used as a fallback if system-probe-lite is unavailable — the container falls back to sleep infinity instead, to avoid unexpected resource usage on older agent images. This field is managed by the operator and should not be set by users. |
```
domalessi: Does this capture the intended meaning? If yes, let's update the line in configuration_public.md as well to match!

```suggestion
| features.serviceDiscovery.enabledByDefault |  Indicates that the operator enabled this feature automatically rather than in response to an explicit user setting. When true, if `system-probe-lite` is unavailable, the container falls back to `sleep infinity` rather than `system-probe`, to avoid unexpected resource usage on older agent images. This field is managed by the operator and must not be set by users. |
```

domalessi (commented):
Left a suggestion to improve clarity! Approving so you're not blocked on me.

docs/configuration.v2alpha1.md
Yumasi: Yes, thank you, I applied it 14334d1

domalessi (approved):
Thanks for the changes!

internal/controller/datadogagent/feature/servicediscovery/feature.go
```diff
@@ -40,21 +44,66 @@ func (f *serviceDiscoveryFeature) ID() feature.IDType {
...
+		Containers: []apicommon.AgentContainerName{apicommon.CoreAgentContainerName, apicommon.SystemProbeContainerName},
+	}
+
+	f.networkStatsEnabled = true
+	if sd.NetworkStats != nil {
+		f.networkStatsEnabled = apiutils.BoolValue(sd.NetworkStats.Enabled)
 	}
 
+	f.useSystemProbeLite = !hasOtherSystemProbeFeatures(ddaSpec.Features)
+	f.userExplicitlyEnabled = apiutils.BoolValue(sd.Enabled)
```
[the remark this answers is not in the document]
Yumasi: Fixed in 8b7ca14 by storing a pointer to the live feature struct so we can get the values after remote config was applied.

internal/controller/datadogagent/feature/servicediscovery/feature.go
```diff
@@ -38,17 +45,61 @@ func (f *serviceDiscoveryFeature) ID() feature.IDType {
...
+	fallback := "sleep infinity"
+	if userOptedIn {
+		fallback = "system-probe --config=/etc/datadog-agent/system-probe.yaml"
+	}
+	return fmt.Sprintf("system-probe-lite run --socket %s --log-level ${DD_LOG_LEVEL:-info} || %s", socketPath, fallback)
+}
+
+// hasOtherSystemProbeFeatures returns true if any feature besides service discovery
+// requires the full system-probe binary. When true, system-probe-lite cannot be used.
+func hasOtherSystemProbeFeatures(features *v2alpha1.DatadogFeatures) bool {
```
vitkyrka, line 87: How can we ensure that list is correct and complete and remains that way? It's not obvious that I have to patch `feature/servicediscovery/feature.go` if I'm adding some unrelated feature that uses system-probe, and that a failure to do so may cause my feature to not work in some cases.

api/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -647,6 +647,13 @@ type ServiceDiscoveryFeatureConfig struct {
 	// +optional
 	Enabled *bool `json:"enabled,omitempty"`
 
+	// Indicates that the operator enabled this feature automatically rather than in response to an
+	// explicit user setting. When true, if `system-probe-lite` is unavailable, the container falls
+	// back to `sleep infinity` rather than `system-probe`, to avoid unexpected resource usage on
+	// older agent images. This field is managed by the operator and must not be set by users.
+	// +optional
+	EnabledByDefault *bool `json:"enabledByDefault,omitempty"`
```
tbavelier: We do not want to expose such a flag in the CRD: this field is managed by the operator and must not be set by the users is not compatible with the CRD approach. Instead, we can remove the default false and work around that to understand about the tri-state nil/false/true: https://github.com/DataDog/datadog-operator/pull/2880 (Codex generated so might be fully wrong, just to highlight how it can possibly work in theory for your use case). If the approach doesn't work, we still don't want to use a CRD flag for that, and instead use an annotation for experimental feature like sd-lite seems to be (see https://github.com/DataDog/datadog-operator/blob/fd197585b379df27e8fe362407eddeb46b48be47/internal/controller/datadogagent/feature/privateactionrunner/feature.go#L66 for an example on how to modify how a feature is configured based on an annotation)

internal/controller/datadogagent/feature/servicediscovery/feature.go
```diff
@@ -38,17 +45,61 @@ func (f *serviceDiscoveryFeature) ID() feature.IDType {
...
+	sd := ddaSpec.Features.ServiceDiscovery
+
+	// Explicit Enabled=false always disables the feature, even if EnabledByDefault=true.
+	if sd.Enabled != nil && !*sd.Enabled {
+		return reqComp
+	}
+	// Feature requires either an explicit opt-in or a default enablement.
+	if !apiutils.BoolValue(sd.Enabled) && !apiutils.BoolValue(sd.EnabledByDefault) {
+		return reqComp
 	}
```
tbavelier: See above on how to possibly change this behaviour without modifying the CRD (removing existing default / use annotation)

internal/controller/datadogagent/feature/servicediscovery/feature.go
```diff
@@ -104,6 +155,20 @@ func (f *serviceDiscoveryFeature) ManageNodeAgent(managers feature.PodTemplateMa
...
 
+	// Direct PodTemplateSpec mutation: no managers API for command overrides.
+	// Re-evaluate here (not cached from Configure) so that RC state merged by other
+	// features' Configure calls (e.g. USM) is taken into account.
+	if !hasOtherSystemProbeFeatures(f.features) {
+		for i := range managers.PodTemplateSpec().Spec.Containers {
+			c := &managers.PodTemplateSpec().Spec.Containers[i]
+			if c.Name == string(apicommon.SystemProbeContainerName) {
+				c.Command = []string{"/bin/sh", "-c"}
+				c.Args = []string{systemProbeLiteCommand(common.DefaultSystemProbeSocketPath, f.userExplicitlyEnabled)}
```
tbavelier, line 164: Are we 100% sure we want to change both cmd and args ? Cmd changes the entrypoint which might have un-intended consequences, see https://github.com/DataDog/datadog-operator/pull/2864 for a more detailed explanation

tbavelier (changes requested):
Need to change the CRD logic: such logic should not be in the CRD as a field we don't want users to expose, see comments
Not sure about the command/args if it warrants changes
Will discuss with team for the other SP features if they have an idea
Conflict with main due to pointer refactor I believe

internal/controller/datadogagent/feature/servicediscovery/feature.go
Yumasi: The reason we change both is because we need the shell logic behind it. One problem we have with the Operator and Helm chart, is that we can't know which agent version we are running, which means if you modify only the args:

- you might try to rely on system-probe's fallback to SPL mechanism. But if you are on a system-probe version that doesn't have the support, it will crash because of an unknown CLI option.
- on supported versions, you spawned the full system-probe to have it check a config we already have in the operator, which means a huge spike in resource consumption which can be enough to trigger pods memory limits and defeats the purpose of having SPL in the first place.
- If only discovery is explicitly enabled, on non-supported versions, you might end up with system-probe starting up, and figuring out it doesn't need to run because it does not have discovery at all, and exits. In this case, you will have a crashloopbackoff.

By using a shell here, we can try to do best effort, run SPL directly when possible, or the full system-probe, and fallback to a sleep infinity on non-supported version, that will prevent the container from going into a CLB.

Of course, we are open to suggestions on improving this.

PR #2611, opened 2026-02-18
[ACTP] Install the private action runner in the cluster agent via the operator

merchristK:
## What does this PR do?

This PR adds support for running the Private Action Runner (PAR) in the Cluster Agent via the Datadog Operator. Previously,

Key changes:
- Adds cluster agent PAR configuration via annotations
- Implements RBAC management for Kubernetes secret access

## Motivation
Currently, PAR is only installable in the Node Agent via the operator. Supports for DCA has been added in 7.77

## Additional Notes
Implementation Details:
Annotation-based Configuration (_uses the same pattern as node agent PAR_)
```yaml
annotations:
cluster-agent.datadoghq.com/private-action-runner-enabled: "true"
cluster-agent.datadoghq.com/private-action-runner-configdata: |
    private_action_runner:
    enabled: true
    self_enroll: true
    identity_use_k8s_secret: true
    identity_secret_name: par-identity
    actions_allowlist:
        - com.datadoghq.http.request
        - com.datadoghq.kubernetes.core.listPod
```

## Minimum Agent Versions

PAR support in Cluster Agent requires:

- Agent: N/A (node agent PAR unchanged)
- Cluster Agent: v7.77.X (version with PAR support in DCA)

### Unit Tests
```bash
go test ./internal/controller/datadogagent/feature/privateactionrunner/...
```

### Local Testing:
A comprehensive test script ([local-test](https://gist.github.com/merchristK/0e561d086148af41c0b3258774284db7)) was used for Docker Desktop Kubernetes testing:

```bash
# Set credentials
export DD_API_KEY=your-api-key
export DD_APP_KEY=your-app-key
export CLUSTER_AGENT_IMAGE=r<INTERNAL_REPO>/datadog-cluster-agent:dev-par-tag

# Deploy operator and test DatadogAgent
./local-test deploy

# Validate deployment
./local-test validate

# Teardown
./local-test teardown
```

### Example Configuration:
See `examples/datadogagent/datadog-agent-with-private-action-runner.yaml` for both node agent and cluster agent PAR
configurations.

## Checklist

- [x] PR has at least one valid label: enhancement
- [x] PR has a milestone or the qa/skip-qa label
- [x] All commits are signed (see:
https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)

### DIFF
--- a/internal/controller/datadogagent/feature/privateactionrunner/config.go
+++ b/internal/controller/datadogagent/feature/privateactionrunner/config.go
@@ -10,9 +10,8 @@ import (
 	"strconv"
 	"strings"
 
-	corev1 "k8s.io/api/core/v1"
-
 	"gopkg.in/yaml.v3"
+	corev1 "k8s.io/api/core/v1"
 )
 
 // Environment variable names for Private Action Runner configuration


merchristK:
* **#2611** <a href="https://app.graphite.com/github/pr/DataDog/datadog-operator/2611?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/datadog-operator/2611?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

examples/datadogagent/datadog-agent-with-private-action-runner.yaml
```diff
@@ -0,0 +1,117 @@
+# Private Action Runner (PAR) enables workflow automation capabilities in the Datadog Agent.
+# This example demonstrates how to configure PAR using annotations.
+#
+# For more information about Private Action Runner:
+# https://docs.datadoghq.com/workflows/private-actions
```
tbavelier: Is this link correct/not public yet ? It leads to a 404 at the moment

internal/controller/datadogagent/feature/utils/utils.go
```diff
@@ -84,3 +87,14 @@ func HasPrivateActionRunnerConfigAnnotation(dda metav1.Object, annotationName st
 	value, ok := dda.GetAnnotations()[annotationName]
 	return value, ok
 }
+
+// HasClusterAgentPrivateActionRunnerAnnotation returns true if the Cluster Agent Private Action Runner is enabled via the dedicated annotation
+func HasClusterAgentPrivateActionRunnerAnnotation(dda metav1.Object) bool {
+	return hasFeatureEnableAnnotation(dda, EnableClusterAgentPrivateActionRunnerAnnotation)
+}
```
tbavelier: Personally, I don't really like this proliferation of these functions. Could we consider changing `hasFeatureEnableAnnotation` to be public `HasFeatureEnableAnnotation` and directly call in each feature with the annotation name ?

internal/controller/datadogagent/feature/utils/utils.go
```diff
@@ -84,3 +87,14 @@ func HasPrivateActionRunnerConfigAnnotation(dda metav1.Object, annotationName st
...
+
+// HasClusterAgentPrivateActionRunnerConfigAnnotation returns the value and presence of a Cluster Agent Private Action Runner config annotation
+func HasClusterAgentPrivateActionRunnerConfigAnnotation(dda metav1.Object, annotationName string) (string, bool) {
+	value, ok := dda.GetAnnotations()[annotationName]
+	return value, ok
+}
```
tbavelier: This is exactly the same as `HasHostProfilerConfigAnnotion` (typo in the function name). Can we consolidate under a `GetFeatureConfigAnnotation(dda metav1.Object, annotationName string)` and re-use it ?

internal/controller/datadogagent/feature/privateactionrunner/feature.go
```diff
@@ -124,7 +161,94 @@ func (f *privateActionRunnerFeature) getConfigMapName() string {
...
+
+	if !config.IdentityUseK8sSecret {
+		managers.EnvVar().AddEnvVarToContainer(
+			apicommon.ClusterAgentContainerName,
+			&corev1.EnvVar{
+				Name:  "DD_PRIVATE_ACTION_RUNNER_IDENTITY_USE_K8S_SECRET",
+				Value: "false",
+			},
+		)
+	}
```
tbavelier:
```suggestion
	identityUseK8sSecret := config.IdentityUseK8sSecret == nil || *config.IdentityUseK8sSecret
	managers.EnvVar().AddEnvVarToContainer(
		apicommon.ClusterAgentContainerName,
		&corev1.EnvVar{
			Name:  "DD_PRIVATE_ACTION_RUNNER_IDENTITY_USE_K8S_SECRET",
			Value: strconv.FormatBool(identityUseK8sSecret),
		},
	)
```
replacing also
```go
	managers.EnvVar().AddEnvVarToContainer(
		apicommon.ClusterAgentContainerName,
		&corev1.EnvVar{
			Name:  "DD_PRIVATE_ACTION_RUNNER_IDENTITY_USE_K8S_SECRET",
			Value: "true",
		},
	)
```
Explanation: instead of setting a default `true` env var, then overriding it later to false if user requests it, could we use a single call ? We do this in other places, such as dogstatsd feature. (this assumes config.IdentityUseK8sSecret is changed to `*bool`

internal/controller/datadogagent/feature/privateactionrunner/feature.go
```diff
@@ -177,10 +301,38 @@ func (f *privateActionRunnerFeature) ManageOtelAgentGateway(managers feature.Pod
...
+	checksum, err := comparison.GenerateMD5ForSpec(configData)
 	if err != nil {
 		return "", "", fmt.Errorf("failed to generate MD5 for Private Action Runner config: %w", err)
 	}
 	return object.GetChecksumAnnotationKey(feature.PrivateActionRunnerIDType), checksum, nil
 }
+
+func getPrivateActionRunnerRbacResourcesName(owner metav1.Object) string {
+	return owner.GetName() + "-private-action-runner"
+}
```
tbavelier: While not a const, for consistency with features such as helmcheck, orchestratorexplorer, ksm, could you move this helper to a `const.go` ?

internal/controller/datadogagent/feature/privateactionrunner/feature.go
```diff
@@ -124,7 +161,94 @@ func (f *privateActionRunnerFeature) getConfigMapName() string {
...
+
+	if config.PrivateKey != "" {
+		managers.EnvVar().AddEnvVarToContainer(
+			apicommon.ClusterAgentContainerName,
+			&corev1.EnvVar{
+				Name:  "DD_PRIVATE_ACTION_RUNNER_PRIVATE_KEY",
+				Value: config.PrivateKey,
+			},
+		)
+	}
```
tbavelier: Considering the variable name (private key), I doesn't seem "good practice" to store it in plain text as an env var on the container. Could we possibly use `valueFrom.secretKeyRef` to set it from a secret ? I'm missing context as it might add friction (storing the key in a secret)/this value might not be sensitive

internal/controller/datadogagent/feature/privateactionrunner/feature.go
```diff
@@ -88,31 +113,43 @@ const (
...
+	if f.clusterEnabled {
+		rbacResourcesName := getPrivateActionRunnerRbacResourcesName(f.owner)
+
+		// Add Role (namespaced) for secret access - parse config to get identity_secret_name
+		return managers.RBACManager().AddPolicyRules(
+			f.owner.GetNamespace(),
+			rbacResourcesName,
+			f.clusterServiceAccountName,
+			getClusterAgentRBACPolicyRules(f.clusterConfigData),
+		)
```
tbavelier, line 155: Just wondering: is the RBAC always needed when DCA PAR is enabled ? Or are there cases where it's not needed (e.g. depending on the enroll config or something) ? Because if yes, let's not create un-necessary roles in these cases. (Should such a logic be needed, it would be in Configure() and store the result in the feature struct, see global comment)

examples/datadogagent/datadog-agent-with-private-action-runner.yaml
merchristK: Good catch. It is supposed to be https://docs.datadoghq.com/actions/private_actions/

internal/controller/datadogagent/feature/utils/utils.go
merchristK: yes sure. I was following the current pattern blindly 👍

internal/controller/datadogagent/feature/privateactionrunner/feature.go
merchristK: Hum the thinking here is I want to default to `True` and only disable when user explicitly requires it. https://github.com/DataDog/datadog-agent/blob/048f85df5da9195e0671046a0c4949fc1f2592ac/pkg/config/setup/privateactionrunner.go#L46

Not sure if there is a better way to achieves it

internal/controller/datadogagent/feature/privateactionrunner/feature.go
merchristK: The RBAC is needed when `self_enroll` is true. In such scenarios, we want the leader of the DCA to create the identity (private_key, urn) and store it in a k8s secret so that followers can retrieve this later. 

Will address this 👍

internal/controller/datadogagent/feature/privateactionrunner/feature.go
tbavelier: The code I shared achieves this same behaviour: if not provided `config.IdentityUseK8sSecret == true` is true, so the variable is true. If provided to true, it stays true, if provided to false, it's false. And that will be done in the Configure part.
1. In Configure -> f.identityUseK8sSecret := config.IdentityUseK8sSecret == nil || *config.IdentityUseK8sSecret
2. In ManageDCA, strconv.FormatBool(f.identityUseK8sSecret)

merchristK:
Thanks for the review @tbavelier 

> Could you also please add to the PR description QA instructions ? These will most likely be done by someone from your team, but the idea is to test "e2e" it does work as expected, not only through unit tests. We're not asking for all possible scenarios, but a basic scenario we expect customers/beta testers to do

My PR description already includes a test plan and I've linked this https://gist.github.com/merchristK/0e561d086148af41c0b3258774284db7 for easier testing... But global idea is to just build the image locally, and deploy the operator in a local docker k8s cluster.

tbavelier:
My bad, didn't catch the script to test it, I glanced at it, but a lot of it is helpers to deploy, so the QA instructions would be `make IMG=custom-image docker-build-ci` then `make IMG=custom-image deploy` after loading the image in the cluster, then check DCA logs for PAR logs, check DCA env vars, etc.

internal/controller/datadogagent/feature/privateactionrunner/feature.go
merchristK: True... I am just blind and didn't read carefully. Thks 👍

internal/controller/datadogagent/feature/privateactionrunner/feature.go
merchristK: The value is indeed sensitive and I agree it's not a good practice to store it in plain text. Here is my current thinking:

- Most users should use `self_enroll: true` instead of providing `urn, private_key`
- `private_key` is for quick validation (and not recommended for prod per-se)
- The config we add via the annotation is supposed to be the real private runner config and for the node agent, we just create a config map out of it. I want for now to have the same config. I am not using a config map as for the node agent because I realized other components are using envvar mostly.
- This whole implementation is temporary since we will iterate on it to configure the PAR with real "features config" instead of annotations. And it will look like this
```yaml
feature:
  privateActionRunner:
    enabled: true
    selfEnroll: true
    privateKeySecret:
      key: private-key
      name: datadog-private-actions-runner-identity
    urn: ...
```

tbavelier: Sounds good then, thanks!

> config map as for the node agent because I realized other components are using envvar mostly.

yup, that's something we might try to change eventually as it increases the pod size but it's outside the scope of this PR (as long as the env vars are a reasonable size, not a full JSON encoded string or w/e)

merchristK: I actually liked the env setup because envvar takes precedence over all config and for me it was the best way to be extra sure the pods have the right config haha!

The only json string that might go crazy is the `actions_allowlist`. But it should still be reasonable 👍 

So if I understand correctly, you do recommend configmaps over envvars right?

tbavelier: > So if I understand correctly, you do recommend configmaps over envvars right?

For very long env vars/config yea and at one point, we might try to align everything in `datadog.yaml` configmap instead of env vars, but that's a long term project. We indeed use env vars everywhere and they do take priority over files config, so let's keep as is. Was more of a footnote / rumbling.

internal/controller/datadogagent/feature/privateactionrunner/feature.go
```diff
@@ -56,26 +73,48 @@ const defaultConfigData = "private_action_runner:\n    enabled: true\n"
...
+		f.clusterConfig = &PrivateActionRunnerConfig{Enabled: true}
+		configData, ok := featureutils.GetFeatureConfigAnnotation(dda, featureutils.ClusterAgentPrivateActionRunnerConfigDataAnnotation)
+		if !ok {
+			configData = defaultConfigData
+		}
+
+		clusterConfig, err := parsePrivateActionRunnerConfig(configData)
+		if err != nil {
+			return reqComp
+		}
```
tbavelier, line 93: We early return if the parsing errors out, but we've already set `f.clusterConfig`, so ManageClusterAgent won't no-op based on the current conditionals and will do some operations despite DCA not being a required component.
Can be fixed with `f.clusterConfig = nil` before the early return ? Moreover, it might be useful to add an error log to surface the error as I assume an error config means the feature won't really work if at all ?

internal/controller/datadogagent/feature/privateactionrunner/feature.go
```diff
@@ -124,7 +174,73 @@ func (f *privateActionRunnerFeature) getConfigMapName() string {
...
+		},
+	)
+
+	managers.EnvVar().AddEnvVarToContainer(
+		apicommon.ClusterAgentContainerName,
+		&corev1.EnvVar{
+			Name:  "DD_PRIVATE_ACTION_RUNNER_PRIVATE_KEY",
+			Value: f.clusterConfig.PrivateKey,
+		},
+	)
```
tbavelier: These 3 env vars are always added even if not needed (e.g. `f.clusterConfig.URN == ""`). Not sure if it's a possible scenario but doesn't hurt to not add them if the extracted config is empty for these specific fields

```go
if f.clusterConfig.URN != "" {
      managers.EnvVar().AddEnvVarToContainer(...)
  }
```

internal/controller/datadogagent/feature/privateactionrunner/feature_test.go
tbavelier, line 1: Not added by this PR, but lots of tests don't provide any values, they should be removed so we only keep relevant tests like the ones you added (they're great coverage, thanks!).

* `Test_privateActionRunnerFeature_ManageSingleContainerNodeAgent`
* `Test_privateActionRunnerFeature_ManageClusterChecksRunner`
* `Test_privateActionRunnerFeature_ManageClusterChecksRunner`
* `Test_privateActionRunnerFeature_ManageOtelAgentGateway`
* `Test_privateActionRunnerFeature_ManageDependencies` -> this one is "critical" in the sense, as it's like we don't manage dependencies while we do
* `Test_privateActionRunnerFeature_ManageClusterAgent` -> same as above, it hides the real behaviour, we do manage cluster agent (with this PR)
* `Test_buildPrivateActionRunnerFeature_WithLogger` -> testing that a Go struct assignment works

internal/controller/datadogagent/feature/privateactionrunner/feature.go
merchristK: yeah I was doing this and it felt like the if is too much... But will add it back 👍

internal/controller/datadogagent/feature/privateactionrunner/feature.go
merchristK: Good catch! I've removed the line where I first set `f.clusterConfig`

I think we should either do this:
```go
// after parsing
f.clusterConfig = clusterConfig
if f.clusterConfig != nil {
	// Due-diligence: if EnableClusterAgentPrivateActionRunnerAnnotation is set to true, it has priority
	// over the config
	f.clusterConfig.Enabled = true
}
```

OR

```go
// after parsing
if f.clusterConfig != nil && !f.clusterConfig.Enabled {
  // FAIL but I dn't think we can fail so do not enable PAR as we have a conflict
}
```

I think option 1 is better

internal/controller/datadogagent/feature/privateactionrunner/feature.go
```diff
@@ -56,26 +73,52 @@ const defaultConfigData = "private_action_runner:\n    enabled: true\n"
...
+
+		clusterConfig, err := parsePrivateActionRunnerConfig(configData)
+		if err != nil {
+			f.logger.Error(err, "failed to parse private action runner config")
+			return reqComp
+		}
+		f.clusterConfig = clusterConfig
+		if !f.clusterConfig.Enabled {
+			// Due-diligence
+			f.logger.Info("private_action_runner.enabled=false in configdata is overridden by the enable annotation")
```
tbavelier:
```suggestion
			f.logger.V(1).Info("private_action_runner.enabled=false in configdata is overridden by the enable annotation")
```
Move this to debug, as this will get spammed every 15 seconds I believe

merchristK:
> Approving to unblock, but please move the log to debug

Done!! Thks for the thorough review

PR #2617, opened 2026-02-19
Update sysprobe src volume mount to have write permissions

gabedos:
### What does this PR do?

Provide system probe container with write access to `/usr/src`

### Motivation

Failing Agent deployments with read only mode enabled on system probe (by default). Tested on both Agetn v7.70 and latest, both throw errors whenever any of these system-probe features are enabled.

```
  - image: gcr.io/datadoghq/agent:7.75.0
    imageID: ""
    lastState: {}
    name: system-probe
    ready: false
    restartCount: 0
    started: false
    state:
      waiting:
        message: 'failed to generate container "f41bed240b0d9daa14f4dca68f1cc53889e3ec627c001aeb4cf5bdeb263fa05b"
          spec: failed to generate spec: failed to mkdir "/usr/src": mkdir /usr/src:
          read-only file system'
```

### Describe your test plan

This PR updates the existing volume mount to no longer be read only. We should expect the above k8s error message to no longer appear on the `system-probe` container because it will have permissions to `mkdir /usr/src`

```
    image: gcr.io/datadoghq/agent:7.75.0
    imagePullPolicy: IfNotPresent
    name: system-probe
    resources: {}
    securityContext:
      readOnlyRootFilesystem: true
    ...
    volumeMounts:
    - mountPath: /host/proc
      name: procdir
      readOnly: true
    - mountPath: /opt/datadog-agent/run
      name: pointerdir
    - mountPath: /tmp
      name: tmp
    - mountPath: /lib/modules
      name: modules
      readOnly: true
    - mountPath: /usr/src
      name: src
    ...
```

### Checklist

### DIFF
--- a/internal/controller/datadogagent/feature/oomkill/feature_test.go
+++ b/internal/controller/datadogagent/feature/oomkill/feature_test.go
@@ -67,7 +67,7 @@ func Test_oomKillFeature_Configure(t *testing.T) {
 			{
 				Name:      common.SrcVolumeName,
 				MountPath: common.SrcVolumePath,
-				ReadOnly:  true,
+				ReadOnly:  false,
 			},
 			{
 				Name:      common.DebugfsVolumeName,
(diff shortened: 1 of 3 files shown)


tbavelier:
Closing, expected behaviour for nodes with `/usr` in read-only FS, e.g. Google COS. Covered by introspection: https://github.com/DataDog/datadog-operator/blob/8df7c4555290cc020d65b3a14b2bead9f2d327ce/internal/controller/datadogagent/feature/servicediscovery/feature.go#L106-L110
Changing the volumeMount to false (so it's not readOnly) won't change the underlying issue (can't have a hostPath directory on a read-only FS when the path doesn't exist)

PR #2635, opened 2026-02-24
[CONTP-1329] Cleanup component properly when name is overridden

tbavelier:
### What does this PR do?

Use the component name override deployment when cleaning up

### Motivation

Found an issue: when disabling a component with an override name, it won't clear the deployment. It does work when the default name is used. https://datadoghq.atlassian.net/browse/CONTP-1329 

### Additional Notes

There's still the edge case where in a single step a customer disables a component and changes its name. A fix would be possibly ugly so leaving as it is for now.

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

1. Deploy a basic DDA
2. Disable the clusterAgent component with an override `spec.override.clusterAgent.disabled=true`
    1. Verify the deployment is deleted
3. Remove the disable override and add a name `spec.override.clusterAgent.name=foo`
    1. A `foo-` deployment should be created
4. Add again the disable override and verify the `foo` deployment is cleaned up

### Checklist


tbavelier:
Done with https://github.com/DataDog/datadog-operator/pull/2651

PR #2654, opened 2026-02-25
Don't create a config map if it's empty

lpmi-13:
### What does this PR do?

Update the config map implementation to not be created if there's no data to be added there.

### Motivation

We were deploying the DD operator in one of our clusters, but we didn't specify any of the values for `.databases` or `.autodiscovery`, and so the template rendered that as `configDataMap: {}`, and then the operator tries to mount a volume referencing a config map that has nothing in it.

Then we got cluster-agent pods stuck in `ContainerCreating`, and we had to manually resolve it.

### Additional Notes

N/A

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist


lpmi-13:
it doesn't look like I can add a label, so I'm not super sure how to fix the failing CI check 😿

gabedos (approved):

gabedos:
Thanks for your contribution @lpmi-13 🙏  This will be included in the next Operator release for `v1.30.0`+

PR #2655, opened 2026-02-25
feat: add support for SSI injectionMode

mtoffl01:
### What does this PR do?

Adds a new `injectionMode` field to the `spec.features.apm.instrumentation` configuration in the DatadogAgent CRD. This field allows users to specify which injection method the Cluster Agent should use for APM library injection.

Valid values are:
- `auto` - Let the Cluster Agent decide the best injection method
- `init_container` - Use init containers for library injection
- `csi` - Use CSI driver for library injection (experimental)
- `image_volume` - Use image volumes for library injection (experimental)

### Motivation

Users need the ability to explicitly configure the injection mode for Single Step Instrumentation (SSI) instead of relying solely on the Cluster Agent's automatic selection. This is particularly useful for environments where a specific injection method is required or preferred.

### Additional Notes

The field is optional and empty by default, allowing the Cluster Agent to apply its own defaults when not specified.

The value is passed to the Cluster Agent via the `DD_APM_INSTRUMENTATION_INJECTION_MODE` environment variable.

### Minimum Agent Versions

* Agent: N/A
* Cluster Agent: v7.76.0 (for `csi` mode), v7.77.0 (for `image_volume` mode)

### Describe your test plan

1. Deploy the operator with the updated CRD
2. Create a DatadogAgent with `spec.features.apm.instrumentation.injectionMode: "csi"`
3. Verify the Cluster Agent deployment has the env var `DD_APM_INSTRUMENTATION_INJECTION_MODE=csi`
4. Verify invalid values are rejected by the CRD validation

### Checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -401,6 +401,9 @@ e2e_autoscaling:
   needs:
     - "go_e2e_deps"
   timeout: 90m
+  id_tokens:
+    DDOCTOSTS_ID_TOKEN:
+      aud: dd-octo-sts
   rules:
     # Disable on Conductor-triggered jobs (ex: nightly) — must be first so it wins
     - if: '$DDR == "true"'


api/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -195,6 +195,13 @@ type SingleStepInstrumentation struct {
 	// +optional
 	Injector *InjectorConfig `json:"injector,omitempty"`
 
+	// InjectionMode is the injection mode to use for libraries injection.
+	// Valid values are: "auto", "init_container", "csi" (experimental, requires Cluster Agent 7.76.0+ and Datadog CSI Driver), "image_volume" (experimental, requires Cluster Agent 7.77.0+).
```
adel121: What is the expected behavior if it is set to "auto" ? 

Should we document this?

internal/controller/datadogagent/feature/apm/feature.go
```diff
@@ -210,6 +212,11 @@ func (f *apmFeature) shouldSetCustomInjectorImage() bool {
 		f.singleStepInstrumentation.injector.imageTag != ""
 }
 
+func (f *apmFeature) shouldSetInjectionMode() bool {
```
tbavelier, line 215: Not convinced this helper has value in this state, we call it inside already inside a block `if f.singleStepInstrumentation != nil`, so we could use directly `f.singleStepInstrumentation.injectionMode != ""` in this block. However, it does have value if we add to it `f.singleStepInstrumentation.enabled` (so we only set the env var if SSI is enabled, no need to bloat the pod with a useless env var if disabled)

api/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -195,6 +195,13 @@ type SingleStepInstrumentation struct {
...
+	// Empty by default so the Cluster Agent can apply its own defaults.
+	// +optional
+	// +kubebuilder:validation:Enum=auto;init_container;csi;image_volume
+	InjectionMode string `json:"injectionMode,omitempty"`
```
tbavelier: Could we use a named string type for consistency with things like `NetworkPolicyFlavor`, `MetricType`, etc. ?
```go
  type InjectionModeType string

  const (
      InjectionModeAuto           InjectionModeType = "auto"
      InjectionModeInitContainer  InjectionModeType = "init_container"
      InjectionModeCSI            InjectionModeType = "csi"
      InjectionModeImageVolume    InjectionModeType = "image_volume"
  )
```

api/datadoghq/v2alpha1/datadogagent_types.go
iamluc: `auto` lets the Cluster Agent decides the best injection mode.

It is now documented in the enum: https://github.com/DataDog/datadog-operator/pull/2655/changes#diff-bb61879086cb5d8e4bd2d033770ca5943d58d914262b1c1f6cab53ad5a1b263aR170-R179

internal/controller/datadogagent/feature/apm/feature.go
iamluc: :+1: Updated to include `f.singleStepInstrumentation.enabled`

api/datadoghq/v2alpha1/datadogagent_types.go
iamluc: Done

PR #2667, opened 2026-02-27
[DatadogAgent] Copy annotations change to DDAI when DDAI enabled

tbavelier:
### What does this PR do?

* Make sure to copy over the DDA annotations to DDAI not only initially, but "always"

### Motivation

* Easier dogfooding of experimental features: such features are configured through annotations on DDA and thus require a delete/recreate of DDA to test / change the lease for a reconciliation to happen. On DDA path, this happens at the next reconcile (15s), but for DDAI, the annotations were not propagated, so even at next reconcile, features would not update

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

2. Deploy a DDA with the annotation `agent.datadoghq.com/fine-grained-kubelet-authorization-enabled: "false"`
3. Verify the node Agent clusterrole (`<dda-name>-agent`) has access to `nodes/proxy`:
    ```yaml
    ╰─❯ k auth can-i get nodes --subresource proxy --as=system:serviceaccount:system:datadog-agent-agent
    Warning: resource 'nodes' is not namespace scoped
    
    yes
    ```
5. Apply/edit the DDA annotation to change it to `true`, wait for next requeue (15s~)
6. Verify the updated annotation is also present in the corresponding `DatadogAgentInternal` object 
7. Verify the node Agent clusterrole doesn't have access anymore to `nodes/proxy`:
    ```yaml
    ╰─❯ k auth can-i get nodes --subresource proxy --as=system:serviceaccount:system:datadog-agent-agent
    Warning: resource 'nodes' is not namespace scoped
    
    no
    ```

### Checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -405,6 +405,9 @@ e2e_autoscaling:
     # Disable on Conductor-triggered jobs (ex: nightly) — must be first so it wins
     - if: '$DDR == "true"'
       when: never
+    # Disable on manual pipelines to test staging images
+    - if: $PUSH_IMAGES_TO_STAGING == 'true'
+      when: never
     # Run automatically on changes to autoscaling code
     - if: $CI_COMMIT_BRANCH
       changes:
--- a/internal/controller/datadogagent/feature/privateactionrunner/config.go
+++ b/internal/controller/datadogagent/feature/privateactionrunner/config.go
@@ -32,6 +32,11 @@ const (
 	DDPARLogFile               = "DD_PRIVATE_ACTION_RUNNER_LOG_FILE"
 )
 
+const (
+	privateActionRunnerConfigKey  = "private_action_runner"
+	privateActionRunnerEnabledKey = "enabled"
+)
+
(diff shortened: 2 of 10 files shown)


levan-m (approved):
Change looks good but wondering whether filtering out hash annotation _we generate_ is overkill (or correct). If we regenerate annotations this implies that spec had significant change shouldn't we reconcile DDAIs as well, what am I missing? Are config map changes reflected in DDA MD5 annotations?

tbavelier:
> Change looks good but wondering whether filtering out hash annotation _we generate_ is overkill (or correct). If we regenerate annotations this implies that spec had significant change shouldn't we reconcile DDAIs as well, what am I missing? Are config map changes reflected in DDA MD5 annotations?

It's indeed overkill, we should simply compare annotations which includes both the deployment annotation (from spec calculation of DDAI) and other annotations (experimental features or something else). If the spec changes, the DDAI controller will apply the update instantly. If only annotations are changing, we still update the object, and it will reconcile after the period.

PR #2676, opened 2026-03-02
Datadog pod autoscaler profile CRD

ADKosm:
### What does this PR do?

Introduction of DatadogPodAutoscalerProfile. Please check [this](https://datadoghq.atlassian.net/wiki/spaces/CAUT/pages/6280052935/GitOps+implementation+notes) and [this](https://datadoghq.atlassian.net/wiki/spaces/CAUT/pages/6248235928/brainstorm+GitOps-Based+Batch+Onboarding+Workflow) document to learn more about this project.

### Motivation

Initial draft of new CRD for Profile entity.

### Additional Notes

I want to explain my decision for moving some parts into `common` package.
From one side, it is beneficial to have exactly `v1alpha2.DatadogPodAutoscalerSpec` type inside `DatadogPodAutoscalerProfile `struct. In that case all changes make to the original `v1alpha2.DatadogPodAutoscalerSpec` will automatically be propagated to profiles and usage in the code will be a bit easier.
However there are two main problems with that
* Profile is an entity inside `v1alpha1` but we very likely want it to generate the latest version of DPA - `v1alpha2`
* There are some fields which we don't want to provide user to change via Profile, namely - `Owner`, `RemoteVersion` and `Target`

Therefore the plan is
* Move internal types from v1alpha2 to common, providing aliases for backward-compatibility.
* Create dedicated Template struct which will replicate Spec but without there three fields.
* Later provide function which will create spec from this template object.

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

* Check that other repos are still compiling after that change.
* Trying to create this object in local k8s cluster.

### Checklist


api/datadoghq/v1alpha1/datadogpodautoscalerprofile_types.go
```diff
@@ -0,0 +1,63 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+package v1alpha1
```
vboulineau, line 6: I don't think you can un-align the `DatadogPodAutoscalerProfile` and the `DatadogPodAutoscaler`, otherwise it creates a confusing mixup of things.
Also the `Profile` naming may imply things we do not currently implement.

api/datadoghq/v1alpha1/datadogpodautoscalerprofile_types.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/DataDog/datadog-operator/api/datadoghq/common"
+)
+
+type DatadogPodAutoscalerProfileConditionType string
```
vboulineau, line 14: nit: needs to be defined close to the `DatadogPodAutoscalerProfileStatus`

api/datadoghq/v1alpha1/datadogpodautoscalerprofile_types.go
```diff
@@ -0,0 +1,63 @@
...
+
+const (
+	DatadogPodAutoscalerProfileReadyCondition DatadogPodAutoscalerProfileConditionType = "Ready"
```
vboulineau, line 17: What would this condition means? How can there be any sense of readiness?

api/datadoghq/v1alpha1/datadogpodautoscalerprofile_types.go
```diff
@@ -0,0 +1,63 @@
...
+type DatadogPodAutoscalerProfileStatus struct {
+	// +listType=map
+	// +listMapKey=type
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	ControlledAutoscalers int32 `json:"controlledAutoscalers"`
+
+	// +optional
+	LastSyncTime *metav1.Time `json:"lastSyncTime,omitempty"`
```
vboulineau, line 33: What do you plan to put in there? Objects are constantly "synced" by a controller, so it usually does not make sense and forces to trigger an `Update` API Call for no other reason than updating the this timestamp.

api/datadoghq/v1alpha1/datadogpodautoscalerprofile_types.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+type DatadogPodAutoscalerProfileConditionType string
+
+const (
+	DatadogPodAutoscalerProfileReadyCondition DatadogPodAutoscalerProfileConditionType = "Ready"
+)
+
+type DatadogPodAutoscalerProfileSpec struct {
+	Template common.DatadogPodAutoscalerTemplate `json:"template"`
```
vboulineau, line 21: Should we inline instead?

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -544,6 +544,124 @@ const (
...
+}
+
+// DatadogPodAutoscalerTemplate contains the autoscaling behavior configuration.
+// These settings can be shared across multiple autoscalers via DatadogPodAutoscalerProfile.
+// +kubebuilder:object:generate=true
+type DatadogPodAutoscalerTemplate struct {
+	// ApplyPolicy defines how recommendations should be applied.
+	// +optional
+	// +kubebuilder:default={}
+	ApplyPolicy *DatadogPodAutoscalerApplyPolicy `json:"applyPolicy,omitempty"`
```
vboulineau, line 644: I am not entirely sure it should be part of a template

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -544,6 +544,124 @@ const (
 	DatadogPodAutoscalerRolloutTriggeredVerticalActionType DatadogPodAutoscalerVerticalActionType = "RolloutTriggered"
 )
 
+// DatadogPodAutoscalerApplyModeV2 specifies if the controller should apply recommendations.
+// +kubebuilder:validation:Enum:=Apply;Preview
+type DatadogPodAutoscalerApplyModeV2 string
```
vboulineau, line 549: I don't think adding aliases is something we want, nor is it required if we define properly a `DatadogPodAutoscalerTemplate` per API Version

api/datadoghq/v1alpha2/datadogpodautoscaler_types.go
vboulineau, line 194: DatadogPodAutoscalerSpec should inline DatadogPodAutoscalerTemplate

api/datadoghq/v1alpha1/datadogpodautoscalerprofile_types.go
ADKosm: I put this meaning: Ready = no changes in managed DPA is required.
So basically when we do reconciliation and see, that we need to perform some updated - create/update/remove DPAs - we unset Ready and put it back when all updated are performed and there are no new updates to do.

Maybe it is a bit of a overkill since all updates are done locally and very quickly 🤷 . In that case we can just remove conditions section completely.

api/datadoghq/v1alpha1/datadogpodautoscalerprofile_types.go
ADKosm: Well, indeed, probably I underestimate the pace of in-cluster operations. 
The more I think of it the more I agree that we can remove typical attributes of a object lifecycle. 

That being said, I still find useful to see a time of last real update on a cluster related to DPA = ts of the last update/create/delete of a DPA in the cluster. What do you think, does it make sense? 🤔

api/datadoghq/v1alpha1/datadogpodautoscalerprofile_types.go
ADKosm: It is indeed tempting to do that, however I explicitly secured some space for adding additional parameters into the spec. 
For example, last time we synced about features of profile, there was an idea, that we can specify behaviour on deletion or other update of the profile. E.g. remove all created DPA on Profile removal or not.
In the end we decided that we will not do this level of control right now, so there is nothing related to it in the code, but I wanted to design even such simple spec now with a potential to be expanded.

api/datadoghq/common/datadogpodautoscaler_types.go
ADKosm: Why? I mean, it looks useful to be able to setup apply policy since it affect how autoscaler will behave 🤔

api/datadoghq/v1alpha2/datadogpodautoscaler_types.go
ADKosm: I initially tried this, but faced the problem that initializing `DatadogPodAutoscalerSpec` struct will be backward incompatible (because now you need to explicitly initialize `Template` struct).
So I decided to go in a more safe way and just recreate this layer.

api/datadoghq/common/datadogpodautoscaler_types.go
ADKosm: I can duplicate all parts (structs,enums,types) which we need for DPA spec.
I initially decided to go with more reusable approach, but I don't see any problem in just doing a proper copy.
Do you agree on full disconnection from original types?

api/datadoghq/v1alpha1/datadogpodautoscalerprofile_types.go
ADKosm: It is indeed a good point to discuss, I don't see technical issues with that per-se, only usability so I think it should be more of a business decision. Will put that question in a channel

api/datadoghq/v1alpha1/datadogpodautoscalerprofile_types.go
vboulineau: As we do not have any rollout feature as of now, it's probably not very useful indeed.

api/datadoghq/v1alpha1/datadogpodautoscalerprofile_types.go
vboulineau: One thing that's commonly done to denote that is the `observedGeneration` field in `Status`. The controller reports there the last `Generation` it has observed (and thus normally handled). Any timestamp that is ever increasing (i.e. each reconcile loop increases this value, like lastUpdate) forces you to continuously run reconcile+Update calls, which is increasing load both on the controller and the APIServer (even though, in the case of `Profile`, the number of objects per cluster will be ~low).

api/datadoghq/v1alpha1/datadogpodautoscalerprofile_types.go
vboulineau: Inlining does not mean that you cannot add more fields, it just means there's no dedicated `key` (like `template`), though most built-in Kubernetes objects referencing templates (`PodTemplateSpec`), they do have an explicit `template` field, so we can go with that.

api/datadoghq/common/datadogpodautoscaler_types.go
vboulineau: Well, the status of an Autoscaler is disjoint from its actual `Spec`, it would imply that you cannot pause autoscaling on a single Autoscaler managed by a template, you would need to pause all autoscalers.

api/datadoghq/common/datadogpodautoscaler_types.go
vboulineau: I meant the opposite, I do not want any duplication, a single type for everything.

api/datadoghq/v1alpha2/datadogpodautoscaler_types.go
vboulineau: The YAML is backward compatible, that's what is important, the fact that the Go struct is not is actually not a problem.

ADKosm:
Close in a favour of https://github.com/DataDog/datadog-operator/pull/2718

PR #2519, opened 2026-01-23
Start DDAI logging standardization

khewonc:
### What does this PR do?

Start standardization logging a bit for DDAI. Goal is to have some kind of structure for what info goes into logs to make troubleshooting easier. Logs now have the following fields:
* namespace
* name
* kind
* reconcileID

`<workload>.xxx` fields are replaced with `object.xxx` within each component reconcile (in addition to component, which was already there):
* object.kind: kind of object being created, e.g. deployment
* object.namespace: namespace of object being created
* object.name: name of object being created

Commits
* https://github.com/DataDog/datadog-operator/commit/d8a67159d4cd4654fa9e0e3f28208a7c0667d9cf: add custom encoder to filter out some controller runtime fields. Remove log in reconciler struct to use context based logging instead. Example log:
```
{"level":"INFO","ts":"2026-01-23T20:14:54.987Z","logger":"controllers.DatadogAgentInternal","msg":"Reconciling DatadogAgentInternal","namespace":"default","name":"datadog-agent","reconcileID":"199ff97c-0b6d-4cae-8f81-cbe8624af0a4","kind":"DatadogAgentInternal"}
```
* https://github.com/DataDog/datadog-operator/commit/472123f13871724ce20fac59fc53a2a78780c9aa: use context based logging for components. Add object fields for component reconciles. Example log:
```
{"level":"INFO","ts":"2026-01-23T20:14:55.743Z","logger":"controllers.DatadogAgentInternal","msg":"Creating Deployment","namespace":"default","name":"datadog-agent","reconcileID":"199ff97c-0b6d-4cae-8f81-cbe8624af0a4","kind":"DatadogAgentInternal","component":"clusterAgent","object.kind":"Deployment","object.namespace":"default","object.name":"datadog-agent-cluster-agent"}
```
* https://github.com/DataDog/datadog-operator/commit/a4786c372e147b86205bda48685923bec7f6b81e: fix tests

### Motivation

https://datadoghq.atlassian.net/browse/AGENTONB-2798

### Additional Notes

Anything else we should know when reviewing?

### Minimum Agent Versions

Are there minimum versions of the Datadog Agent and/or Cluster Agent required?

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

DDAI logs should have namespace, name, kind, and reconcileID at minimum. If the log is about a specific workload, it should also have component, object.kind, object.namespace, and object.name. See above description for example logs

### Checklist

[...]

PR #2698, opened 2026-03-03
Use context based logger

khewonc:
### What does this PR do?

Pass context instead of logger in functions

### Motivation

https://datadoghq.atlassian.net/browse/CONTP-1334
Follow-up to https://github.com/DataDog/datadog-operator/pull/2519

### Additional Notes

Skipping DDA v2 reconcile methods

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Logs will mostly be the same as before. The main differences are:
* `Delete DaemonSet` → `Deleted DaemonSet` for DS deletion. Triggered when a node agent DS exists and the node agent is then disabled via overrides
* `Delete DaemonSet` → `Deleted ExtendedDaemonSet` for EDS deletion. Triggered when a node agent EDS exists and the node agent is then disabled via overrides

### Checklist

### DIFF
--- a/internal/controller/datadogagentinternal/controller_reconcile_v2_helpers.go
+++ b/internal/controller/datadogagentinternal/controller_reconcile_v2_helpers.go
@@ -3,7 +3,6 @@ package datadogagentinternal
 import (
 	"context"
 
-	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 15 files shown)


internal/controller/datadogagentinternal/controller_reconcile_v2_helpers.go
```diff
@@ -87,8 +87,8 @@ func (r *Reconciler) manageFeatureDependencies(logger logr.Logger, features []fe
 }
 
 // overrideDependencies wraps the dependency override logic.
-func (r *Reconciler) overrideDependencies(logger logr.Logger, resourceManagers feature.ResourceManagers, instance *datadoghqv1alpha1.DatadogAgentInternal) error {
-	errs := override.Dependencies(logger, resourceManagers, instance.GetObjectMeta(), &instance.Spec)
+func (r *Reconciler) overrideDependencies(ctx context.Context, resourceManagers feature.ResourceManagers, instance *datadoghqv1alpha1.DatadogAgentInternal) error {
+	errs := override.Dependencies(ctrl.LoggerFrom(ctx), resourceManagers, instance.GetObjectMeta(), &instance.Spec)
```
khewonc, line 91: override.Dependencies is also used in controller v2, which I'm not going to modify to use context based logging since it won't be used starting in operator 1.27. This can be changed to pass context instead of logging then

PR #2700, opened 2026-03-04
[datadog-operator] Expose DCA Agent Sidecar TLS Config

gabedos:
### What does this PR do?

Expose config option + creates RBACs for creating and mounting the cluster agent's certificate onto agent sidecar container it creates in application namespaces.

### Motivation

Expose config option for new Agent feature for TLS communication on agent sidecar in admission controller

### Minimum Agent Versions

This feature works only with newer Agent versions. However, the configmap rbac is fine to be applied everywhere.

* Agent: v7.78.0+
* Cluster Agent: v7.78.0+

### Describe your test plan

Apply the following agent crd config
```yaml
kind: DatadogAgent
metadata:
  name: datadog-test-tls
  namespace: system
spec:
  global:
    credentials:
      apiSecret:
        secretName: datadog-secret
        keyName: api-key
      appSecret:
        secretName: datadog-secret
        keyName: app-key
  features:
    admissionController:
      enabled: true
      agentSidecarInjection:
        enabled: true
        provider: fargate
        clusterAgentTlsVerification:
          enabled: true
          copyCaConfigMap: true
```

Check that the configmaps rbacs are present on the dca role
```
kubectl get clusterrole datadog-test-tls-cluster-agent -o yaml | grep -A 5 "configmaps"
  - configmaps
  verbs:
  - create
  - get
  - list
  - update
```

### Checklist

### DIFF
--- a/internal/controller/datadogagent/feature/admissioncontroller/feature.go
+++ b/internal/controller/datadogagent/feature/admissioncontroller/feature.go
@@ -74,8 +74,8 @@ type AgentSidecarInjectionConfig struct {
 	imageTag                         string
 	selectors                        []*v2alpha1.Selector
 	profiles                         []*v2alpha1.Profile
-	tlsVerificationEnabled           bool
-	tlsVerificationCopyCaConfigMap   bool
+	tlsVerificationEnabled           *bool
+	tlsVerificationCopyCaConfigMap   *bool
 }
 
 type KubernetesAdmissionEventConfig struct {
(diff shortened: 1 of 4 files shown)


cswatt:
For documentation styleguide reasons, can we capitalize `Cluster Agent` and `Agent`?

internal/controller/datadogagent/feature/admissioncontroller/feature.go
```diff
@@ -264,6 +266,16 @@ func (f *admissionControllerFeature) Configure(dda metav1.Object, ddaSpec *v2alp
...
 				}
 			}
+
+			// Configure TLS verification settings
+			if sidecarConfig.ClusterAgentTLSVerification != nil {
+				if sidecarConfig.ClusterAgentTLSVerification.Enabled != nil {
+					f.agentSidecarConfig.tlsVerificationEnabled = apiutils.BoolValue(sidecarConfig.ClusterAgentTLSVerification.Enabled)
+				}
+				if sidecarConfig.ClusterAgentTLSVerification.CopyCaConfigMap != nil {
+					f.agentSidecarConfig.tlsVerificationCopyCaConfigMap = apiutils.BoolValue(sidecarConfig.ClusterAgentTLSVerification.CopyCaConfigMap)
```
levan-m: should we set this (and pass env var to DCA) even if `ClusterAgentTLSVerification.Enabled == false`?

levan-m: would be nice to have unit test confirming expected behavior.

gabedos: I can add a unit test for this to confirm that the values are being set properly.

But I think it _should be fine_ to pass the envvar to DCA even if it's explicitly set to false on the Operator. Generally, I think this should be good to do because Agent defaults can change and I'd want a manually defined `false` by the Operator user to still apply.

PR #2718, opened 2026-03-06
CASCL-1077 v1alpha2 datadog pod autoscaler profile CRD

ADKosm:
### What does this PR do?

Introduces DatadogPodAutoscalerProfile CRD in v1alpha2 to enable GitOps batch onboarding for workload autoscaling. The Profile contains a template with autoscaling behavior configuration that can be applied to multiple DatadogPodAutoscaler (DPA) objects at once.
Please check [this](https://datadoghq.atlassian.net/wiki/spaces/CAUT/pages/6280052935/GitOps+implementation+notes) and [this](https://datadoghq.atlassian.net/wiki/spaces/CAUT/pages/6248235928/brainstorm+GitOps-Based+Batch+Onboarding+Workflow) document to learn more about this project.

### Motivation

Enable users to manage autoscaling configurations centrally via Profiles instead of creating individual DPA objects for each workload. This simplifies GitOps workflows where multiple workloads need the same autoscaling behavior.

Mental model here can be similar to how Kubernetes ReplicaSet manages Pods => DatadogPodAutoscalerProfile will manage DPAs.

### Additional Notes

**API Design:**

- Profile lives in `v1alpha2` package
- DatadogPodAutoscalerTemplate contains shareable autoscaling config: ApplyPolicy, Objectives, Fallback, Constraints, Options. In the future it can be embedded into DatadogPodAutoscalerSpec. For that a construction function is provided - we first will need to update all instantiations in all other repos to use this build method and after this we are free to do embedding in the code.
- Workload-specific fields (TargetRef, Owner, RemoteVersion) remain in DPA spec
- For now Template fields use existing v1alpha2 and common types directly. So no ability to make computable fields for now.

**Status fields:**

- ControlledAutoscalers  - count of managed DPAs
- Strategy - derived strategy: "vertical", "horizontal", or "multi-dimension". Only for illustrative purposes
- Conditions - standard K8s condition object with one state Ready. This state indicates whether the profile is ready and all managed autoscalers are up to date. 

**Printcolumns:**

- Display Name and Description from annotations (`autoscaling.datadoghq.com/profile-name`, `autoscaling.datadoghq.com/profile-description`)
- Strategy, Ready status, Controlled Autoscalers count, Apply Mode, Min/Max Replicas, Age

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Check that other repos are still compiling after that change.
Trying to create this object in local k8s cluster.

### Checklist

### DIFF
--- a/api/datadoghq/v1alpha2/datadogpodautoscalerprofile_types.go
+++ b/api/datadoghq/v1alpha2/datadogpodautoscalerprofile_types.go
@@ -68,8 +68,8 @@ type DatadogPodAutoscalerProfileStatus struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:resource:shortName=dpap
 // +kubebuilder:subresource:status
-// +kubebuilder:printcolumn:name="Display Name",type="string",JSONPath=".metadata.annotations['autoscaling\\.datadoghq\\.com/profile-name']"
-// +kubebuilder:printcolumn:name="Description",type="string",JSONPath=".metadata.annotations['autoscaling\\.datadoghq\\.com/profile-description']"
+// +kubebuilder:printcolumn:name="Name",type="string",JSONPath=".metadata.name"
+// +kubebuilder:printcolumn:name="Description",type="string",JSONPath=".metadata.annotations['kubernetes\\.io/description']"
 // +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=".status.conditions[?(@.type=='Ready')].status"
 // +kubebuilder:printcolumn:name="Controlled Autoscalers",type="integer",JSONPath=".status.controlledAutoscalers"
 // +kubebuilder:printcolumn:name="Apply Mode",type="string",JSONPath=".spec.template.applyPolicy.mode"


api/datadoghq/v1alpha2/datadogpodautoscalerprofile_types.go
```diff
@@ -0,0 +1,105 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/DataDog/datadog-operator/api/datadoghq/common"
+)
+
+// DatadogPodAutoscalerTemplate contains the autoscaling behavior configuration
+// that can be shared across multiple autoscalers via DatadogPodAutoscalerProfile.
+type DatadogPodAutoscalerTemplate struct {
```
ADKosm, line 16: It duplicates fields presented in Spec instead of embedding Template struct because there are instantiations of Spec in other repos ([for example this one](https://github.com/DataDog/dd-source/blob/155bab153680e962a5a8cbe8527f4bad7dbfc7c9/domains/container_apps/shared/libs/autoscaling/datadogpodautoscaler/dpa_test.go#L206-L206)).

The plan is to move all such cases to use `BuildDatadogPodAutoscalerSpecV1Alpha2FromTemplate` instead of base creation and after that we are free to embed.

api/datadoghq/v1alpha2/datadogpodautoscalerprofile_types.go
```diff
@@ -0,0 +1,105 @@
...
+// DatadogPodAutoscalerProfileStatus defines the observed state of DatadogPodAutoscalerProfile.
+type DatadogPodAutoscalerProfileStatus struct {
+	// Conditions represents the latest available observations of the profile's current state.
+	// +listType=map
+	// +listMapKey=type
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// ControlledAutoscalers is the number of DatadogPodAutoscaler objects managed by this profile.
+	// +optional
```
LaurieScs, line 64: It seems to be required and not optional

api/datadoghq/v1alpha2/datadogpodautoscalerprofile_types.go
```diff
@@ -0,0 +1,105 @@
...
+	ControlledAutoscalers int32 `json:"controlledAutoscalers"`
+
+	// Strategy represents the autoscaling strategy derived from the template configuration.
+	// Possible values: "vertical" (VPA only), "horizontal" (HPA only), "multi-dimension" (HPA+VPA).
+	// +optional
+	Strategy string `json:"strategy,omitempty"`
```
LaurieScs: Should it be typed instead of a simple string?

api/datadoghq/v1alpha2/datadogpodautoscalerprofile_types.go
vboulineau: I don't understand the logic there, it's not like upgrading the lib is a kind of automated action? We make the CRDs the way we want and when we upgrade the lib, we update the code using it as well?

api/datadoghq/v1alpha2/datadogpodautoscalerprofile_types.go
vboulineau: If the field is not present in the DPA, it should not be present on the profile. We do not explicitly state this on purpose.

api/datadoghq/v1alpha2/datadogpodautoscalerprofile_types.go
```diff
@@ -0,0 +1,105 @@
...
+	// Template contains the autoscaling behavior configuration to apply to managed DatadogPodAutoscalers.
+	Template DatadogPodAutoscalerTemplate `json:"template"`
+}
+
+// DatadogPodAutoscalerProfileConditionType represents a condition type for DatadogPodAutoscalerProfile.
+type DatadogPodAutoscalerProfileConditionType string
+
+const (
+	// DatadogPodAutoscalerProfileReadyCondition indicates whether the profile is ready and all managed autoscalers are up to date.
+	DatadogPodAutoscalerProfileReadyCondition DatadogPodAutoscalerProfileConditionType = "Ready"
```
vboulineau, line 52: I still don't understand the purpose of this condition. Seems to me like you need more things like `Valid`, `Error` and then fields to represent aggregated state and observed object. (`managedAutoscalers`, `updatedAutoscalers`, `observedGeneration`).

api/datadoghq/v1alpha2/datadogpodautoscalerprofile_types.go
vboulineau: It cannot be required as in case of errors you would not be able to generate a status or would need to put `0` while not true.

api/datadoghq/v1alpha2/datadogpodautoscalerprofile_types.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/DataDog/datadog-operator/api/datadoghq/common"
+)
+
+// DatadogPodAutoscalerTemplate contains the autoscaling behavior configuration
+// that can be shared across multiple autoscalers via DatadogPodAutoscalerProfile.
+type DatadogPodAutoscalerTemplate struct {
+	// ApplyPolicy defines how recommendations should be applied.
+	// +optional
+	// +kubebuilder:default={}
+	ApplyPolicy *DatadogPodAutoscalerApplyPolicy `json:"applyPolicy,omitempty"`
```
vboulineau, line 20: Did we get a decision on including this?

api/datadoghq/v1alpha2/datadogpodautoscalerprofile_types.go
ADKosm: I didn't know that update of datadog-operator is not automated, in that case yeah, this is unneccesary

api/datadoghq/v1alpha2/datadogpodautoscalerprofile_types.go
ADKosm: The original RFC is suggesting to have STRATEGY field for illustrative purposes - https://datadoghq.atlassian.net/wiki/spaces/CAUT/pages/6248235928/brainstorm+GitOps-Based+Batch+Onboarding+Workflow

Since it is a derived field, I decided to just add it to the status. I didn't find a way to express condition using `kubebuilder:printcolumn` mechanism. If you know how to achieve the same result without field in status - tell me :) 

Of we can give up of this column completely

api/datadoghq/v1alpha2/datadogpodautoscalerprofile_types.go
ADKosm: Well, I personally see benefits in both scenarios - when we have it only on DPA and when we have it on Profile level as well.
I think it can be useful to be able to disable all managed autoscalers with one change in profile. 
For disabling one specific autoscaler - I remember some idea from the past discussions what we should be able to use some cluster-local way to disable autoscaling immediately. If we still want to implement this I think in that case having ApplyPolicy in Profile and be able to put a label/config/annotation on DPA locally to disable this specific one - is a good complimentary pack of features.
@vboulineau what do you think?

api/datadoghq/v1alpha2/datadogpodautoscalerprofile_types.go
ADKosm: Removed for now, we can rethink this in the future

api/datadoghq/v1alpha2/datadogpodautoscalerprofile_types.go
ADKosm: And one more argument is that controlling the final mode = horizontal, vertical or multidimensional - is done through ApplyPolicy. If I want VPA-only I will need to setup this structure.
So I am more convinced now that we need this property on profile side that the other way around.

PR #2720, opened 2026-03-09
[CONTP-1397] add admission controller probe configuration

adel121:
### What does this PR do?

Adds configuration for `admission_controller.probe`

### Minimum Agent Versions

* Cluster Agent: v7.78.0

### Describe your test plan

Deploy the operator and apply the following:

```
kind: DatadogAgent
metadata:
  name: datadog
spec:
  features:
    admissionController:
      enabled: true
  override:
    nodeAgent:
      env:
        - name: DD_HOSTNAME
          valueFrom:
            fieldRef:
              fieldPath: spec.nodeName
    clusterAgent:
      replicas: 2
```

After the cluster agent is Running, check the output of `agent status` on the leader cluster agent replica, it should contain a section for the admission controller network probe with successful executions:

```
Connectivity Probe
------------------
Namespace: datadog-admission-probe-dryrun
Total Executions: 4
Success Rate: 25.0%
Last Execution: 2026-03-05T13:06:11Z (success)
Last Successful Probe: 2026-03-05T13:06:11Z
```

Or just check the output of `agent config -a` on the cluster agent and ensure that `admission_controller.probe.enabled` is `true`

### Checklist

### DIFF
--- a/api/datadoghq/v2alpha1/datadogagent_types.go
+++ b/api/datadoghq/v2alpha1/datadogagent_types.go
@@ -1194,6 +1194,7 @@ type AdmissionControllerFeatureConfig struct {
 	CWSInstrumentation *CWSInstrumentationConfig `json:"cwsInstrumentation,omitempty"`
 
 	// Probe holds the admission controller connectivity probe configuration.
+	// +doc-gen:exclude
 	// +optional
 	Probe *AdmissionControllerProbeConfig `json:"probe,omitempty"`
 }
--- a/docs/configuration_public.md
+++ b/docs/configuration_public.md
@@ -63,15 +63,6 @@ spec:
 `features.admissionController.mutation.enabled`
 : Enables the Admission Controller mutation webhook. Default: true
 
-`features.admissionController.probe.enabled`
-: Enables the admission controller connectivity probe. The probe periodically sends dry-run ConfigMap creation requests to verify the webhook is reachable from the API server. Requires Cluster Agent 7.78.0+. Default: true
-
-`features.admissionController.probe.gracePeriod`
-: GracePeriod is the number of seconds to wait at startup before the first probe. Default: 60
-
-`features.admissionController.probe.interval`
-: Is the number of seconds between probe executions. Default: 60
-
 `features.admissionController.registry`
 : Defines an image registry for the admission controller.
 


config/crd/bases/v1/datadoghq.com_datadogagentinternals.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.17.3
```
adel121, line 6: not sure why this changed

config/rbac/role.yaml
```diff
@@ -98,6 +98,9 @@ rules:
   - apiservices
   verbs:
   - '*'
+  - deletecollection
+  - list
+  - watch
```
adel121: Not sure why these keep being added when running `make build`. They seem redundant, we already have `'*'` just above.

internal/controller/datadogagent/feature/admissioncontroller/feature.go
```diff
@@ -159,6 +167,23 @@ func (f *admissionControllerFeature) Configure(dda metav1.Object, ddaSpec *v2alp
 			f.kubernetesAdmissionEvents = &KubernetesAdmissionEventConfig{enabled: true}
 		}
 
+		f.probeConfig = &ProbeConfig{
+			enabled:     defaultProbeEnabled,
```
tbavelier, line 171: We enable it by default, but this requires DCA 7.78.0+. Can we gate in `Configure()` to only configure the probe if it's not nil (already done today) and if DCA version is 78+ ? Something like:

```go
func supportsAdmissionControllerProbe(ddaSpec *v2alpha1.DatadogAgentSpec) bool {
	if clusterAgent, ok := ddaSpec.Override[v2alpha1.ClusterAgentComponentName]; ok {
		if clusterAgent.Image != nil {
			return utils.IsAboveMinVersion(common.GetAgentVersionFromImage(*clusterAgent.Image), probeMinClusterAgentVersion, nil)
		}
	}
	return utils.IsAboveMinVersion(images.ClusterAgentLatestVersion, probeMinClusterAgentVersion, nil)
}
```
and then
```go
		if supportsAdmissionControllerProbe(ddaSpec) {
			f.probeConfig = &ProbeConfig{
				enabled:     defaultProbeEnabled,
				interval:    defaultProbeInterval,
				gracePeriod: defaultProbeGracePeriod,
			}
			if ac.Probe != nil {
				if ac.Probe.Enabled != nil {
					f.probeConfig.enabled = apiutils.BoolValue(ac.Probe.Enabled)
				}
				if ac.Probe.Interval != nil {
					f.probeConfig.interval = *ac.Probe.Interval
				}
				if ac.Probe.GracePeriod != nil {
					f.probeConfig.gracePeriod = *ac.Probe.GracePeriod
				}
			}
		}
```
and then update feature_test to confirm it's not added when version is pinned to .77, and added when .78 ?

api/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -1280,6 +1284,26 @@ type KubernetesAdmissionEventsConfig struct {
...
+
+	// Interval is the number of seconds between probe executions.
+	// Default: 60
+	// +optional
+	Interval *int32 `json:"interval,omitempty"`
+
+	// GracePeriod is the number of seconds to wait at startup before the first probe.
+	// Default: 60
+	// +optional
+	GracePeriod *int32 `json:"gracePeriod,omitempty"`
```
tbavelier, line 1305: nit: we could possibly add kubebuilder:validation markers to have values > 0, right now, users could use `-5` and it would pass admission time. It's fine if DCA code handles this properly / might be overkill as we don't really expect users to configure this field that much

docs/configuration_public.md
```diff
@@ -63,6 +63,15 @@ spec:
...
 
+`features.admissionController.probe.enabled`
+: Enables the admission controller connectivity probe. The probe periodically sends dry-run ConfigMap creation requests to verify the webhook is reachable from the API server. Requires Cluster Agent 7.78.0+. Default: true
+
+`features.admissionController.probe.gracePeriod`
+: GracePeriod is the number of seconds to wait at startup before the first probe. Default: 60
+
+`features.admissionController.probe.interval`
+: Is the number of seconds between probe executions. Default: 60
+
```
tbavelier: Can you add in `_types.go` the marker `+doc-gen:exclude` to disable adding this field to the public doc ? I don't think we want to expose them in the public doc, rather something advanced users might want to tweak.
See https://github.com/DataDog/datadog-operator/pull/2706 for context (you'll need to re-generate files after)

api/datadoghq/v2alpha1/datadogagent_types.go
adel121: The DCA already handles this and fallbacks to a sane default if the provided value is invalid. I think it is fine as it is now.

tbavelier: perfect, nothing to add there then

internal/controller/datadogagent/feature/admissioncontroller/feature.go
adel121: I see, good point. But is this really necessary? If the user uses this with an older version of the DCA, it will simply not work, but will not breaking deployment (worst can happen is injecting env vars that the DCA doesn't really need / user).

Additionally, gating the feature configuration of the DCA version makes it impossible to configure the feature via the operator for custom agent image tags, so I am not sure it is worth pursuing.

tbavelier: > Additionally, gating the feature configuration of the DCA version makes it impossible to configure the feature via the operator for custom agent image tags, so I am not sure it is worth pursuing.

The utility assumes a custom tag is latest if it's not following semver to cover this use case.

> If the user uses this with an older version of the DCA, it will simply not work, but will not breaking deployment (worst can happen is injecting env vars that the DCA doesn't really need / user).

yep, only env vars I guess, but that will put some logs in DCA about unknown config keys I believe ? Agree it's mostly harmless/not as critical as other places where we do need this gate. Can discuss offline tonight: I don't necessarily strongly advocate for it, it's how we've done it in the past

PR #2723, opened 2026-03-09
Add autoscaling failover config to DatadogAgent CR

clamoriniere:
### What does this PR do?

Adds a new `failover.enabled` boolean option under `spec.features.autoscaling.workload` in the DatadogAgent CR to control the autoscaling failover mechanism. When enabled (the default), the operator automatically sets `DD_AUTOSCALING_FAILOVER_ENABLED` and `DD_AUTOSCALING_FAILOVER_METRICS` environment variables on the Cluster Agent and Node Agent components.

### Motivation

Previously, users who enabled workload autoscaling had to manually add `DD_AUTOSCALING_FAILOVER_ENABLED` environment variable overrides to both `clusterAgent` and `nodeAgent` sections. This change makes failover a first-class configuration option that is enabled by default when workload autoscaling is active, removing the need for manual env var overrides.

### Additional Notes

The failover is **enabled by default** (`true`) when workload autoscaling is enabled. Users can explicitly disable it:
```yaml
spec:
  features:
    autoscaling:
      workload:
        enabled: true
        failover:
          enabled: false
```

### Minimum Agent Versions

* Agent: N/A
* Cluster Agent: N/A

### Describe your test plan

- Unit tests updated and extended with two new test cases: explicit failover enabled and failover disabled
- All 7 tests pass
- `go vet` and `gofmt` clean
- CRD, deepcopy, and OpenAPI schemas regenerated

### Checklist

---

PR by Bits
[View session in Datadog](https://app.datadoghq.com/code/a5bdc887-0302-478b-b0d0-17e45bea97be)

Comment @datadog to request changes


internal/controller/datadogagent/feature/autoscaling/feature.go
```diff
@@ -116,10 +118,12 @@ func (f *autoscalingFeature) ManageClusterAgent(managers feature.PodTemplateMana
 			Value: "true",
 		})
 
-		managers.EnvVar().AddEnvVarToContainer(apicommon.ClusterAgentContainerName, &corev1.EnvVar{
-			Name:  DDAutoscalingFailoverEnabled,
-			Value: "true",
-		})
+		if f.failoverEnabled {
```
jennchenn, line 121: Previously I believe fallback would already be enabled by default when workload scaling is activated. Is the purpose of adding the failover config to make it easier for users to disable it?

clamoriniere: mainly to simplify enabling it. 

when I checked in the config package it is still disable by default https://github.com/DataDog/datadog-agent/blob/23415b51b7f24bce503e23b47877373e6ef80a4a/pkg/config/setup/config.go#L1564

drichards-87 (approved):

tbavelier (changes requested):
Discussing offline so "blocking" the PR until then: failover was already always enabled when configured with the operator. Metrics env var was only added on node Agent, not on DCA

clamoriniere:
Ok, when I checked that it was missing,it was indeed in the cluster-agent
Still we need to give a way to disable it in some cases.
I will update the PR

PR #2752, opened 2026-03-12
Clarify CSPM host benchmarks configuration hierarchy

k3nz0:
## What does this PR do? What is the motivation?

Clarifies the relationship between `features.cspm.enabled` and `features.cspm.hostBenchmarks.enabled` in the Operator configuration docs:

- `features.cspm.enabled`: Now states it includes Docker and Kubernetes benchmarks, and must be set to `true` before `hostBenchmarks` takes effect.
- `features.cspm.hostBenchmarks.enabled`: Now specifies it controls Linux host benchmarks specifically, and requires `features.cspm.enabled` to be `true`.
## Additional notes

The current descriptions don't make the hierarchy between these two flags clear, which can be confusing for users configuring CSPM.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -74,7 +74,19 @@ unit_tests:
   before_script:
     - make install-tools
   script:
-    - make test
+    - make ci-test
+
+verify_licenses:
+  stage: test
+  tags: ["arch:amd64"]
+  rules:
+    - if: '$DDR == "true"'
+      when: never
+    - when: on_success
+  before_script:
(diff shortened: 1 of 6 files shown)


k3nz0:
> The doc is automatically generated as part of make generate-docs. I think you'll need to change the types file and then run make generate-docs to change it everywhere:


@khewonc Nice catch. Applied your suggestion!

khewonc:
Closing in favor of https://github.com/DataDog/datadog-operator/pull/2765

Issue #2052, opened 2025-07-18
Add untaint controller to Datadog Operator for startup taint removal


imdevin567:
**Summary**

Add an optional untainting controller to the Datadog Operator that removes a specific taint from a node once the Datadog agent is successfully running on it. This would allow users to enforce that the Datadog agent is the first workload scheduled on new nodes, ensuring observability coverage before any other workloads begin execution.

**Use Case**

In many environments, it's important to ensure observability agents like the Datadog agent are running before any application workloads are scheduled on a node. One pattern to achieve this is to apply a "startup" taint (e.g., `node.datadoghq.com/startup=datadog:NoSchedule`) to the node pool at provisioning time. The taint blocks all pods except those that tolerate it (e.g., the Datadog agent).

Currently, there's no automated way to remove this taint once the agent is confirmed to be running. This requires out-of-band scripting or external controllers, which adds operational overhead and complexity.

By having the Datadog Operator manage this behavior, the system can:

* Ensure the agent is the first workload on a node
* Automatically remove the startup taint once the agent is successfully running
* Reduce complexity and eliminate the need for custom automation

**Proposal**

Introduce a new optional controller in the Datadog Operator that:

1. Watches nodes with a configurable taint key (e.g., `node.datadoghq.com/startup=datadog:NoSchedule`)
2. Detects when a healthy Datadog agent pod is running on that node
3. Removes the configured taint from the node

Configuration could be introduced via the `DatadogAgent` CRD, for example:

```yaml
spec:
  untaint:
    enabled: true
    taintKey: "node.datadoghq.com/startup"
    taintValue: "datadog"
    taintEffect: "NoSchedule"
```

**Similar Patterns in the Wild**

Istio provides a [similar feature](https://github.com/istio/istio/blob/master/manifests/charts/istio-control/istio-discovery/values.yaml#L41) in their Operator to ensure their control plane components are prioritized before other workloads.

**Benefits**

* Ensures observability is initialized before application workloads
* Simplifies node bootstrapping workflows in cluster autoscaling environments
* Reduces reliance on custom scripts or external controllers
* Aligns with patterns used in other Operators (e.g., Istio)

Thank you for considering this feature! I'd be happy to contribute or test this functionality if it's accepted.

drcrees:
+1, we are running on EKS and experience an issue occasionally where user pods are able to start before the node agent is ready.

levan-m:
[...]

PR #2753, opened 2026-03-12
Untaint controller

levan-m:
### What does this PR do?
CONTP-1607, CONTP-1613

Untaint controller to prevent scheduling workloads before Agent when nodes join cluster.  

When enabled, untaint controller will watch nodes and remove `agent.datadoghq.com/not-ready=presence:NoSchedule` taints once Agent reaches ready state. Until then pods can't be scheduled on the node unless they tolerate this taint. This is effective during cluster scale out when new nodes join with the taint. Once removed Operator will not add taint back for any reason.

Feature can be enabled via `--untaintControllerEnabled=flag` arg, optionally one can enable taint removal events using `DD_UNTAINT_CONTROLLER_EVENTS_ENABLED=true` env var.

Commits

0b567c68e5d518334adc4ac38285b1e7445ddb59 - POC

359ab79dc6f95ad57a17f723c77a3b83518310bb - improvements:
  - Conflict-handling fix, requeues in 1s instead of returning an error (no exponential backoff).
  - Defensive JSON-patch marshaling. Guarantees taint slices serialize as [] rather than null in both test and replace ops.
  - Field-indexer collision-proofing. Renames the pod-by-nodeName cache index to a namespaced key untaint.spec.nodeName.
  - Two-tier timeout policy. Adds DD_UNTAINT_CONTROLLER_TIMEOUT (pod-exists-not-ready, default 10m) and DD_UNTAINT_CONTROLLER_SCHEDULING_TIMEOUT (no pod, default 5m), with DD_UNTAINT_CONTROLLER_TIMEOUT_POLICY (remove
  default | keep) controlling whether to untaint anyway or just emit observability.
  - Node watch added. Filters to nodes carrying the target taint, so nodes that never get an agent pod still get reconciled.
  - Pod predicate widened. Enqueues on any agent-pod Create/Update/Delete (not just Ready transitions) so the readiness clock can't stall under policy=keep or when NotReady pods appear between reconciles.
  - Log error on config errors  - similar to other controllers untaint-controller startup error is logged — no silent fallback to defaults and other controllers continue running.
  - New metric. untaint_taint_timeouts_total{reason, policy} for alerting on stuck nodes; existing taint_removals_total and taint_removal_latency_seconds kept.
  - Tests. Adds a unit suite (fake client + fake clock, full coverage of reconcile outcomes / predicates / conflict paths / env-var errors); rewrites the integration suite around fresh per-test nodes and adds
  scheduling/readiness-timeout scenarios.
  - Docs. Adds docs/untaint_controller.md.

### Motivation

#2052 and similar requests through other channels.

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Can be tested locally on `kind` cluster:
1. Create kind cluster with multiple nodes (assume 1 control plane, 3 workers).
2. Deploy Operator with feature enabled.
  ```yaml
        args:
        - --untaintControllerEnabled=true
        ...
        env:
        - name: DD_UNTAINT_CONTROLLER_EVENTS_ENABLED
          value: "true"
  ```
4. Taint 2 worker nodes
```
kubectl taint nodes tainted-worker agent.datadoghq.com/not-ready=presence:NoSchedule
kubectl taint nodes tainted-worker2 agent.datadoghq.com/not-ready=presence:NoSchedule
```
5. Create nginx deployment with 3 replicas and antiaffinity not to schedule two on same node
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-test
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx-test
  template:
    metadata:
      labels:
        app: nginx-test
    spec:
      affinity:
        podAntiAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
          - labelSelector:
              matchExpressions:
              - key: app
                operator: In
                values:
                - nginx-test
            topologyKey: kubernetes.io/hostname
      containers:
      - name: nginx
        image: nginx
``` 
6. One replica should be schedule on tainted-worker3, other two will be pending.
7. Apply DatadogAgent manifest (add toleration to agent pods)
```yaml
kind: DatadogAgent
metadata:
  name: datadog-agent
spec:
  global:
    credentials:
      apiSecret:
        secretName: datadog-secret
        keyName: api-key
    kubelet:
      tlsVerify: false
  features:
    logCollection:
      enabled: true
      containerCollectUsingFiles: true
      containerCollectAll: true
  override:
    nodeAgent:
      tolerations:
        - key: agent.datadoghq.com/not-ready
          operator: Exists
          effect: NoSchedule
```
8. Agents will be schedule to all nodes, once Agents become ready on tainted-worker and tainted-worker2 nginx pods will be scheduled too.

```
# monitor events and taint removal
kubectl get event -w -n default | grep -i taint
kubectl get nodes -w -o custom-columns='NAME:.metadata.name,TAINTS:.spec.taints[*].key'
```

### Checklist


docs/untaint_controller.md
```diff
@@ -0,0 +1,79 @@
...
+This feature was introduced in Datadog Operator v1.x and is currently in beta.
+
+## Overview
+
+The Untaint controller watches Kubernetes Nodes carrying the taint
+`agent.datadoghq.com/not-ready=presence:NoSchedule` and removes it once the
+Datadog Agent pod on that node is `Ready`. It is intended to run alongside a
+separate mechanism (cluster-autoscaler hook, CCM, admission webhook, etc.)
+that adds the taint to new nodes. The use case is keeping workloads off a
+node until the Datadog Agent is Ready, and recovering gracefully if the agent never
```
janine-c:
```suggestion
node until the Datadog Agent is Ready, and recovering gracefully if the Agent never
```

docs/untaint_controller.md
```diff
@@ -0,0 +1,79 @@
...
+becomes Ready.
+
+Agent pods are matched by the label `agent.datadoghq.com/component=agent` in
+the operator's watched namespaces (`WATCH_NAMESPACE` /
+`DD_AGENT_WATCH_NAMESPACE`).
+
+If the agent pod never reaches Ready on a tainted node, a configurable timeout
```
janine-c:
```suggestion
If the Agent pod never reaches Ready on a tainted node, a configurable timeout
```

docs/untaint_controller.md
```diff
@@ -0,0 +1,79 @@
...
+policy ensures the node is never permanently unschedulable. Two clocks cover
+the two failure modes:
+
+- **Readiness timeout** — the agent pod is on the node but not Ready. Clock:
```
janine-c:
```suggestion
- **Readiness timeout** — the Agent pod is on the node but not Ready. Clock:
```

docs/untaint_controller.md
```diff
@@ -0,0 +1,79 @@
...
+  `pod.Status.StartTime`. Pod recreation restarts the window; container
+  restarts inside the same Pod do not.
```
janine-c:
```suggestion
  restarts inside the same pod do not.
```

docs/untaint_controller.md
```diff
@@ -0,0 +1,79 @@
...
+- **Scheduling timeout** — no agent pod is on the node. Clock:
```
janine-c:
```suggestion
- **Scheduling timeout** — no Agent pod is on the node. Clock:
```

docs/untaint_controller.md
```diff
@@ -0,0 +1,79 @@
...
+  `node.metadata.creationTimestamp`. The expected path when a DaemonSet never
+  schedules a pod onto the node (taint not tolerated, missing labels, etc.).
+
+A Pod-recreation crash-loop faster than the readiness window can hold a node
```
janine-c:
```suggestion
A pod-recreation crash-loop faster than the readiness window can hold a node
```

docs/untaint_controller.md
```diff
@@ -0,0 +1,79 @@
...
+tainted indefinitely; run with `policy=keep` and alert on
+`untaint_taint_timeouts_total{policy="keep"}` to catch this.
+
+The controller removes only this fixed taint and does not add it; both
+timeouts are global and cannot be tuned per Node (Group), DDA or DAP.
```
janine-c:
```suggestion
timeouts are global and cannot be tuned per Node (Group), DDA, or DAP.
```

docs/untaint_controller.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Prerequisites
+
+- Operator v1.x+
+- Tested on Kubernetes versions >= `1.27.0`
```
janine-c:
```suggestion
- Tested on Kubernetes 1.27.0+
```

docs/untaint_controller.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Enabling the Untaint controller
```
janine-c:
```suggestion
## Enable the Untaint controller
```

docs/untaint_controller.md
```diff
@@ -0,0 +1,79 @@
...
+
+```yaml
+args:
+  - --untaintControllerEnabled=true
+```
+
+## Configuration
+
+All other tuning knobs are environment variables on the operator pod. Values
+use Go's `time.ParseDuration` format (`90s`, `5m`, `1h`, …). Invalid values
```
janine-c:
```suggestion
use Go's `time.ParseDuration` format (`90s`, `5m`, `1h`, etc.). Invalid values
```

docs/untaint_controller.md
```diff
@@ -0,0 +1,79 @@
...
+| `DD_UNTAINT_CONTROLLER_TIMEOUT`            | `10m`    | Readiness timeout. Tune to the upper bound of legitimate agent startup on your nodes; 2–5m is often enough on clusters with cached images.                                                                                                                                                          |
+| `DD_UNTAINT_CONTROLLER_SCHEDULING_TIMEOUT` | `5m`     | Scheduling timeout. Set larger than your scheduler retry window; raise it on clusters with large pending queues or aggressive autoscaling.                                                                                                                                                          |
+| `DD_UNTAINT_CONTROLLER_TIMEOUT_POLICY`     | `remove` | Action when a timeout fires. `remove` untaints the node anyway (favors scheduling availability over telemetry — default, lowest operational risk). `keep` leaves the taint in place and emits a Warning event (favors telemetry; pair with an alert on the timeout counter to surface stuck nodes). |
```
janine-c:
```suggestion
| `DD_UNTAINT_CONTROLLER_TIMEOUT_POLICY`     | `remove` | Action when a timeout fires. `remove` untaints the node anyway (favors scheduling availability over telemetry; lowest operational risk). `keep` leaves the taint in place and emits a Warning event (favors telemetry; pair with an alert on the timeout counter to surface stuck nodes). |
```

docs/untaint_controller.md
```diff
@@ -0,0 +1,79 @@
...
+| `DD_UNTAINT_CONTROLLER_EVENTS_ENABLED`     | `false`  | Emit Kubernetes Events on Nodes for taint removals and timeout decisions.                                                                                                                                                                                                                           |
+
+## Observability
+
+Metrics, under the `untaint` Prometheus subsystem:
+
+- `untaint_taint_removals_total` — counter, every taint removal regardless of cause.
+- `untaint_taint_removal_latency_seconds` — histogram, time between pod Ready and taint removal.
+- `untaint_taint_timeouts_total{reason, policy}` — counter, timeout decisions. `reason` ∈ {`readiness`, `scheduling`}; `policy` ∈ {`remove`, `keep`}. Alert on `policy="keep"` to investigate stuck nodes.
```
janine-c: We don't seem to use this `∈` notation very often. If you think users will understand it, leave it, but I like to err on the side of clarity just in case a user hasn't encountered it before.

docs/untaint_controller.md
```diff
@@ -0,0 +1,79 @@
...
+
+Kubernetes Events (gated by `DD_UNTAINT_CONTROLLER_EVENTS_ENABLED=true`):
+
+- `TaintRemoved` (Normal) — taint removed because the agent pod became Ready.
```
janine-c:
```suggestion
- `TaintRemoved` (Normal) — taint removed because the Agent pod became Ready.
```

janine-c (approved):
Some very boring style edits that you can take or leave 🙂

docs/untaint_controller.md
levan-m: I'll replace it with `in`.

Makefile
```diff
@@ -166,7 +166,7 @@ docker-build: generate docker-build-ci docker-build-check-ci
 # For local use
 .PHONY: docker-build-ci
 docker-build-ci:
-	docker build . -t ${IMG} --build-arg FIPS_ENABLED="${FIPS_ENABLED}" --build-arg LDFLAGS="${LDFLAGS}" --build-arg GOARCH="${GOARCH}"
+	docker build . -t ${IMG} --build-arg FIPS_ENABLED="${FIPS_ENABLED}" --build-arg LDFLAGS="${LDFLAGS}" --build-arg GOARCH="${GOARCH}" --platform=linux/${GOARCH}
```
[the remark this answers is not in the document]
levan-m: this was supposed to be local-only change.

internal/controller/setup.go
```diff
@@ -36,7 +37,10 @@ type SetupOptions struct {
 	SupportExtendedDaemonset      ExtendedDaemonsetOptions
 	SupportCilium                 bool
 	CredsManager                  *config.CredentialManager
+	Creds                         config.Creds
+	SecretRefreshInterval         time.Duration
 	DatadogAgentEnabled           bool
+	DatadogAgentInternalEnabled   bool
```
adel121: Are the 3 fields added here needed? 

I don't see where they are used/referenced, and they seem unrelated to the untain controller.

internal/controller/suite_v2_test.go
```diff
@@ -88,6 +88,13 @@ var _ = BeforeSuite(func(ctx context.Context) {
 	node2 := testutils.NewNode("node2", nil)
 	Expect(k8sClient.Create(context.Background(), node2)).Should(Succeed())
 
+	// Configure the untaint controller via its env vars before
+	// SetupControllers is called (NewUntaintReconciler reads them on construction).
+	Expect(os.Setenv(EnvEventsEnabled, "true")).Should(Succeed())
+	Expect(os.Setenv(EnvReadinessTimeout, (4 * time.Second).String())).Should(Succeed())
+	Expect(os.Setenv(EnvSchedulingTimeout, (4 * time.Second).String())).Should(Succeed())
+	Expect(os.Setenv(EnvTimeoutPolicy, string(PolicyRemove))).Should(Succeed())
```
adel121, line 96: Curious what could cause this assertion to fail?

What can make setting an env var fail ?

internal/controller/suite_v2_test.go
```diff
@@ -107,6 +114,10 @@ var _ = BeforeSuite(func(ctx context.Context) {
 		DatadogMonitorEnabled:      true,
 		DatadogAgentProfileEnabled: true,
 		V2APIEnabled:               true,
+		// Untaint controller is enabled for integration tests. Tuning knobs
+		// are set via env vars (see setUntaintEnv below) so timeout-path
```
adel121: Nit: I think the env vars are now set above; there is no `setUntaintEnv` below,

internal/controller/untaint_controller.go
```diff
@@ -0,0 +1,515 @@
...
+		if ts, ok := podReadyTransition(&podList.Items[i]); ok {
+			readyAt, anyReady = ts, true
+			break
+		}
+	}
+	if anyReady {
+		result, err := r.removeTaint(ctx, node)
+		if err != nil {
+			return result, err
+		}
```
adel121, line 224: Should we track these with prometheus metrics as well ?

internal/controller/untaint_controller.go
```diff
@@ -0,0 +1,515 @@
...
+		if result.RequeueAfter > 0 {
+			return result, nil
+		}
+		log.Info("Removed agent-not-ready taint from node")
+		metrics.TaintRemovalsTotal.Inc()
+		if !readyAt.IsZero() {
+			metrics.TaintRemovalLatency.Observe(r.clock.Since(readyAt).Seconds())
+		}
```
adel121, line 232: Why shouldn't we emit the metric when the pod is instantly ready? (Although in practice I believe the pod will take more than 1 second to be ready, but just asking if there is a reason to skip this in theory)

internal/controller/untaint_controller.go
```diff
@@ -0,0 +1,515 @@
...
+		if err != nil {
+			return result, err
+		}
+		if result.RequeueAfter > 0 {
+			return result, nil
+		}
+		log.Info("Removed agent-not-ready taint from node")
+		metrics.TaintRemovalsTotal.Inc()
+		if !readyAt.IsZero() {
+			metrics.TaintRemovalLatency.Observe(r.clock.Since(readyAt).Seconds())
```
adel121, line 231: I am not familiar with the kubernetes "Real Clock" object, but is this a global clock synchronized clusterwide ?

Is their a risk of drift between the pod events timestamps and the global clock ? 

(Even if there is a small drift, I don't think it is a big issue for us, but just asking out of curiosity and for the sake of understanding)

internal/controller/untaint_controller.go
```diff
@@ -0,0 +1,515 @@
...
+
+	// Otherwise we're in a timeout-evaluation path. Two cases:
+	//   1. Any pod exists on the node → readiness timeout
+	//      (clock = max(pod.Status.StartTime), so a fresh restart resets it).
+	//      If pods exist but no StartTime is set yet (very early lifecycle —
+	//      kubelet hasn't transitioned through PodInitializing yet), requeue
+	//      after the readiness window so we re-check once StartTime appears.
+	//      We do NOT fall through to the scheduling clock here: a pod *is*
+	//      scheduled, so the "no agent pod ever scheduled" condition does not
+	//      hold and applying the scheduling timeout would be incorrect.
```
adel121: If I understand correctly, this is assuming that the agent is the only pod that is tolerating the no-schedule taint. 

If this is the case, I see 2 potential issues:
- What if the user wanted to exclude specific pods from being blocked? I think this can happen easily if the user is running their own CSI driver (not referring to Datadog CSI Driver, can be any system critical component that should not depend on the existence of the agent itself). 
- I think the current implementation will break (or at least cause friction) if we wanted in the future to support covering other components with this controller. The example I have in mind now is Datadog CSI Driver + Datadog Agent Daemonset. 

Wouldn't it be more reliable to filter for target pods ? (here it is only the agent pods, but should be configurable to support multiple selectors I assume)

internal/controller/untaint_controller.go
```diff
@@ -0,0 +1,515 @@
...
+	log := r.log.WithValues("node", node.Name, "reason", reason, "policy", policy, "elapsed", elapsed)
+
+	metrics.TaintTimeoutsTotal.WithLabelValues(reason, string(policy)).Inc()
+	if policy == PolicyKeep {
+		// logr has no Warn level; the codebase convention is log.Error(nil, ...)
+		// to surface a warning-level event without an associated Go error.
+		log.Error(nil, "Untaint controller timeout fired; keeping taint per policy=keep — operator action may be required")
+	} else {
+		log.Info("Untaint controller timeout fired")
+	}
```
adel121: Should we include the node name in the logs to make investigations/debugging easier?

internal/controller/untaint_controller.go
```diff
@@ -0,0 +1,515 @@
...
+
+	result, err := r.removeTaint(ctx, node)
+	if err != nil {
+		return result, err
+	}
+	if result.RequeueAfter > 0 {
+		return result, nil
+	}
+	log.Info("Removed agent-not-ready taint from node by timeout policy")
```
adel121: Same comment as above

adel121 (commented):
Left minor comments/questions

internal/controller/setup.go
levan-m: thanks for catching this! I botched the merge, cleaned up now.

internal/controller/suite_v2_test.go
levan-m: I don't really know but since `Setenv` returns error we assert it succeeds here.

internal/controller/suite_v2_test.go
levan-m: thanks, this is stale and cleaned up.

internal/controller/untaint_controller.go
levan-m: Yes we should, added.

internal/controller/untaint_controller.go
levan-m: [clock](https://pkg.go.dev/k8s.io/utils@v0.0.0-20260507154919-ff6756f316d2/clock) is `time.Time` wrapper which makes writing test easier, like PassiveClock is initialized to `now` and doesn't change.

internal/controller/untaint_controller.go
levan-m: iirc this was added for test safety, in practice this should be initialized.

internal/controller/untaint_controller.go
levan-m: Are you referring to timeout behavior specifically or in general? 
* If user wants to exclude specific pods from being blocked they'll add tolerations. That would be case for CSI, cilium, some other infra daemonsets. Applications won't have that toleration and they'll be blocked until agent is ready.
* To support CSI driver, which we plan to, we can extend this controller. It will handle new taint, cache filter will change. Agent DS will not tolerate CSI driver taints to ensure CSI pods start first (if that's the intention).

internal/controller/untaint_controller.go
levan-m: good point, added in several places.

internal/controller/untaint_controller.go
levan-m: ✅

internal/controller/untaint_controller.go
adel121: I am failing to find it, are you sure it was pushed? 

To be clear, I was referring to emitting a prometheus metric in the `if err != nil` branch so we can track errors on dashboards easily.

internal/controller/untaint_controller.go
adel121: I see. 

But then we are comparing time values sourced from 2 different clocks.

The `k8s.io/utils.Clock` is the clock of the node on which the operator pod is running. 
The pod events timestamp are coming from the clock of the node on which the pod is running (or from the api server's clock, didn't verify, but certainly not from the operator's node clock). 

If there is a significant clock skew in the cluster, the comparison becomes irrelevant; we will compare timestamps sourced from 2 different clocks that are not synchronized. 

Was this evaluated? At least identifying what could happen in the worst case scenario; even if we don't handle it (to avoid over-engineering the implementation), we could at least document it.

For example, if the operator's local clock is ahead of the api server's clock by more than readiness timeout, we can have unexpected behavior like tainted node losing the taint before the agent is actually scheduled and ready.

But now I realize this is an extreme edge case, I don't think it is worth handling, but not sure if it is worth documenting that this assumes the cluster nodes clocks are synchronized (with NTP for example), but most cloud based clusters are time-synchronized if I am not mistaken.

internal/controller/untaint_controller.go
adel121: > If user wants to exclude specific pods from being blocked they'll add tolerations. That would be case for CSI, cilium, some other infra daemonsets. Applications won't have that toleration and they'll be blocked until agent is ready.

But if I am not mistaken, the code doesn't look for agent pods to consider the node as ready to lose the taint. 
It just accepts any pod (see [here](https://github.com/DataDog/datadog-operator/blob/302a746b9822900483a6a0efe3ea864e18dcf9c5/internal/controller/untaint_controller.go#L252-L263)). So for example, if the user adds the toleration to some infra pod, and it gets scheduled first on the node, then the operator will remove the taint from the node, even if the agent pod is not yet scheduled and healthy. 

> To support CSI driver, which we plan to, we can extend this controller. It will handle new taint, cache filter will change. Agent DS will not tolerate CSI driver taints to ensure CSI pods start first (if that's the intention).

This is clear, but I still see the same concern as the previous point.

levan-m: `podList` here is [list](https://github.com/DataDog/datadog-operator/blob/302a746b9822900483a6a0efe3ea864e18dcf9c5/internal/controller/untaint_controller.go#L214-L219) of agent pods. 
```go
	podList := &corev1.PodList{}
	labelSelector := labels.SelectorFromSet(map[string]string{
		common.AgentDeploymentComponentLabelKey: constants.DefaultAgentResourceSuffix,
	})
	if err := r.client.List(ctx, podList,
		client.MatchingLabelsSelector{Selector: labelSelector},
		client.MatchingFields{untaintPodNodeIndex: req.Name},
```
idea is that we check timeout against to most recent agent pod start. I'll fix the comment as it doesn't explicitly mention that we are looking at agent pods.

internal/controller/untaint_controller.go
levan-m: Ah, i misunderstood your original question. Yes time skew could happen and may lead to untainting too early or too late but given our timeouts are in minutes (for startup and scheduling failures) in practice should have no impact. Well managed clusters skew should be well under 1sec.

internal/controller/untaint_controller.go
levan-m: This was added in `removeTaint` not to count update conflict errors which get requeued/retried quicly.
https://github.com/DataDog/datadog-operator/blob/1b7739e03a418ca7d179762a1a990c1a1720864c/internal/controller/untaint_controller.go#L399
https://github.com/DataDog/datadog-operator/blob/1b7739e03a418ca7d179762a1a990c1a1720864c/internal/controller/untaint_controller.go#L414

internal/controller/untaint_controller.go
adel121: you are right, I missed the label selector in the `client.List`.

But then now I see that the controller hardcodes the selector to match agent pods only. So currently there is no way to dynamically support other target pods. But this is non-blocking for this PR, can be refactored in the following PRs adding support for CSI driver 👍

adel121 (approved):
Thanks for addressing all the comments and answering all questions 🙇

docs/untaint_controller.md
```diff
@@ -0,0 +1,79 @@
+# Untaint Controller
+
+This feature was introduced in Datadog Operator v1.x and is currently in beta.
```
levan-m:
```suggestion
This feature was introduced in Datadog Operator v1.28 and is currently in preview.
```

PR #2772, opened 2026-03-18
[CONTP-1283] Fix missing workqueue_* metrics in /metrics endpoint

### LINKED ISSUE #2771: [BUG] workqueue_* metrics missing from /metrics endpoint after controller-runtime v0.20.0 upgrade
### Pre-submission Checklist

### Operator version

1.23.1

2.18.1

### Bug Report

**TL;DR:** The 7 standard `workqueue_*` metrics (including `workqueue_depth`) are missing from the operator's `/metrics` endpoint. These metrics were available transitively via controller-runtime v0.19.x but disappeared when the operator upgraded to controller-runtime v0.20.x (operator v1.16.0+). Without `workqueue_depth`, there is no way to detect reconciliation queue saturation under load. We're willing to contribute a PR with the fix (~130 LOC).

**What happened:**

The operator's `/metrics` endpoint does not expose any `workqueue_*` metrics, making it impossible to monitor reconciliation queue health when managing large numbers of `DatadogMonitor` or `DatadogSLO` CRs. This affects all controllers in the operator (DatadogAgent, DatadogMonitor, DatadogSLO, etc.) since workqueue metrics are per-controller-queue.

Without these metrics we cannot:

- **Detect queue saturation** — `workqueue_depth` (current backlog size) is the only way to know if the operator is falling behind. The existing `controller_runtime_active_workers` shows workers are busy but not whether 10 or 4,000 items are waiting.
- **Isolate latency sources** — `workqueue_queue_duration_seconds` separates queue wait time from processing time. A CR might take 10s end-to-end but only 200ms to reconcile — the other 9.8s is queue wait, invisible without this metric.
- **Monitor retry storms** — `workqueue_retries_total` reveals if failed reconciliations are creating churn.
- **Catch straggler reconciliations** — `workqueue_longest_running_processor_seconds` surfaces individual CRs that are stuck.

We discovered this while load testing the operator with thousands of DatadogMonitor and DatadogSLO CRs, where queue visibility is essential for capacity planning and performance analysis.

**Verification:**

```bash
$ kubectl port-forward deploy/datadog-operator 8383:8383
$ curl -s http://localhost:8383/metrics | grep -c workqueue
0
```

**What I expected:**

The operator should expose the standard controller-runtime workqueue metrics on its `/metrics` endpoint, consistent with how other `controller_runtime_*` metrics are available.

**Root Cause:**

This regression was introduced in **operator v1.16.0** ([PR #1924](https://github.com/DataDog/datadog-operator/pull/1924)), which upgraded controller-runtime from v0.19.0 to v0.20.4.

In controller-runtime v0.20.0 ([PR #3014](https://github.com/kubernetes-sigs/controller-runtime/pull/3014)), workqueue metric registration was moved from the public `pkg/metrics/workqueue.go` — which had an `init()` that auto-registered metrics and called `workqueue.SetProvider()` — to the internal `pkg/internal/metrics/workqueue.go`, which is only imported by the priority queue implementation (`pkg/controller/priorityqueue/`). Since the operator uses standard workqueues, the internal package is never imported, the `init()` never executes, and the 7 workqueue metrics are never registered. The public file was reduced to just metric name constants.

This was a side effect of the upstream priority queue feature ([issue #2374](https://github.com/kubernetes-sigs/controller-runtime/issues/2374)), which moved metric registration to an internal package. Operators that depended on the transitive registration need to explicitly opt back in.

**Proposed Solution:**

We're willing to contribute a PR if maintainers agree this is valuable. The implementation would:

1. Register the 7 workqueue Prometheus metrics with `metrics.Registry`
2. Implement a `WorkqueueMetricsProvider` satisfying `workqueue.MetricsProvider`
3. Call `workqueue.SetProvider()` during operator initialization

Reference implementation: [controller-runtime v0.19.0 pkg/metrics/workqueue.go](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.19.0/pkg/metrics/workqueue.go) (~130 lines)

**Alternative:** If registering all 7 metrics is too invasive, even just `workqueue_depth` would provide the most critical queue saturation visibility.

### Steps to Reproduce

1. Deploy the operator with `collectOperatorMetrics: true` (the default)
2. Port-forward to the operator pod: `kubectl port-forward deploy/datadog-operator 8383:8383`
3. Query the metrics endpoint: `curl -s http://localhost:8383/metrics | grep workqueue`
4. Observe zero results

### Environment

**Kubernetes version:** 1.34 (EKS)
**Helm version:** 3.x
**Cloud provider:** AWS
**controller-runtime version (in operator):** v0.20.4

### Additional Context

The 7 missing workqueue metrics are:

| Metric | Type | What It Measures |
|---|---|---|
| `workqueue_depth` | gauge | Current number of items waiting in the reconciliation queue |
| `workqueue_adds_total` | counter | Total items enqueued |
| `workqueue_queue_duration_seconds` | histogram | Time items wait in queue before processing begins |
| `workqueue_work_duration_seconds` | histogram | Time spent in the `Reconcile()` call per item |
| `workqueue_retries_total` | counter | Number of items re-enqueued after failed reconciliation |
| `workqueue_unfinished_work_seconds` | gauge | Sum of in-progress work durations |
| `workqueue_longest_running_processor_seconds` | gauge | Duration of the longest-running reconciliation |

- The `datadog.operator.*` namespace metrics (e.g., `datadog.operator.reconcile.success`, `datadog.operator.*.custom_resource.count`) sent via the `-operatorMetricsEnabled=true` path do not include workqueue metrics.
- Confirmed the same issue persists on the operator's `main` branch (controller-runtime v0.22.4) — `pkg/metrics/workqueue.go` still contains only constants, no registration.

tbavelier:
### What does this PR do?

Restores the 7 standard `workqueue_*` Prometheus metrics that disappeared from the operator's `/metrics` endpoint after the controller-runtime upgrade to v0.20+:

- `workqueue_depth`
- `workqueue_adds_total`
- `workqueue_queue_duration_seconds`
- `workqueue_work_duration_seconds`
- `workqueue_retries_total`
- `workqueue_unfinished_work_seconds`
- `workqueue_longest_running_processor_seconds`

### Motivation

Fixes #2771. These metrics are essential for observing controller backlog, processing latency, and retry storms.

### Additional Notes

**Root cause** — a `sync.Once` race in `workqueue.SetProvider()`:

`cmd/main.go` imports `k8s.io/kube-aggregator`, which transitively pulls in `k8s.io/apiserver/pkg/storageversion`, which directly imports `k8s.io/component-base/metrics/prometheus/workqueue`. Because Go initialises packages depth-first and `kube-aggregator` appears before `sigs.k8s.io/controller-runtime` in the import list, `component-base`'s `init()` wins the `sync.Once` in `workqueue.SetProvider()`. This routes all workqueue metrics to the k8s *legacy* Prometheus registry, not to controller-runtime's `ctrlmetrics.Registry` — the registry hardcoded in the metrics server (line 221 of `pkg/metrics/server/server.go`).

The metric *descriptor* vectors are registered in `ctrlmetrics.Registry` by `sigs.k8s.io/controller-runtime/pkg/internal/metrics`, but no label combinations are ever populated because the workqueue itself is using the component-base provider. Prometheus therefore emits nothing for those metric names.

**Why other approaches don't work:**
- Reordering imports has no effect — Go `init()` ordering is determined by the dependency graph, not import list order.
- Adding our own `SetProvider()` call in `main()` is too late — all `init()` functions run before `main()` starts.
- Registering the 7 metrics manually into `ctrlmetrics.Registry` causes a `MustRegister` panic because `pkg/internal/metrics.init()` already registered some of them with different labels (`workqueue_depth` uses `{name, controller, priority}` vs `{name, controller}`).
- There is no `Registry` field on `metricsserver.Options` to substitute a combined gatherer.

**Fix** — enable `UsePriorityQueue` globally on the manager:

The [priority queue](https://github.com/kubernetes-sigs/controller-runtime/issues/2374) **bypasses `SetProvider` entirely**: it directly instantiates `metrics.WorkqueueMetricsProvider{}` on each queue, writing observations to `ctrlmetrics.Registry`. Setting `UsePriorityQueue: ptr.To(true)` in `config.Controller` applies this to every controller managed by the operator.

**Behavioural impact:**
- No change in reconciliation order — all reconcile events are enqueued at the default priority (0 for normal events; -100 `LowPriority` for initial list-watch/resync events via `WithLowPriorityWhenUnchanged`). Processing order is unchanged.
- `workqueue_depth` gains an additional `priority` label (`"-100"` during initial sync, `"0"` for subsequent events). The other 6 metrics keep the existing `{name, controller}` label set.
- `UsePriorityQueue` is marked *beta* in controller-runtime but is the path forward — used in Kubebuilder scaffolding and controller-runtime's own examples.

### Minimum Agent Versions

N/A — operator-only change.

### Describe your test plan

1. Build and run the operator locally.
2. `k port-forward deploy/datadog-operator-manager 8080:8080` (adapt depending on your namespace, deployment name, etc.)
3. Verify `workqueue_` metrics are present

### Checklist

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -31,8 +31,10 @@ import (
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	"k8s.io/klog/v2"
 	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrlconfig "sigs.k8s.io/controller-runtime/pkg/config"
 	ctrlzap "sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/metrics/filters"
@@ -289,6 +291,16 @@ func run(opts *options) error {
 			DatadogDashboardEnabled:       opts.datadogDashboardEnabled,
 			DatadogGenericResourceEnabled: opts.datadogGenericResourceEnabled,
 		}),
+		// UsePriorityQueue makes all controllers use the priority queue, which
+		// directly registers workqueue metrics into controller-runtime's metrics
+		// registry (ctrlmetrics.Registry) rather than routing them through the
+		// global workqueue.SetProvider() call. This is necessary because
+		// k8s.io/kube-aggregator (transitively via k8s.io/component-base) wins
+		// the sync.Once in SetProvider, routing standard workqueue metrics to the
+		// k8s legacy registry instead of controller-runtime's registry.
+		Controller: ctrlconfig.Controller{
+			UsePriorityQueue: ptr.To(true),
+		},
 	})
 	if err != nil {
 		return setupErrorf(setupLog, err, "Unable to start manager")


khewonc:
Associated card for tracking: https://datadoghq.atlassian.net/browse/CONTP-1283

Issue #2791, opened 2026-03-20
Operator 1.24.0 crashloops on startup when cluster-wide secrets RBAC is missing (regression from #2530)


Paul-Weaver:
### Pre-submission Checklist

- [x] I have searched [existing issues](https://github.com/DataDog/datadog-operator/issues) and this is not a duplicate
- [x] This is a Datadog Operator issue (CRDs, reconciliation, etc.), not a Datadog Agent or Datadog service problem (dashboards, monitors, etc.)

### Operator version

1.24.0

### Operator Helm chart version

2.19.1

### Bug Report

Operator 1.24.0 (chart 2.19.x) crashloops on startup when the ServiceAccount
lacks cluster-wide `list`/`watch` on `secrets`.

We run `rbac.create: false` with a hand-managed ClusterRole scoped to only
what we need (Datadog CRDs, events, namespaces, leases, configmaps). We do
not grant cluster-wide secrets access by policy.

On 1.23.1 (chart 2.18.1) this was fine — the operator logged a warning
("Unable to get Secret informer, Helm metadata collection will be disabled"),
skipped the Helm metadata path, and started normally.

On 1.24.0 the same missing permission causes a cache sync timeout on
`*v1.Secret`, which kills `mgr.Start()` and the pod exits with code 1.
The behavior is nondeterministic — sometimes it starts, sometimes it doesn't,
depending on informer sync ordering/timing.

This appears to be a side effect of #2530 ("Remove early exit in
credential-dependent controllers when credentials are missing"), which
changed how the operator handles missing credentials/informers at startup.

Expected: operator starts and degrades gracefully (disables Helm metadata)
when secrets RBAC is absent, same as 1.23.1.

Actual: nondeterministic crash on startup due to Secret informer cache
sync timeout.

### Steps to Reproduce

1. Deploy chart 2.19.x with `rbac.create: false`
2. Create a ClusterRole that does NOT include `secrets` list/watch at cluster scope
3. Enable at least one controller (e.g. `datadogMonitor` + `datadogDashboard`)
4. Set `operatorMetricsEnabled: "true"` (default)
5. Observe pod — it will either start with forbidden errors or crashloop
   with "failed waiting for *v1.Secret Informer to sync" / exit code 1

### Environment

- Operator: 1.24.0 (works on 1.23.1)
- Chart: datadog-operator 2.19.1
- Kubernetes: 1.29
- Helm: 3.x

### Additional Context

Workarounds:
[...]

PR #2793, opened 2026-03-20
Avoid RBAC errors when Operator can't list or watch Secrets

levan-m:
### What does this PR do?

Fixes bug reported in https://github.com/DataDog/datadog-operator/issues/2791. 
* Don't start secrets watch in helm metadata forwarder if Operator doesn't have list/watch RBAC.
* Update CredentialManager to avoid unnecessary secrets informer initialization.

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

1. install Operator
```sh
❯ helm install operator datadog/datadog-operator \
--set datadogAgent.enabled=false \
--set datadogAgentInternal.enabled=false
```

2. Update rbac to drop secret watch, list permissions, restart operator pod.
```sh
  kubectl get clusterrole operator-datadog-operator -o json \
    | jq '
        .rules |= map(
          if (.resources // [] | index("secrets")) != null
          then .resources |= map(select(. != "secrets"))
          else .
          end
        )
        | .rules += [{"apiGroups":[""],"resources":["secrets"],"verbs":["create","delete","get","patch","update"]}]
      ' \
    | kubectl apply -f -

kubectl rollout restart deployment/operator-datadog-operator
```

4. Observer errors
`..."logger":"klog","msg":"Failed to watch","logger":"UnhandledError","reflector":"pkg/mod/k8s.io/client-go@v0.33.3/tools/cache/reflector.go:285","type":"*v1.Secret","error":"failed to list *v1.Secret: secrets is forbidden: User \"system:serviceaccount:default:operator-datadog-operator\" cannot list resource \"secrets\" in API group \"\" in`

5. Update to fixed image; after restart Operator should log
`{"level":"INFO","ts":"2026-03-20T19:23:55.561Z","logger":"metadata.helm","msg":"No permission to list/watch Secrets, Helm metadata collection from Secrets will be disabled"}`

### Checklist

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -309,7 +309,7 @@ func run(opts *options) error {
 
 	// Get call on a cached client initializes informer which requires list and watch permissions.
 	// If RBAC restricts list and watch permissions, the informer will log errors and may cause crash loops.
-	// Passing Reader avoids informer initialization.
+	// Reader interface as returned from mgr.GetAPIReader() reads directly from API server bypassing cache and informer initialization.
 	credsManager := config.NewCredentialManagerWithDecryptor(mgr.GetAPIReader(), secrets.NewSecretBackend())
 	creds, err := credsManager.GetCredentials()
 


cmd/main.go
```diff
@@ -307,8 +307,10 @@ func run(opts *options) error {
 
 	}
 
-	// Client is needed when Creds should be resolved from DDA so cached client is fine
-	credsManager := config.NewCredentialManagerWithDecryptor(mgr.GetClient(), secrets.NewSecretBackend())
+	// Get call on a cached client initializes informer which requires list and watch permissions.
+	// If RBAC restricts list and watch permissions, the informer will log errors and may cause crash loops.
+	// Passing Reader avoids informer initialization.
```
tbavelier: nit: This is not really true. What bypasses the initialization is the direct call using `mgr.GetAPIReader()` which returns a `client.Reader` instead of `client.Client` leading to these changes. Which is fine as we don't issue writes, only need read.

levan-m: updated comment.

PR #2796, opened 2026-03-23
[ACTP] Fix PAR ConfigMap name for profile DDAIs

merchristK:
## Description

When the PrivateActionRunner feature is enabled on a profile DDAI, the DaemonSets created for those profiles fail to start because the ConfigMap they reference is never created.

Root cause: `getConfigMapName()` and `getClusterAgentConfigMapName()` used `f.owner.GetName()` (the DDAI name, e.g. `compute-nodeless-200m-v2`) to build the ConfigMap name. Only the default DDAI runs `ManageDependencies` and creates the ConfigMap, so profile DDAIs end up referencing a ConfigMap that does not exist.

Fix: use `constants.GetDDAName(f.owner)` instead, which reads the `DatadogAgentNameLabelKey` label set on every DDAI at creation time and returns the parent DDA name. This means all DDAIs (default + profile) converge on the same ConfigMap name, which the default DDAI creates once.

## Changes

- Use `constants.GetDDAName()` in `getConfigMapName()` and `getClusterAgentConfigMapName()` so the name is derived from the parent DDA, not the DDAI
- Add `Test_privateActionRunnerFeature_ProfileDDAI_ConfigMapNames` to explicitly cover the profile DDAI case and guard against regression

## Testing

- New unit test verifies that a profile DDAI (with `DatadogAgentNameLabelKey` label pointing to parent DDA) produces ConfigMap names based on the DDA name, not the profile DDAI name
- All existing PAR tests continue to pass

Tested locally as well with a 2 nodes kind cluster with these 2 profiles:
```yaml
kind: DatadogAgentProfile
metadata:
  name: test-profile
spec:
  profileAffinity:
    profileNodeAffinity:
      - key: kubernetes.io/hostname
        operator: In
        values:
          - desktop-worker
  config:
    override:
      nodeAgent:
        containers:
          agent:
            resources:
              requests:
                cpu: 200m
---
kind: DatadogAgentProfile
metadata:
  name: test-profile-2
spec:
  profileAffinity:
    profileNodeAffinity:
      - key: kubernetes.io/hostname
        operator: In
        values:
          - desktop-control-plane
  config:
    override:
      nodeAgent:
        containers:
          agent:
            resources:
              requests:
                cpu: 300m
```

### DIFF
--- a/internal/controller/datadogagent/feature/privateactionrunner/feature.go
+++ b/internal/controller/datadogagent/feature/privateactionrunner/feature.go
@@ -206,11 +206,11 @@ func (f *privateActionRunnerFeature) ManageDependencies(managers feature.Resourc
 }
 
 func (f *privateActionRunnerFeature) getConfigMapName() string {
-	return fmt.Sprintf("%s-privateactionrunner", f.owner.GetName())
+	return fmt.Sprintf("%s-privateactionrunner", constants.GetDDAName(f.owner))
 }
 
 func (f *privateActionRunnerFeature) getClusterAgentConfigMapName() string {
-	return fmt.Sprintf("%s-clusteragent-privateactionrunner", f.owner.GetName())
+	return fmt.Sprintf("%s-clusteragent-privateactionrunner", constants.GetDDAName(f.owner))
 }
 
 func (f *privateActionRunnerFeature) getRbacResourcesName() string {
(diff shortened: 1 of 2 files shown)


merchristK:
* **#2796** <a href="https://app.graphite.com/github/pr/DataDog/datadog-operator/2796?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/datadog-operator/2796?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

PR #2805, opened 2026-03-24
Fix DatadogMonitor metrics forwarder goroutine leak on CR deletion

### LINKED ISSUE #2803: [BUG] DatadogMonitor metrics forwarder goroutines not unregistered on CR deletion
### Pre-submission Checklist

### Operator version

1.23.1

2.18.1

### Bug Report

We run the Datadog Operator at scale on EKS and have been load-testing its behavior with large numbers of CRs. During this work we found what appears to be a goroutine leak in the `DatadogMonitor` controller's metrics forwarder lifecycle. We wanted to share our findings in case they are useful.

**TL;DR:** The `DatadogMonitor` controller registers a metrics forwarder goroutine for every CR via `metricForwardersMgr.Register()` but does not appear to call `Unregister()` during finalization. This appears to leak one goroutine per CR that persists after deletion. With 1,000 `DatadogMonitor` CRs, the operator accumulates ~1,000 orphaned goroutines that are not reclaimed — eventually breaching the `maximumGoroutines` health check and triggering liveness probe failures. We noticed that the `DatadogAgent` finalizer calls `Unregister()`, but the `DatadogMonitor` finalizer does not, which may be an oversight.

**What happened:**

During load testing with 1,000 `DatadogMonitor` CRs on a dedicated EKS 1.34 cluster, we observed the following goroutine lifecycle:

| Phase | `go_goroutines` | Expected |
|---|---|---|
| Startup (no CRs) | ~30 | ~30 |
| Peak (1,000 monitors reconciling) | ~1,279 | Elevated during reconciliation |
| After all CRs deleted + 20 min idle | **~1,078** | **Should return to ~30-50** |
| New pod after restart | ~30 (until CRs recreated) | ~30 |

The ~1,048 goroutine delta (1,078 - 30) persists indefinitely and closely matches the number of CRs created. The goroutines only reclaim on pod restart.

In our environment, the goroutine accumulation caused liveness probe failures with the following cycle:

1. Goroutines exceed `maximumGoroutines` threshold
2. `/healthz` returns HTTP 500
3. Kubelet kills the pod after `failureThreshold` consecutive failures
4. New pod starts, assumes leadership, re-reconciles the same CRs
5. Goroutines spike again — repeat

We hit this at 1,000 monitors with `maximumGoroutines: 1000`. Raising the limit works as a workaround but does not address the underlying goroutine growth.

**Verification:**

```bash
# Create 1000 DatadogMonitor CRs, wait for reconciliation, then delete all
$ kubectl delete datadogmonitors -n datadog-operator-load-test --all

# Wait 5+ minutes for API-side deletions to complete and HTTP connections to drain
# Then check goroutine count via the operator's Prometheus metrics:
$ kubectl port-forward deploy/datadog-operator 8383:8383
$ curl -s http://localhost:8383/metrics | grep go_goroutines
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 1078
```

Expected: `go_goroutines` should return to approximately 30-50 (baseline) within a few minutes of all CRs being deleted and HTTP idle connections draining (90s default `IdleConnTimeout`). A goroutine profile (`/debug/pprof/goroutine`) would definitively confirm whether the leaked goroutines are blocked in the forwarder's `start()` function — we are happy to capture one if it would help the investigation.

**Suspected Root Cause:**

The `DatadogMonitor` controller registers a metrics forwarder goroutine in [`SetupWithManager()`](https://github.com/DataDog/datadog-operator/blob/main/internal/controller/datadogmonitor_controller.go) via an event filter:

```go
// internal/controller/datadogmonitor_controller.go
CreateFunc: func(e event.CreateEvent) bool {
    metricForwardersMgr.Register(e.Object)  // spawns a goroutine via go f.metricsForwarders[id].start(&f.wg)
    return true
},
```

Each `Register()` call spawns a persistent goroutine in the `ForwardersManager`. However, the [`DatadogMonitor` finalizer](https://github.com/DataDog/datadog-operator/blob/main/internal/controller/datadogmonitor/finalizer.go) only handles Datadog API-side cleanup (deleting the remote monitor) — it **never calls `Unregister()`** on the metrics forwarder.

In contrast, the [`DatadogAgent` finalizer](https://github.com/DataDog/datadog-operator/blob/main/internal/controller/datadogagent/finalizer.go) **does call `r.forwarders.Unregister(obj)`** during finalization. Since the `ForwardersManager.Register()` / `Unregister()` API is the same generic interface used by all controllers, we suspect the missing `Unregister()` in the `DatadogMonitor` path may be an oversight rather than intentional — but we may be missing context on the design intent.

**Ruling out other explanations:**

- **Controller-runtime workers**: Fixed-count (set by `MaxConcurrentReconciles`), not per-CR instance
- **Informer goroutines**: Per-resource-type, not per-CR instance
- **HTTP transport**: Go's default `IdleConnTimeout` (90s) drains idle connections well within the 20+ minute observation window

The ~1:1 correlation between CR count and goroutine delta points to a per-CR goroutine, consistent with the metrics forwarder registration path.

**Proposed Solution:**

Add an `Unregister()` call to the `DatadogMonitor` finalizer, matching the existing pattern in the `DatadogAgent` and `DatadogAgentInternal` finalizers:

```go
// In datadogmonitor/finalizer.go, finalizeDatadogMonitor():
func (r *Reconciler) finalizeDatadogMonitor(logger logr.Logger, dm *datadoghqv1alpha1.DatadogMonitor) {
    // Unregister metrics forwarder BEFORE deleteMonitor to ensure cleanup
    // even if the API call fails
    if r.operatorMetricsEnabled {
        r.forwarders.Unregister(dm)
    }

    if dm.Status.Primary {
        err := deleteMonitor(r.datadogAuth, r.datadogClient, dm.Status.ID)
        // ...
    }
}
```

**Implementation note:** We placed `Unregister()` before `deleteMonitor()` because the current function has an early-return path when the monitor is primary and the API call fails — placing it after could skip cleanup and still leak the goroutine. `Unregister()` is idempotent (no-op with a debug log if the forwarder is already removed), so calling it first is safe.

We considered placing the `Unregister()` call in a `DeleteFunc` predicate (symmetric with the `CreateFunc` registration) but prefer the finalizer approach because it matches the established `DatadogAgent`/`DatadogAgentInternal` pattern, and side effects in predicates are generally an anti-pattern in controller-runtime.

**Secondary observation: finalizer error handling.** While reviewing this code path, we noticed that `finalizeDatadogMonitor` returns `void` and `handleFinalizer` always removes the finalizer regardless of whether `deleteMonitor()` succeeded. In contrast, the `DatadogAgent` finalizer returns an error and gates finalizer removal on success — meaning a `DatadogAgent` API deletion failure is retried, but a `DatadogMonitor` API deletion failure may leave the remote monitor in place without retrying. This is a separate concern from the goroutine leak. We are happy to file it as a distinct issue if it would be useful.

**Workaround:** Setting `collectOperatorMetrics: false` (or `--operatorMetricsEnabled=false`) prevents the metrics forwarder goroutines from being spawned, but disables operator-level metrics forwarding.

### Steps to Reproduce

1. Deploy the operator with `collectOperatorMetrics: true` (the default) and `maximumGoroutines: 4000` (high enough to avoid liveness failures)
2. Note baseline goroutine count: `curl -s http://localhost:8383/metrics | grep go_goroutines` (~30)
3. Create 100+ `DatadogMonitor` CRs
4. Wait for reconciliation to complete (all CRs have `status.id`)
5. Delete all CRs: `kubectl delete datadogmonitors --all -n <namespace>`
6. Wait 5 minutes for all finalizers, API deletions, and HTTP idle connections to complete
7. Check goroutine count again — observe it remains ~100+ above baseline, matching the number of CRs created
8. Restart the operator pod — goroutine count returns to baseline

### Environment

**Kubernetes version:** 1.34 (EKS)
**Helm version:** 3.x
**Cloud provider:** AWS
**controller-runtime version (in operator):** v0.20.4

### Additional Context

**Related issues:**

- [#666](https://github.com/DataDog/datadog-operator/issues/666) — "Operator has a 200 goroutine limit" — led to `maximumGoroutines` being made configurable
- [#2285](https://github.com/DataDog/datadog-operator/issues/2285) — "datadog-operator-manager keep restarting" — a Datadog engineer noted the goroutine limit is *"used as a stopgap on routine leaks"*
- [#2344](https://github.com/DataDog/datadog-operator/issues/2344) — "Operator times out while reconciling - CrashLoopBackOff" — resolved by raising `maximumGoroutines`

These prior issues suggest the goroutine leak described above may be a contributing factor for users at scale.

**Affected controllers:** `DatadogMonitor` (calls `Register()` without a corresponding `Unregister()` on deletion). `DatadogAgent` and `DatadogAgentInternal` are not affected (both correctly call `Unregister()`). `DatadogSLO` is not affected (does not use the `ForwardersManager`).

**Goroutine breakdown at ~1,078 steady state (after creating and deleting 1,000 DatadogMonitor CRs):** Of the ~1,078 remaining goroutines, approximately 30 are the runtime baseline (informers, watches, leader election, health probes) and ~48 appear attributable to HTTP transport and controller-runtime overhead. The remaining ~1,000 correlate 1:1 with CR count, consistent with the suspected metrics forwarder leak. We have not yet captured a pprof dump but are happy to provide one, along with any other data that would help the investigation. We are also happy to submit a PR for the proposed fix if the approach looks reasonable to the team.

majiayu000:
## Summary
- Add `r.forwarders.Unregister(dm)` call in `finalizeDatadogMonitor()` to stop the metrics forwarder goroutine when a DatadogMonitor CR is deleted
- Guarded by `r.operatorMetricsEnabled` check, matching the existing pattern in `finalizeDadV2()` for the DatadogAgent controller
- Placed before `deleteMonitor()` to ensure cleanup happens even if `deleteMonitor()` returns early on error

Fixes #2803

## Test plan
- [x] Existing `Test_handleFinalizer` tests pass (deletion case uses `operatorMetricsEnabled=false` by default)
- [x] `go vet ./...` passes
- [x] `go build ./internal/controller/datadogmonitor/...` passes
- [x] All datadogmonitor package tests pass

### DIFF
--- a/internal/controller/datadogmonitor/finalizer.go
+++ b/internal/controller/datadogmonitor/finalizer.go
@@ -52,6 +52,10 @@ func (r *Reconciler) handleFinalizer(logger logr.Logger, dm *datadoghqv1alpha1.D
 }
 
 func (r *Reconciler) finalizeDatadogMonitor(logger logr.Logger, dm *datadoghqv1alpha1.DatadogMonitor) {
+	if r.operatorMetricsEnabled {
+		r.forwarders.Unregister(dm)
+	}
+
 	if dm.Status.Primary {
 		err := deleteMonitor(r.datadogAuth, r.datadogClient, dm.Status.ID)
 		if err != nil {


tbavelier:
Thank you for the contribution, I was working on the same issue at the same time unfortunately! Closing in favour of https://github.com/DataDog/datadog-operator/pull/2804

PR #2807, opened 2026-03-24
Remove collector config from host profiler.

mackjmr:
### What does this PR do?

Remove collector config from host profiler, the config is now fully inferred from agent config.
### Motivation

OTAGENT-926

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Enable host profiler and override image:
```
  annotations:
    agent.datadoghq.com/host-profiler-enabled: "true"
    experimental.agent.datadoghq.com/image-override-config: '{"host-profiler":{"name":"datadog/ddot-ebpf-dev:mackjmr-hp-run-optional"}}'
```

Ensure that the host-profiler only has core config flag:
```
    Command:
      /opt/datadog-agent/embedded/bin/full-host-profiler
      run
      --core-config=/etc/datadog-agent/datadog.yaml
```

Ensure there are no error logs in host profiler.

Ensure you can see profiles in the org corresponding to API Key. EBPF profiles do not show by default, so you need to add filter pod_name:<agent_pod_name>

### Checklist


internal/controller/datadogagent/feature/hostprofiler/feature.go
```diff
@@ -62,23 +47,6 @@ func (o *hostProfilerFeature) ID() feature.IDType {
 func (o *hostProfilerFeature) Configure(dda metav1.Object, ddaSpec *v2alpha1.DatadogAgentSpec, _ *v2alpha1.RemoteConfigConfiguration) feature.RequiredComponents {
 	o.hostProfilerEnabled = featureutils.HasFeatureEnableAnnotation(dda, featureutils.EnableHostProfilerAnnotation)
 
-	o.owner = dda
```
mackjmr, line 65: I have small doubt about removing owner. In the feature this is only used below, so not needed, but not sure if this has any impact outside this feature ?

khewonc: It's fine to remove owner if it's not needed in the feature 👍

internal/controller/datadogagent/feature/hostprofiler/const.go
```diff
@@ -1,8 +1 @@
 package hostprofiler
-
```
khewonc, line 2: I think you can delete this file since there are no more constants defined in it

mackjmr: removed in last commit 👍

PR #2822, opened 2026-03-25
Remove NetworkStats sub-option from ServiceDiscovery feature

Yumasi:
### What does this PR do?

This PR removes the discovery.networkStats config option as this feature has been removed from the agent and has never been released.

### Motivation

Remove deprecated options, and remove unneeded volume mounts in the system-probe container.

### Additional Notes

### Minimum Agent Versions

No. Discovery is not yet available publicly.

### Describe your test plan

N/A

### Checklist

### DIFF
--- a/api/datadoghq/v2alpha1/datadogagent_types.go
+++ b/api/datadoghq/v2alpha1/datadogagent_types.go
@@ -647,7 +647,7 @@ type ServiceDiscoveryFeatureConfig struct {
 	// +optional
 	Enabled *bool `json:"enabled,omitempty"`
 
-	// DEPRECATED: NetworkStats is no longer configurable and will be ignored.
+	// DEPRECATED: NetworkStats is no longer configurable and will be ignored. Scheduled for removal in v1.28.
 	// +deprecated
 	// +optional
 	NetworkStats *ServiceDiscoveryNetworkStatsConfig `json:"networkStats,omitempty"`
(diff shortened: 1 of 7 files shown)


cmd/yaml-mapper/mapper/mapping_datadog_helm_to_datadogagent_crd.yaml
```diff
@@ -393,7 +393,6 @@ datadog.dd_url: spec.endpoint.url
 datadog.disableDefaultOsReleasePaths: ""
 datadog.disablePasswdMount: ""
 datadog.discovery.enabled: ""
-datadog.discovery.networkStats.enabled: ""
```
vitkyrka, line 396: What Codex said. If someone has explicitly set it in their values file it will error it now, is that correct/expected behavior?

docs/configuration_public.md
```diff
@@ -384,9 +384,6 @@ spec:
 `features.serviceDiscovery.enabled`
 : Enables the service discovery check. Default: false
 
-`features.serviceDiscovery.networkStats.enabled`
```
vitkyrka, line 387: If someone has a spec where they explicitly set this (to whatever value), it will fail to deploy with the new operator version, correct? Is that acceptable and expected behaviour? What are the backwards compatibility requirements for the operator?

Yumasi: Indeed. I am going to add back the config fields but as no-ops (and still remove them from the documentation).
> Is that acceptable and expected behaviour? What are the backwards compatibility requirements for the operator?

There is a `docs/deprecated_configs.md` file that lists deprecated configs, along with a version removal. I guess we could add a notice in here? But I am not sure of the relevance of doing so given we never ga'd discovery. I can ask container-platform for advice. :)

PR #2834, opened 2026-03-26
Only add the AppArmor annotation if the container actually exists

lebauce:
### What does this PR do?

Do not fail on AppArmor annotation when the security-agent container is missing.

### Motivation

When both CWS and CSPM are running in system-probe, the operator fails on trying
to add an annotation on a container that doesn't exist:

DaemonSet.apps        
\"compute-nodeless-adp-uds-v1-agent\" is invalid: spec.template.annotations[container.apparmor.security.beta.kubernetes.io/security-agent]: Invalid value:  
\"security-agent\": container not found

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist

### DIFF
--- a/internal/controller/datadogagent/override/container_test.go
+++ b/internal/controller/datadogagent/override/container_test.go
@@ -1145,7 +1145,7 @@ func TestContainer(t *testing.T) {
 				})
 			},
 			override: v2alpha1.DatadogAgentGenericContainer{
-				AppArmorProfileName: apiutils.NewStringPointer("my-app-armor-profile"),
+				AppArmorProfileName: ptr.To("my-app-armor-profile"),
 			},
 			validateManager: func(t *testing.T, manager *fake.PodTemplateManagers, containerName string) {
 				annotation := fmt.Sprintf("%s/%s", common.AppArmorAnnotationKey, apicommon.SecurityAgentContainerName)


internal/controller/datadogagent/override/container.go
```diff
@@ -234,13 +234,26 @@ func overrideSeccompProfile(containerName apicommon.AgentContainerName, manager
...
-		} else {
-			annotation = fmt.Sprintf("%s/%s", common.AppArmorAnnotationKey, containerName)
+			effectiveName = *override.Name
+		}
+
+		// Only add the AppArmor annotation if the container actually exists in the pod spec.
+		// This avoids invalid DaemonSet configurations when a container is not present
+		// (e.g. security-agent is absent when directSendFromSystemProbe is enabled).
+		containerExists := false
+		for _, c := range manager.PodTemplateSpec().Spec.Containers {
```
levan-m: do initiContainers need the annotation too? `containerName`  could be one of initContainers as well.

internal/controller/datadogagent/override/container_test.go
```diff
@@ -1133,6 +1133,26 @@ func TestContainer(t *testing.T) {
...
+			existingManager: func() *fake.PodTemplateManagers {
+				// Pod spec does not contain the security-agent container
+				return fake.NewPodTemplateManagers(t, corev1.PodTemplateSpec{
+					Spec: corev1.PodSpec{
+						Containers: []corev1.Container{*agentContainer},
+					},
+				})
+			},
+			override: v2alpha1.DatadogAgentGenericContainer{
+				AppArmorProfileName: apiutils.NewStringPointer("my-app-armor-profile"),
```
levan-m:
```suggestion
				AppArmorProfileName: ptr.To("my-app-armor-profile"),
```

PR #2838, opened 2026-03-26
Add DDA rollback functionality

khewonc:
### What does this PR do?

Adds rollback functionality for DDAs with fleet experiments. Includes:
* stop: trigger rollback (for eventual stopExperiment)
* timeout: rollback after 15min of a running experiment
* abort: user makes a manual change (ignores manual change if done at the same time as the timeout b/c of complexity)

### Motivation

https://datadoghq.atlassian.net/browse/CONTP-1404

### Additional Notes

Anything else we should know when reviewing?

### Minimum Agent Versions

Are there minimum versions of the Datadog Agent and/or Cluster Agent required?

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Setup:
* Deploy the operator with `createControllerRevisions` and `datadogAgentInternalEnabled` set to true: https://github.com/DataDog/datadog-operator/blob/62fdb0618c3a62289f84e936b813b2789d674c21/cmd/main.go#L188-L189
* Create a DDA

Stop signal:
* Start an experiment (see below for how to mock)
* Mock stopping the experiment by patching the status:
```shell
kubectl patch dd <name> --type=merge --subresource status --patch 'status: {experiment: {phase: stopped}}'
```
* Check that the DDA spec is rolled back to the initial pre-experiment state
* Check that the experiment phase is `rollback` (`kubectl describe dd <name>`)

Timeout
* Start an experiment (see below for how to mock)
* Wait 15 minutes (not configurable, sorry)
* Check that the DDA spec is rolled back to the initial pre-experiment state
* Check that the experiment phase is `timeout` (`kubectl describe dd <name>`)

Abort
* Start an experiment (see below for how to mock)
* Patch the spec again
```shell
kubectl patch datadogagent <name> -n <namespace> --type merge -p '{"spec":{"global":{"tags":["experiment:true","manual-change:true"]}}}'
```
* Check that the experiment phase is `aborted` (`kubectl describe dd <name>`)
* Check that the DDA spec matches the change you just made (no rollback to pre-experiment or experiment spec)

Start an experiment (mock the start experiment signal):
```shell
# patch dda spec
# note: this can be any spec change
kubectl patch datadogagent <name> -n <namespace> --type merge -p '{"spec":{"global":{"tags":["experiment:true"]}}}'
# patch dda status
# generation should match the current (post spec patch) dda's generation
[...]

PR #2872, opened 2026-04-07
Add experiment signals to fleet remote config

khewonc:
### What does this PR do?

Adds experiment signals: start, stop, and promote. This depends on https://github.com/DataDog/datadog-operator/pull/2838. The latest commit contains the relevant changes

### Motivation

https://datadoghq.atlassian.net/browse/CONTP-1424
https://datadoghq.atlassian.net/browse/CONTP-1425
https://datadoghq.atlassian.net/browse/CONTP-1426

### Additional Notes

Merge after https://github.com/DataDog/datadog-operator/pull/2838

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

TBD

### Checklist


internal/controller/datadogagent/experiment.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+	if bytes.Equal(currentSnap, cr.Data.Raw) {
+		ctrl.LoggerFrom(ctx).Info("Rollback spec already matches target, skipping update", "rollbackTarget", rollbackTarget)
+		return nil
+	}
+
+	// Merge snapshot annotations (Datadog-only keys) on top of current annotations
+	// so that non-Datadog annotations (user metadata, tooling labels, etc.) are preserved.
+	merged := maps.Clone(current.Annotations)
+	maps.Copy(merged, snapshot.Annotations)
```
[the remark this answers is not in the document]
khewonc: e686774a017d7e03a9e892bfdfc97d030ab3e598

internal/controller/datadogagent/controller_reconcile_v2.go
```diff
@@ -71,11 +71,17 @@ func (r *Reconciler) reconcileInstanceV3(ctx context.Context, logger logr.Logger
 	ddaStatusCopy := instance.Status.DeepCopy()
 	newDDAStatus := generateNewStatusFromDDA(ddaStatusCopy)
 
-	// Manage ControllerRevision snapshots
 	if r.options.CreateControllerRevisions {
```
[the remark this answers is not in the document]
khewonc: Controller revisions are required for experiments so the feature gate is fine. To make this more obvious, added another gate for all experiment signals to require `createControllerRevisions` and `datadogAgentInternalEnabled`: e686774a017d7e03a9e892bfdfc97d030ab3e598

pkg/fleet/daemon.go
```diff
@@ -88,40 +126,365 @@ func (d *Daemon) getConfig(id string) (installerConfig, error) {
...
+			d.setPackageConfigVersions(req.Package, stable, req.Params.Version)
+			return nil
+		}
 		return fmt.Errorf("start DatadogAgent experiment: %w", err)
 	}
-	d.logger.Info("Starting DatadogAgent experiment", "id", req.ID, "package", req.Package, "config_id", cfg.ID)
+
+	// Apply the spec patch.
+	if err := retryWithBackoff(ctx, func() error {
+		return d.client.Patch(ctx, dda, client.RawPatch(types.MergePatchType, op.Config))
```
coignetp, line 248: **❓ question:** ‏Is this a blocking call? Else, any UPDATER_TASK instruction will block any further UPDATER_TASK until the first one is done.
This means we can't `stop` a change from the backend

pkg/fleet/daemon.go
```diff
@@ -88,40 +126,365 @@ func (d *Daemon) getConfig(id string) (installerConfig, error) {
...
+			})
+			found = true
+		} else {
+			updated = append(updated, pkg)
+		}
+	}
+	if !found {
+		updated = append(updated, &pbgo.PackageState{
+			Package:                 pkgName,
+			StableVersion:           stable,
```
coignetp: **🔨 warning:** ‏We shouldn't update `stableVersion` and `experimentVersion`, only `StableConfigVersion` and `ExperimentConfigVersion`, since we're not upgrading the operator

pkg/fleet/daemon.go
```diff
@@ -88,40 +126,365 @@ func (d *Daemon) getConfig(id string) (installerConfig, error) {
...
+		return
+	}
+	current := d.rcClient.GetInstallerState()
+	updated := make([]*pbgo.PackageState, 0, len(current)+1)
+	found := false
+	for _, pkg := range current {
+		if pkg.GetPackage() == pkgName {
+			updated = append(updated, &pbgo.PackageState{
+				Package:                 pkg.GetPackage(),
+				StableVersion:           stable,
```
coignetp: **🔨 warning:** ‏We shouldn't update `stableVersion` and `experimentVersion`, only `StableConfigVersion` and `ExperimentConfigVersion`, since we're not upgrading the operator

pkg/fleet/remote_config.go
coignetp, line 132: **❓ question:** ‏Should it be before the `h(req)`?

pkg/fleet/experiment.go
```diff
@@ -0,0 +1,118 @@
...
+	Cap:      10 * time.Second,
+	Steps:    math.MaxInt32,
+}
+
+// retryWithBackoff retries fn on any error with exponential backoff.
+// The total retry window is bounded by a 3-minute context timeout.
+func retryWithBackoff(ctx context.Context, fn func() error) error {
+	ctx, cancel := context.WithTimeout(ctx, 3*time.Minute)
+	defer cancel()
+	return retry.OnError(experimentBackoff, func(err error) bool {
```
coignetp, line 125: **🥜 nitpick:** ‏We can probably only retry on specific errors

pkg/fleet/daemon.go
```diff
@@ -88,40 +126,365 @@ func (d *Daemon) getConfig(id string) (installerConfig, error) {
...
+	} else if isNoOp {
+		return nil
+	}
+
+	// Update status: set phase=stopped; leave ID and generation for the reconciler.
+	if err := retryWithBackoff(ctx, func() error {
+		if err := d.client.Get(ctx, nsn, dda); err != nil {
+			return err
+		}
+		dda.Status.Experiment.Phase = v2alpha1.ExperimentPhaseStopped
```
coignetp, line 305: **🥜 nitpick:** ‏Do we need to check for `nil` here?

pkg/fleet/daemon.go
```diff
@@ -88,40 +126,365 @@ func (d *Daemon) getConfig(id string) (installerConfig, error) {
...
+		return nil
 	}
-	d.logger.Info("Promoting DatadogAgent experiment", "id", req.ID, "package", req.Package, "config_id", cfg.ID)
+
+	// Update status: set phase=promoted; leave ID and generation intact.
+	if err := retryWithBackoff(ctx, func() error {
+		if err := d.client.Get(ctx, nsn, dda); err != nil {
+			return err
+		}
+		dda.Status.Experiment.Phase = v2alpha1.ExperimentPhasePromoted
```
coignetp, line 354: **🥜 nitpick:** ‏Do we need to check for `nil` here?

pkg/fleet/daemon.go
khewonc: This is indeed a blocking call. Both spec patch and status update are needed to start the experiment before stoping it though. If we patch the spec, then cancel the start signal to stop, stopExperiment would see there's no active experiment (since the status wasn't updated yet) and fail. The patched spec would still be applied, but with no experiment to track it. If we cancel the start signal before patching, then there wouldn't be an experiment to stop either

pkg/fleet/remote_config.go
khewonc: I'm not entirely certain on this, but I think either would be fine

pkg/fleet/daemon.go
khewonc: Fixed in a816a2ac60bcafd4ac389e4e3e0ea854cb6d98b8

pkg/fleet/daemon.go
khewonc: a816a2ac60bcafd4ac389e4e3e0ea854cb6d98b8

pkg/fleet/experiment.go
khewonc: Added a function to check for known non-retryable errors: a816a2ac60bcafd4ac389e4e3e0ea854cb6d98b8

pkg/fleet/daemon.go
khewonc: Good call: a816a2ac60bcafd4ac389e4e3e0ea854cb6d98b8

pkg/fleet/daemon.go
khewonc: a816a2ac60bcafd4ac389e4e3e0ea854cb6d98b8

coignetp (approved):
Looks good for fleet, let's iterate on this

levan-m (approved):
Approving since this is gated behind configs and disabled by default.
I have a reservation about `daemon.go` doing status updates increases ways things can go wrong, will discuss offline.

PR #2879, opened 2026-04-08
[CONTP-1335] organize finalizers

nlchung:
### What does this PR do?

Migrates DatadogAgent, DatadogAgentInternal, DatadogMonitor, DatadogDashboard, and DatadogGenericResource to the shared `internal/controller/finalizer/` package, removing ~40–60 lines of duplicated finalizer scaffolding per controller. (DatadogSLO already used it.) Each controller now exposes a `deleteResource()` callback wrapping its existing cleanup logic.

### Motivation

[CONTP-1335](https://datadoghq.atlassian.net/browse/CONTP-1335): consolidate the six copy-pasted finalizer implementations into one.

### Notes for reviewers

The six controllers weren't actually doing the same thing — three small differences had to be reconciled before they could share code:

1. **What to return after adding the finalizer.** Agent, DDAI, Dashboard, and GenericResource returned `Requeue: true` (reconcile again right away). Monitor returned `RequeueAfter: 60s`. SLO didn't return at all — it kept running the body of the reconcile. We standardized on `Requeue: true`. An earlier iteration tried `RequeueAfter: 60s` for everyone and it caused 34 integration-test timeouts: the tests wait up to 10s for the controller to do real work, but with a 60s delay the controller just sat there.
2. **What to do if the Datadog API delete fails.** Dashboard and GenericResource log and move on (remove the finalizer anyway). Monitor and SLO keep the finalizer and retry. We kept each controller's existing behavior — the `ResourceDeleteFunc` type has a doc comment explaining that returning an error blocks deletion and retries, while returning nil lets deletion proceed.
3. **The Datadog resource ID.** Dashboard and SLO have it as a string. GenericResource calls the field `Id` instead of `ID`. Monitor stores it as an int. Agent and DDAI don't have one. The shared callback takes `datadogID string`; callers convert whatever they have. Empty string means "nothing to delete in Datadog."

`NewFinalizer` takes two durations: `defaultRequeuePeriod` (how often to check in while Kubernetes is garbage-collecting the deleted object — 60s for most controllers, 15s for Agent and DDAI) and `defaultErrRequeuePeriod` (5s retry on `client.Update` or `deleteFunc` failures).

### Describe your test plan

- Shared `HandleFinalizer` unit tests cover add/no-op/delete-success/delete-failure paths.
- `make integration-tests` passes in ~15s, including the 34 specs that timed out in an earlier iteration.
- Deployed to minikube: DatadogAgent create + delete reconciles cleanly, nothing stuck in Terminating.

### Minimum Agent Versions

N/A — operator-side refactor only.

### Checklist

- [x] PR has at least one valid label: `refactoring`
- [x] All commits are signed


internal/controller/datadogagent/controller_reconcile_v2.go
```diff
@@ -40,7 +41,8 @@ func (r *Reconciler) internalReconcileV2(ctx context.Context, instance *datadogh
 	}
 
 	// 2. Handle finalizer logic.
-	if result, err := r.handleFinalizer(reqLogger, instance, r.finalizeDadV2); utils.ShouldReturn(result, err) {
+	final := finalizer.NewFinalizer(reqLogger, r.client, r.deleteResource(reqLogger), 0, 0)
```
Mathew-Estafanous: Why don't we pass in an error re-queue period like what's done for `datadogslo`? Same for all the other migrated controllers.

nlchung: good catch, am looking into this. thanks!

nlchung: fixed and should be good for a second review now :)

Mathew-Estafanous (approved):
looks good!

Issue #2886, opened 2026-04-09
[BUG]Operator v1.25.0 crashloops on missing cluster-wide ConfigMaps RBAC L same class of bug as #2791 but for ConfigMaps


Paul-Weaver:
### Pre-submission Checklist

- [x] I have searched [existing issues](https://github.com/DataDog/datadog-operator/issues) and this is not a duplicate
- [x] This is a Datadog Operator issue (CRDs, reconciliation, etc.), not a Datadog Agent or Datadog service problem (dashboards, monitors, etc.)

### Operator version

1.25.0

### Operator Helm chart version

2.21.0

### Bug Report

**What happened:**

After upgrading to v1.25.0 (which includes the fix for #2791), the operator no longer crashloops on missing Secrets RBAC — thanks for that fix. However, the same crashloop behavior now occurs due to ConfigMaps RBAC. The operator fails to sync the `*v1.ConfigMap` informer cache and exits, causing a CrashLoopBackOff:

`Failed to watch *v1.ConfigMap: configmaps is forbidden: User "system:serviceaccount:datadog-operator:datadog-operator" cannot list resource "configmaps" in API group "" at the cluster scope`

This is the exact same pattern as #2791 — the controller-runtime cache sync times out on a resource the ServiceAccount can't list at cluster scope, which kills `mgr.Start()` and the pod exits. The fix in #2793/#2800 addressed this for Secrets but not for ConfigMaps.

We run `rbac.create: false` with a hand-managed ClusterRole and deliberately do not grant cluster-wide `list`/`watch` on ConfigMaps. Granting blanket cluster-scope ConfigMap access is not something we want to do — ConfigMaps can contain sensitive data across all namespaces, and we follow least-privilege RBAC.

**What I expected:**

The operator should degrade gracefully when ConfigMap list/watch is missing at cluster scope — same treatment as #2793 gave to Secrets. Specifically it should either:

1. Scope its ConfigMap informer to specific namespaces it actually needs rather than defaulting to cluster-wide via controller-runtime's cache
2. Use the uncached `APIReader` for one-off ConfigMap reads so the cluster-wide informer isn't triggered
3. Log a warning and disable the dependent feature (e.g. Helm metadata collection), same as the operator did in v1.23.1 for Secrets

[...]

PR #2889, opened 2026-04-10
Avoid RBAC errors when Operator can't list or watch ConfigMaps

levan-m:
### What does this PR do?

reported in https://github.com/DataDog/datadog-operator/issues/2886

Fix similar to https://github.com/DataDog/datadog-operator/pull/2793 for `configmaps`
* Don't start CM watch in helm metadata forwarder if Operator doesn't have list/watch RBAC.

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

1. install Operator
```sh
❯ helm install operator datadog/datadog-operator \
--set datadogAgent.enabled=false \
--set datadogAgentInternal.enabled=false
```

2. Update rbac to drop secret watch, list permissions, restart operator pod.
```sh
kubectl get clusterrole operator-datadog-operator -o json \
    | jq '
        .rules |= map(
          if (.resources // [] | index("configmaps")) != null
          then .resources |= map(select(. != "configmaps"))
          else .
          end
        )
        | .rules += [{"apiGroups":[""],"resources":["configmaps"],"verbs":["create","delete","get","patch","update"]}]
      ' \
    | kubectl apply -f -

kubectl rollout restart deployment/operator-datadog-operator
```

4. Observer errors
{"level":"ERROR","ts":"2026-04-10T14:22:21.312Z","logger":"controller-runtime.cache.UnhandledError","msg":"Failed to watch","reflector":"pkg/mod/k8s.io/client-go@v0.35.1/tools/cache/reflector.go:289","type":"*v1.ConfigMap","error":"failed to list *v1.ConfigMap: configmaps is forbidden: User \"system:serviceaccount:default:operator-datadog-operator\" cannot list resource \"configmaps\" in API group \"\" in the namespace \"default\"","stacktrace":"k8s.io/apimachinery/pkg/util/runtime.logError\n\t/go/pkg/mod/k8s.io/apimachinery@v0.35.1/pkg/util/runtime/runtime.go:221\n

5. Update to fixed image; after restart Operator should log
{"level":"INFO","ts":"2026-04-10T14:18:38.100Z","logger":"metadata.helm","msg":"No permission to list/watch ConfigMaps, Helm metadata collection from ConfigMaps will be disabled"}

### Checklist


tbavelier (approved):
Approving with a small nit: there's a rare/unlikely possibility that RBAC passes but informer registration fails but the flag doesn't reflect that (same for the previous PR) for `processKey` (it's true despite informer registration failing), so while we still log we disable the collection, we actually do not change the flag. See suggestion below, it applies to both secrets and configmaps: https://github.com/DataDog/datadog-operator/pull/2894
Note in practice I'm not sure how this could happen, so this is mostly defensive / meant to showcase how a unit test could fail

PR #2897, opened 2026-04-13
Fix AppArmor annotations for absent containers in annotation overrides

lebauce:
…verrides

The fix in a0dc8c07 added a container existence check to overrideAppArmorProfile(), preventing invalid AppArmor annotations when a container (e.g. security-agent with directSendFromSystemProbe=true) is absent from the pod spec.

However, the same guard was missing from the direct annotation loop in PodTemplateSpec(), which blindly copies spec.override.nodeAgent.annotations to the pod template. Any AppArmor annotation set via that path would bypass the existing fix and still produce an invalid DaemonSet.

Apply the same container existence check when iterating override.Annotations: skip AppArmor annotations (container.apparmor.security.beta.kubernetes.io/<name>) if <name> does not match any container in the pod spec.

### What does this PR do?

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist

### DIFF
--- a/internal/controller/datadogagent/override/container.go
+++ b/internal/controller/datadogagent/override/container.go
@@ -242,15 +242,7 @@ func overrideAppArmorProfile(containerName apicommon.AgentContainerName, manager
 		// Only add the AppArmor annotation if the container actually exists in the pod spec.
 		// This avoids invalid DaemonSet configurations when a container is not present
 		// (e.g. security-agent is absent when directSendFromSystemProbe is enabled).
-		containerExists := false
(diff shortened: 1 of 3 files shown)


tbavelier:
Since we're duplicating the logic, should we use a shared helper https://github.com/DataDog/datadog-operator/pull/2898 ?  Also uses the opportunity to use slices.ContainsFunc instead of looping

lebauce:
@tbavelier Thanks. I cherry-pick your commit

tbavelier (approved):

PR #2902, opened 2026-04-14
feat(DPA): Add options.burstable +  CPURequestsRemoveLimitsMemoryRequestsAndLimits controlled values

clamoriniere:
## Summary

- Adds a new `CPURequestsRemoveLimitsMemoryRequestsAndLimits` enum value to `DatadogPodAutoscalerContainerControlledValues`
- This mixed-mode value applies different strategies per resource type:
  - **CPU**: only requests are controlled and any existing CPU limit is removed, allowing the container to burst freely and avoid CPU throttling
  - **Memory**: both requests and limits are controlled
- Adds doc comments to all three enum values to clarify their respective behaviours
- Adds a new `burstable` option to `DatadogPodAutoscalerOptions` (spec) and `DatadogPodAutoscalerOptionsStatus` (status); when set, removes CPU limits from containers granting Burstable QoS. The Cluster Agent default is controlled by the setting `autoscaling.workload.options.burstable`
- Regenerates CRD manifests via `make generate`

## Test plan

- [ ] Verify the new enum value is accepted by the CRD validation
- [ ] Verify CPU limits are removed when this mode is active
- [ ] Verify memory requests and limits are both updated when this mode is active
- [ ] Verify `burstable: true` in spec removes CPU limits and sets Burstable QoS
- [ ] Verify `burstable` unset in spec falls back to the Cluster Agent setting `autoscaling.workload.options.burstable`

🤖 Assisted-by: Claude:claude-sonnet-4-6 sparse smatch


config/crd/bases/v1/datadoghq.com_datadogpodautoscalers.yaml
```diff
@@ -104,6 +104,7 @@ spec:
                             enum:
                               - RequestsAndLimits
                               - RequestsOnly
+                              - CPURequestsRemoveLimitsMemoryRequestsAndLimits
```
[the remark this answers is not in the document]
clamoriniere: no need, the bundle files will be updated during the operator release process

gabedos (approved):

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -359,12 +359,21 @@ type DatadogPodAutoscalerConstraints struct {
...
-	DatadogPodAutoscalerContainerControlledValuesRequestsOnly      DatadogPodAutoscalerContainerControlledValues = "RequestsOnly"
+
+	// DatadogPodAutoscalerContainerControlledValuesRequestsOnly controls only requests for all resources (CPU and Memory).
+	// Existing limits are left unchanged.
+	DatadogPodAutoscalerContainerControlledValuesRequestsOnly DatadogPodAutoscalerContainerControlledValues = "RequestsOnly"
+
+	// DatadogPodAutoscalerContainerControlledValuesCPURequestsRemoveLimitsMemoryRequestsAndLimits applies different strategies per resource:
+	// - CPU: only requests are controlled and any existing CPU limit is removed, allowing the container to burst freely and avoid CPU throttling.
+	// - Memory: both requests and limits are controlled.
+	DatadogPodAutoscalerContainerControlledValuesCPURequestsRemoveLimitsMemoryRequestsAndLimits DatadogPodAutoscalerContainerControlledValues = "CPURequestsRemoveLimitsMemoryRequestsAndLimits"
```
AlexanderYastrebov, line 376: nit: Here I'm concerned about the "remove" verb. Is there a difference between "not controlled" limit and "removed" limit? If not, then maybe just list all the controlled resources explicitly like `CPURequestMemoryRequestMemoryLimit`?

AlexanderYastrebov (approved):

api/datadoghq/common/datadogpodautoscaler_types.go
clamoriniere: I would keep "remove", because we stop controlling the limit (not setting it) but also if it exist we remove it. That is why I think "remove" fit more the behavior.
Let me know if you agree.

AlexanderYastrebov: It is ok as long as we have clear doc that explains the behaviour, which we do.

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -419,6 +428,12 @@ type DatadogPodAutoscalerOptions struct {
 	// OutOfMemory configures behavior when OOM events are detected.
 	// +optional
 	OutOfMemory *DatadogPodAutoscalerOutOfMemoryOptions `json:"outOfMemory,omitempty"`
+
+	// Burstable, if true, removes CPU limits from containers while keeping CPU request recommendations,
+	// granting the pod a Burstable QoS class and allowing it to consume idle node CPU capacity beyond its requests.
+	// If not set, the default value is determined by the Cluster Agent setting autoscaling.workload.options.burstable.
+	// +optional
+	Burstable *bool `json:"burstable,omitempty"`
```
vboulineau, line 436: As I already shared, I don't think this option should be exposed, I believe both the naming and the duplication is causing an issue.

api/datadoghq/common/datadogpodautoscaler_types.go
```diff
@@ -431,6 +446,16 @@ type DatadogPodAutoscalerOutOfMemoryOptions struct {
...
 }
 
+// DatadogPodAutoscalerOptionsStatus reflects the effective options applied by the autoscaler.
+// +kubebuilder:object:generate=true
+type DatadogPodAutoscalerOptionsStatus struct {
+	// Burstable is the effective value of the burstable setting applied by the autoscaler.
+	// When not set in the spec, this reflects the default determined by the Cluster Agent
+	// setting autoscaling.workload.options.burstable.
+	// +optional
+	Burstable *bool `json:"burstable,omitempty"`
```
vboulineau, line 456: Given that it can be set per-container, I don't believe the option makes sense at all.

PR #2930, opened 2026-04-24
fix(experiment): propagate fresh ResourceVersion after spec rollback

tbavelier:
### What does this PR do?

Persists the terminal experiment phase (`timeout`/`rollback`) in the same reconcile as the spec rollback, via a targeted `Status().Patch()` guarded by a JSON patch `test` op on `status.experiment.phase`. This eliminates a cross-reconcile race where the terminal phase could be silently dropped, without compromising concurrent-write safety or crash-recovery of the rollback annotation.

### Motivation

`Test_Experiment_StateTransitions` subtests (`rollback_then_timeout`, `promoted_then_timeout`, with/without `stale_old_revision`) were flaking on CI — failing with `expected=timeout, actual=running` at `controller_experiment_integration_test.go:947`. Reproduced locally (5+ failures in 20 runs).

Root cause: reconcile 1 of the timeout path updated spec (bumping `ResourceVersion`) but the full-status write used the stale RV → 409 → `phase=running` persisted. On reconcile 2, the refetched spec already matched a freshly-timestamped, rollback-annotated revision, so `findMostRecentMatchingRevision` returned it, `elapsed < timeout`, and the terminal phase was never written.

### Fix

`restorePreviousSpec` now writes the terminal phase via a targeted status subresource JSON patch (`patchExperimentPhase`) after the spec rollback. The patch is narrow (only `status.experiment.phase`) and carries a `test` op that asserts the server's phase still matches the phase the rollback decision was based on. Four safety properties are preserved:

1. **Lands in the same reconcile**: the targeted patch bypasses the RV bump from the spec Update, so `phase=timeout`/`phase=rollback` persists without waiting for a follow-up reconcile.
2. **Concurrent phase writes survive**: if the RC daemon (e.g. `pkg/fleet/daemon.go:354` writing `phase=promoted`) moves the phase between our reconcile fetch and our patch, the `test` op fails and the patch is rejected — our stale decision does not clobber their accepted state. Mirrors the 409-and-retry semantics of a full-status Update.
3. **Concurrent writes to other status fields survive**: the patch is applied to an `instance.DeepCopy()` scratch so the client's response decoding does not advance the caller's `instance.ResourceVersion`. The downstream full-status write in `updateStatusIfNeededV2` therefore retains its own 409 protection for every other status field.
4. **Crash-safe annotation ordering**: `annotateRevision` runs BEFORE `patchExperimentPhase`. If the operator exits between the two writes, the phase is still non-terminal on restart, `handleRollback` re-fires, `annotateRevision` is idempotent, and both writes eventually succeed. Reversing this order would leave a terminal phase without the rollback annotation — and since `handleRollback` ignores terminal phases, the annotation would never be applied, causing a later re-apply of the same spec to fire an immediate false timeout on the stale `CreationTimestamp`.

The patch is best-effort — any error (test op failure or transient) is logged at V(1) and the next reconcile re-computes and retries.

### Additional Notes

- `restorePreviousSpec` signature change: now takes `*v2alpha1.DatadogAgent` instead of `metav1.ObjectMeta` (needed for `Status().Patch()`).
- `rollback` signature unchanged (`metav1.ObjectMeta`).
- `TestRestorePreviousSpec_PhaseSetOnlyOnSuccess` was updated to populate `instance.Status.Experiment` (the new observed-phase capture in `restorePreviousSpec` dereferences it — safe in production because `handleRollback` guards against nil before delegating, but needed for this direct unit-test call).
- Regression guards added:
  - `Test_Experiment_RollbackPersistsInSingleReconcile` — asserts spec rollback and `phase=timeout` both persist in a single reconcile.
  - `Test_Experiment_RollbackPreservesConcurrentStatusWrite` — asserts `restorePreviousSpec` does not mutate the caller's `ResourceVersion` (bisected: reverting the scratch-copy fails this test).
  - `Test_Experiment_RollbackDoesNotClobberConcurrentPhaseWrite` — asserts a concurrent `phase=promoted` write survives a stale `phase=rollback` decision, AND that the rollback annotation still lands even when the phase patch is rejected (bisected: reverting to an unconditional merge patch fails the first assertion).

### Minimum Agent Versions

* Agent: n/a
* Cluster Agent: n/a

### Describe your test plan

- Reproduced the flake: `go test ./internal/controller/datadogagent/ -run Test_Experiment_StateTransitions -count=20` — multiple subtest failures before fix.
- Validated after fix: `Test_Experiment_StateTransitions + Test_Experiment_RollbackPersistsInSingleReconcile + Test_Experiment_RollbackPreservesConcurrentStatusWrite + Test_Experiment_RollbackDoesNotClobberConcurrentPhaseWrite` `-count=20` — all green.
- Full experiment suite (`Test_Experiment_|TestManageExperiment|TestHandleRollback|TestRollback|TestRestorePreviousSpec|TestReapplySameSpecAfterRollback`) `-count=10` — clean.
- Full `internal/controller/datadogagent` package — clean.

### Checklist

- [x] PR has at least one valid label: `bug`
- [x] PR has a milestone: `v1.26.0` (also `qa/skip-qa` — test-infrastructure stability, no user-facing change)
- [x] All commits are signed

### DIFF
--- a/internal/controller/datadogagent/controller_experiment_integration_test.go
+++ b/internal/controller/datadogagent/controller_experiment_integration_test.go
@@ -182,6 +182,16 @@ func Test_Experiment_RollbackDoesNotClobberConcurrentPhaseWrite(t *testing.T) {
 	concurrent.Status.Experiment.Phase = v2alpha1.ExperimentPhasePromoted
 	assert.NoError(t, r.client.Status().Update(context.TODO(), concurrent))
 
+	// Identify the experiment revision for later annotation assertions.
+	var experimentRevName string
+	maxRev := int64(0)
+	for _, rev := range listOwnedRevisions(t, r.client, ns, uid) {
+		if rev.Revision > maxRev {
+			maxRev = rev.Revision
+			experimentRevName = rev.Name
+		}
+	}
+
 	newStatus := &v2alpha1.DatadogAgentStatus{Experiment: snapshot.Status.Experiment.DeepCopy()}
 	revisions := mustListRevisions(t, r, snapshot)
 	assert.NoError(t, r.restorePreviousSpec(context.TODO(), snapshot, newStatus, revisions, v2alpha1.ExperimentPhaseRollback))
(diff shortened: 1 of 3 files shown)


tbavelier:
@codex review

tbavelier:
@codex review

tbavelier:
@codex review

tbavelier:
@codex review

PR #2953, opened 2026-04-28
[PROF-14068] Remove privileges for host-profiler

theomagellan:
### What does this PR do?

This PR mirrors https://github.com/DataDog/helm-charts/pull/2586 for `datadog-operator`:
- removes `privileges: true` and replaces by list of capabilities
- adds support for apparmor profiles
- embeds seccomp profile
- adds host-profiler related FQDN to Agent's Cilium allow-list
  - `intake.profile.%s`: profiling intake
  -  `sourcemap-intake.%s`: symbol intake
  - `otlp.%s`: OTLP metrics intake

### Motivation

https://datadoghq.atlassian.net/browse/REVIEW-85?focusedCommentId=3201542

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Tested on a cluster with the host-profiler feature enabled via `agent.datadoghq.com/host-profiler-enabled: "true"` annotation on the DDA.

Profiles for both supported architectures can be found [here](https://ddstaging.datadoghq.com/profiling/explorer?query=%28host%3Ai-09516ab7ee1385240%20OR%20host%3Ai-07c1951615c0817e3%29&agg_m=%40prof_ebpf_cpu_cores&agg_m_source=base&agg_t=sum&color_by=total&extra_search_fields=%7B%22filters_query%22%3A%22%22%2C%22sample_type%22%3A%22cpu-time%22%7D&fromUser=true&my_code=enabled&refresh_mode=paused&viz=stream&from_ts=1778000737908&to_ts=1778004337908&live=false)
### Checklist

### DIFF
--- a/internal/controller/datadogagent/component/agent/default.go
+++ b/internal/controller/datadogagent/component/agent/default.go
@@ -79,11 +79,13 @@ func NewDefaultAgentPodTemplateSpec(dda metav1.Object, agentComponent feature.Re
 // Capabilities for the Host Profiler container
 func DefaultCapabilitiesForHostProfiler() []corev1.Capability {
 	return []corev1.Capability{
-		"SYS_ADMIN",
+		"BPF",
+		"PERFMON",
 		"SYS_PTRACE",
 		"SYS_RESOURCE",
 		"DAC_READ_SEARCH",
 		"SYSLOG",
+		"CHECKPOINT_RESTORE",
 	}
 }
 
--- a/internal/controller/datadogagent/component/agent/default.go
+++ b/internal/controller/datadogagent/component/agent/default.go
@@ -704,7 +704,8 @@ func hostProfilerContainer(dda metav1.Object) corev1.Container {
 		VolumeMounts: volumeMountsForOtelAgent(),
 		Ports:        []corev1.ContainerPort{},
 		SecurityContext: &corev1.SecurityContext{
-			ReadOnlyRootFilesystem: ptr.To(true),
+			ReadOnlyRootFilesystem:   ptr.To(true),
+			AllowPrivilegeEscalation: ptr.To(false),
 			SeccompProfile: &corev1.SeccompProfile{
 				Type:             corev1.SeccompProfileTypeLocalhost,
 				LocalhostProfile: ptr.To(common.HostProfilerSeccompProfileName),
--- a/internal/controller/datadogagent/component/agent/default.go
+++ b/internal/controller/datadogagent/component/agent/default.go
@@ -75,6 +75,18 @@ func NewDefaultAgentPodTemplateSpec(dda metav1.Object, agentComponent feature.Re
 	}
 }
 
+// DefaultCapabilitiesForHostProfiler returns the default Security Context
+// Capabilities for the Host Profiler container
+func DefaultCapabilitiesForHostProfiler() []corev1.Capability {
+	return []corev1.Capability{
+		"SYS_ADMIN",
+		"SYS_PTRACE",
+		"SYS_RESOURCE",
+		"DAC_READ_SEARCH",
+		"SYSLOG",
+	}
+}
+
 // DefaultCapabilitiesForSystemProbe returns the default Security Context
 // Capabilities for the System Probe container
 func DefaultCapabilitiesForSystemProbe() []corev1.Capability {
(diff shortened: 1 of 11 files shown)


theomagellan:
@codex review

internal/controller/datadogagent/feature/hostprofiler/feature.go
```diff
@@ -82,6 +83,12 @@ func (o *hostProfilerFeature) ManageNodeAgent(managers feature.PodTemplateManage
 	// Host PID
 	managers.PodTemplateSpec().Spec.HostPID = *ptr.To(true)
 
+	// Security context capabilities
+	managers.SecurityContext().AddCapabilitiesToContainer(agent.DefaultCapabilitiesForHostProfiler(), apicommon.HostProfiler)
+
+	// AppArmor annotation
+	managers.Annotation().AddAnnotation(common.AppArmorAnnotationKey+"/"+string(apicommon.HostProfiler), "unconfined")
```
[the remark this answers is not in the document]
theomagellan: The bug requires containerStrategy: single + host-profiler enabled simultaneously, which seems not to be a real use case since the host-profiler now needs its own security context (capabilities, seccomp, AppArmor) that cannot coexist with the core agent in a single container.

That said, the host profiler should be in `IsPrivileged()` for correctness: the container requires elevated capabilities (`BPF`, `PERFMON`, `SYS_PTRACE`, etc.) which also fixes this.
Done in [7847cf2](https://github.com/DataDog/datadog-operator/pull/2953/commits/7847cf22fa4d7df27124684e1e235ce1bb0c0eaf)

r1viollet, line 127: AppArmor: Do we want to expose a setting to override the "unconfined" ? I'm not familiar with how someone would override this.

internal/controller/datadogagent/feature/types.go
```diff
@@ -76,7 +76,7 @@ func (rc *RequiredComponent) IsConfigured() bool {
 // IsPrivileged checks whether component requires privileged access.
 func (rc *RequiredComponent) IsPrivileged() bool {
```
r1viollet, line 77: NIT, we are no longer privileged, so this could be slightly confusing 😄

theomagellan: What do we consider as "privileged?" afaik `system-probe` is also not privileged but runs with elevated capabilities so I was mirroring that. This was done in response to https://github.com/DataDog/datadog-operator/pull/2953#discussion_r3195604876 but I can revert if you feel strongly about it :)

internal/controller/datadogagent/feature/hostprofiler/feature.go
theomagellan: You can change it by changing the datadog agent's manifest and override the `container.apparmor.security.beta.kubernetes.io/host-profiler` annotation

As discussed on Thursday though, the profile provisioning is left to the user; neither the helm chart or the operator allow us to provision one automatically, that's why it's `unconfined` by default

internal/controller/datadogagent/component/agent/default.go
```diff
@@ -697,7 +697,8 @@ func hostProfilerContainer(dda metav1.Object) corev1.Container {
 		VolumeMounts: volumeMountsForOtelAgent(),
 		Ports:        []corev1.ContainerPort{},
 		SecurityContext: &corev1.SecurityContext{
-			ReadOnlyRootFilesystem: ptr.To(true),
+			ReadOnlyRootFilesystem:   ptr.To(true),
+			AllowPrivilegeEscalation: ptr.To(false),
```
theomagellan: @r1viollet regarding https://github.com/DataDog/helm-charts/pull/2586#issuecomment-4381582045 and why we the seccomp allowed for GID/UID operations among other unnecessary syscalls:

It turns out that leaving `AllowPrivilegeEscalation` empty was in effect setting it to true, and this would cause `runc` to apply the seccomp profile before totally finishing its setup, including switching users from root to unprivileged! Those syscalls then got through my first syscall audit and got added to the seccomp profile 😁 

`helm-charts` also has this issue, I will have to do a follow-up PR to both tighten seccomp allowlist and fix this.

I'm including the fix in this PR but will update the seccomp in a follow-up PR to keep the seccomp profile consistent between both operator and helm.

internal/controller/datadogagent/feature/types.go
r1viollet: Your change is good, I was just mentioning the name of the function
This is really minor.

r1viollet (approved):
LGTM

tbavelier (commented):
Not sure how feasible it is, but would it be possible to make all these changes solely inside the feature/hostprofiler code path instead of inside common + default ? I believe/understand you followed the pattern from system-probe but system-probe is in a slightly different position where it is used by multiple features while host-profiler is self-contained, so would be ideal if it's fully managed by the feature code path. It might not be technically possible however, would need to be investigated

theomagellan:
> Not sure how feasible it is, but would it be possible to make all these changes solely inside the feature/hostprofiler code path instead of inside common + default ? I believe/understand you followed the pattern from system-probe but system-probe is in a slightly different position where it is used by multiple features while host-profiler is self-contained, so would be ideal if it's fully managed by the feature code path. It might not be technically possible however, would need to be investigated

That's correct, I followed `system-probe` as a guide for the `host-profiler` 😅 
I took a shot at it in https://github.com/DataDog/datadog-operator/commit/b03042ccf2ea34ba3f32faf6d8d584d819ac3049 and deploying it on my test cluster tells me that everything seems right.
Thank you for your input! Re-requesting a review as soon as CI passes 🙇

tbavelier:
@codex review

internal/controller/datadogagent/feature/hostprofiler/feature_test.go
tbavelier, line 1: Nit but `slices` can probably be used here to simplify/rewrite some helpers/assertions. Does not matter tho

internal/controller/datadogagent/feature/hostprofiler/seccomp.go
```diff
@@ -0,0 +1,200 @@
...
+			]
+		}
+		`, syscalls),
+	}
+}
+
+func buildSeccompSetupInitContainer() corev1.Container {
+	return corev1.Container{
+		Name:  "host-profiler-seccomp-setup",
+		Image: images.GetLatestAgentImage(),
```
tbavelier: See https://github.com/DataDog/datadog-operator/pull/3030 but tl;dr, make sure to use the same image as the other containers to avoid pulling it twice. Can be reproduced in current state with `spec.global.registry: public.ecr.aws` (or something different from the default gcr.io), and init-container will still pull from gcr.io

tbavelier (changes requested):
slices comment not important, but the one about the init-container image is thus requesting changes

internal/controller/datadogagent/feature/hostprofiler/seccomp.go
theomagellan: Thank you so much for the draft PR! Took the liberty to cherry-pick the commit directly :)

tbavelier (approved):

PR #2952, opened 2026-04-28
[FA] Read name from UPDATER_TASK

coignetp:
Re-do of #2913.

### What does this PR do?

Read the target `DatadogAgent` namespace/name from the `UPDATER_TASK` params directly (via a new `parseTaskNSN` helper) for start, stop, and promote operations, instead of storing it in a `experimentTarget` field on the daemon struct.

### Motivation

Previously, stop and promote operations relied on `d.experimentTarget` being set by a prior start call. This stateful approach was fragile — if the daemon restarted between start and stop/promote, the target was lost. Carrying the name in every task payload makes each operation self-contained and idempotent.

### Additional Notes

- Removes the `experimentTarget types.NamespacedName` field from the daemon struct.
- `NamespacedName` is now a field of `experimentParams` (carried in the task itself).
- `validateOperation` tests for empty name/namespace are replaced by `parseTaskNSN` tests.

### Minimum Agent Versions

No minimum version requirements.

### Describe your test plan

Unit tests in `pkg/fleet/daemon_test.go` and `pkg/fleet/remote_config_test.go` are updated to pass `NamespacedName` inside `experimentParams`. New `TestParseTaskNSN_*` tests cover the valid, missing-name, and missing-namespace cases.

### Checklist

- [x] PR has at least one valid label: `bug`, `enhancement`, `refactoring`, `documentation`, `tooling`, and/or `dependencies`
- [x] PR has a milestone or the `qa/skip-qa` label
- [x] All commits are signed (see: [signing commits][1])

[1]: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

PR #2956, opened 2026-04-29
[FA] Serialize UPDATER_TASK execution and fix no-op config version clearing

coignetp:
## What does this PR do?

Ports three fixes from #2952 that are independent of the `parseTaskNSN` / namespace-from-task refactor.

### 1. Serialize UPDATER_TASK execution (`taskMu` + `handleTask`)

Adds a `taskMu sync.Mutex` to `Daemon` and extracts a `handleTask` method that holds the lock for the full lifecycle of one task (RUNNING → DONE/ERROR/INVALID_STATE). The inline block in `Start()` is replaced with `return d.handleTask(ctx, req)`.

**Why:** Multiple concurrent RC callbacks can race on `setTaskState`. This matches the datadog-agent's single-writer model.

### 2. No-op stop clears experiment config version

When `canStop` returns `isNoOp=true` (e.g. spec already rolled back due to timeout), the experiment config version is now cleared before returning.

**Why:** Without this, the backend could not issue new tasks after a timeout-triggered rollback.

### 3. No-op promote clears experiment config version

Same fix for `promoteDatadogAgentExperiment` when the experiment is already in a terminal state.

### Checklist

### DIFF
--- a/pkg/fleet/daemon.go
+++ b/pkg/fleet/daemon.go
@@ -48,6 +48,7 @@ type Daemon struct {
 	mu               sync.RWMutex
 	configs          map[string]installerConfig // keyed by config ID; replaced on each RC update
 	experimentTarget types.NamespacedName       // DDA targeted by the current experiment; set on startExperiment
+	taskMu           sync.Mutex                 // serializes UPDATER_TASK execution
 }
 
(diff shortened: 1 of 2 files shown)


Hitsuji-M (approved):

khewonc:
Incorporated in https://github.com/DataDog/datadog-operator/pull/2944. Closing in favor of that PR

PR #2960, opened 2026-04-29
Migrate `kubectl-datadog autoscaling cluster` to `client-go` `pager`

L3n41c:
### What does this PR do?

Migrates the `kubectl-datadog autoscaling cluster install/uninstall` commands from manual Kubernetes list pagination loops (`metav1.ListOptions{Limit, Continue}`) to `client-go`'s `pager.ListPager.EachListItem`. Replaces several `lo.SomeBy` / `lo.ContainsBy` calls with stdlib `slices.ContainsFunc` where the semantics match.

### Motivation

The previous code mixed manual pagination loops at three sites (Node and Deployment listings) with non-paginated `List()` calls at others. Moving to `pager.ListPager.EachListItem` gives a consistent high-level idiom and extends pagination to two sites that previously risked OOM on dense clusters: `detectClusterAutoscaler` (Deployments cluster-wide) and `listKarpenterNodes` (label-filtered Nodes).

In `GetNodesProperties`, the per-batch EC2 work was implicitly tied to the K8s page size. The migration decouples them via an explicit `ec2DescribeBatchSize = 100`, and stores only the Node fields downstream needs (`Labels`, `Taints`) in a small `pendingNode` struct — avoiding both deep `corev1.Node` copies and the pinning of pager page backing arrays through long-lived `*corev1.Node` references.

### Describe how to test/QA your changes

- `go test ./cmd/kubectl-datadog/autoscaling/cluster/...`
- `make lint`
- `make kubectl-datadog`
- Manual smoke test: `kubectl datadog autoscaling cluster install` and `uninstall` on an EKS cluster.

### Possible Drawbacks / Trade-offs

- `detectClusterAutoscaler` no longer warns when several Cluster Autoscaler Deployments coexist on the same cluster. The cluster-wide enumeration order is unspecified, so when multiple matches coexist (a configuration we don't expect in practice), which one wins is non-deterministic. We accept that to keep the scan short-circuited and avoid materialising every Deployment in memory.

### Additional Notes

None.

### DIFF
--- a/cmd/kubectl-datadog/autoscaling/cluster/common/clusterinfo/classify.go
+++ b/cmd/kubectl-datadog/autoscaling/cluster/common/clusterinfo/classify.go
@@ -2,10 +2,11 @@ package clusterinfo
 
 import (
 	"context"
+	"errors"
 	"fmt"
-	"log"
 	"maps"
 	"regexp"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


L3n41c:
@codex review

AlexanderYastrebov (approved):

PR #2964, opened 2026-05-01
[CONTP-1499] feat(RC): Support API key rotation in RC client and server

Mathew-Estafanous:
### What does this PR do?

Adds Remote Configuration support for API key rotation when secret refresh is enabled.

The RC updater now reads credentials from `CredentialManager` during setup and watches the refreshed credential cache for API key changes. When the cached API key changes, the updater recreates the RC service/client so the new key is used without restarting the operator

* Pass `CredentialManager` into RC setup instead of a startup credential snapshot.
* Add an RC credential watch routine that restarts the RC runtime when the cached API key changes.
* Keep `RemoteConfigUpdater` as a stable `RCClient` wrapper so Fleet and other consumers do not need to handle client replacement.
* Replay built-in and Fleet subscriptions onto each replacement RC client before polling starts.
* Preserve installer state and apply it to replacement clients before startup.

### Motivation

The upstream RC client copies the API key during construction and does not expose a mutable API key API. The existing secret refresh routine updates the operator credential cache, but the RC client would continue using the old key until process restart.

### Describe your test plan

* Added `updator_test.go` test suite

#### Manual QA

1. Create `datadog-secret` secret that's mounted as a volume
```
kubectl create secret generic datadog-secret \                                                                                     
--from-literal api-key="$DD_API_KEY" \
--from-literal app-key="$DD_APP_KEY" \
-n system
```

2. Deploy operator with secrets refresh and remote config in `manager.yaml`
```yaml
spec:
             "metrics": ["*"]
           }]
     spec:
+      volumes:
+      - name: secret-volume
+        secret:
+          secretName: datadog-secret
       containers:
       - command:
         - /manager
         args:
         - --enable-leader-election
         - --pprof
+        - --secretBackendCommand=/readsecret.sh
+        - -secretBackendArgs=/etc/secret-volume
+        - --secretRefreshInterval=20s
+        - --remoteConfigEnabled=true
+        - --remoteUpdatesEnabled=true
         image: controller:latest
         imagePullPolicy: IfNotPresent
         name: manager
@@ -61,6 +70,15 @@ spec:
               fieldPath: metadata.namespace
         - name: DD_TOOL_VERSION
           value: "datadog-operator"
+        - name: DD_API_KEY
+          value: ENC[api-key]
+        - name: DD_APP_KEY
+          value: ENC[app-key]
+        - name: DD_CLUSTER_NAME
+          value: "<YOUR_CLUSTER_NAME>"
+        volumeMounts:
+          - mountPath: /etc/secret-volume
+            name: secret-volume
         resources:
           limits:
             cpu: 100m
```

3. Rotate `datadog-secret` to another API/APP key
```
kubectl patch secret datadog-secret -n system -p "{\"stringData\":{\"api-key\":\"$DD_API_KEY\",\"app-key\":\"$DD_APP_KEY\"}}"
```

4. Confirm RC client rotated in logs.
Screenshot 2026-05-04 at 11 34 29 AM

### Checklist

### DIFF
--- a/pkg/remoteconfig/updater.go
+++ b/pkg/remoteconfig/updater.go
@@ -80,7 +80,7 @@ type RcServiceConfiguration struct {
 	clusterName       string
 	telemetryReporter service.RcTelemetryReporter
 	agentVersion      string
-	rcDatabaseDir     string
+	rcDatabaseFile    string
 }
 
 // RCClient is the interface for subscribing to RC product updates.
@@ -209,6 +209,7 @@ func (r *RemoteConfigUpdater) startRuntimeWithConfig(apiKey string, site string,
 
 	oldSvc := r.rcService
 	oldClient := r.rcClient
+	oldDBFile := r.serviceConf.rcDatabaseFile
 
 	newClient.SetInstallerState(r.installerState)
 
@@ -236,6 +237,11 @@ func (r *RemoteConfigUpdater) startRuntimeWithConfig(apiKey string, site string,
 	}
 	if oldClient != nil {
 		oldClient.Close()
+	}
+	if oldDBFile != "" {
+		if err := os.Remove(oldDBFile); err != nil && !os.IsNotExist(err) {
+			r.logger.Error(err, "Failed to remove old Remote Configuration database file", "file", oldDBFile)
+		}
 	}
 
 	return nil
@@ -269,8 +275,8 @@ func (r *RemoteConfigUpdater) newServiceConfiguration(apiKey, site, clusterName,
 		clusterName:       clusterName,
 		telemetryReporter: dummyTelemetryReporter{},
 		// TODO fix when other values accepted
-		agentVersion:  "7.50.0",
-		rcDatabaseDir: baseDir,
+		agentVersion:   "7.50.0",
+		rcDatabaseFile: filepath.Join(baseDir, fmt.Sprintf("remote-config-%s.db", uuid.New())),
 	}, nil
 }
 
@@ -284,7 +290,7 @@ func defaultRCFactory(conf RcServiceConfiguration) (rcService, rcRuntimeClient, 
 		conf.telemetryReporter,
 		conf.agentVersion,
 		service.WithAPIKey(conf.apiKey),
-		service.WithDatabaseFileName(filepath.Join(conf.rcDatabaseDir, fmt.Sprintf("remote-config-%s.db", uuid.New()))),
+		service.WithDatabaseFileName(conf.rcDatabaseFile),
 		service.WithDirectorRootOverride(conf.cfg.GetString("site"), conf.cfg.GetString("remote_configuration.director_root")),
 		service.WithConfigRootOverride(conf.cfg.GetString("site"), conf.cfg.GetString("remote_configuration.config_root")),
 	)
--- a/pkg/remoteconfig/updater.go
+++ b/pkg/remoteconfig/updater.go
@@ -56,11 +56,15 @@ type RemoteConfigUpdater struct {
 
 type rcRuntimeFactory func(conf RcServiceConfiguration) (rcService, rcRuntimeClient, error)
 
+// rcService abstracts the RC service for testability. In production this is *service.CoreAgentService.
 type rcService interface {
 	Start()
 	Stop() error
 }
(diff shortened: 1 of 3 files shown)


pkg/remoteconfig/updater.go
```diff
@@ -268,6 +318,76 @@ func getEndpoint(prefix, site string) string {
...
+
+	if err := r.startRuntime(creds.APIKey); err != nil {
+		return err
+	}
+	r.activeAPIKey = creds.APIKey
+	r.logger.Info("Remote Configuration credentials changed, runtime restarted")
+	return nil
+}
+
+func (r *RemoteConfigUpdater) StartCredentialWatchRoutine(credsManager *config.CredentialManager, interval time.Duration) {
```
Mathew-Estafanous, line 379: 💭 Not sure if should add some kind of shutdown mechanism here. `RemoteConfigUpdater` doesn't have a `Stop` and many other operator components don't appear to have one either.

pkg/remoteconfig/updater.go
```diff
@@ -41,11 +41,39 @@ const (
 
 type RemoteConfigUpdater struct {
 	kubeClient  kubeclient.Client
-	rcClient    *client.Client
-	rcService   *service.CoreAgentService
+	rcClient    rcRuntimeClient
+	rcService   rcService
```
Mathew-Estafanous, line 45: 1. `rcRuntimeClient` and `rcService` are inner, swappable interfaces mainly to enable mocking in tests.
2. `RemoteConfigUpdater` itself implements the `RCClient` which is the stable interface used by external callers. The interface remained unchanged because callers don't care that the client/service can swap.

pkg/remoteconfig/updater.go
```diff
@@ -135,98 +158,93 @@ type dummyTelemetryReporter struct{}
...
+	}
 
-	rcClient.Subscribe(string(state.ProductAgentConfig), r.agentConfigUpdateCallback)
+	newClient.Start()
+	r.logger.Info("Remote Configuration client started")
 
-	rcClient.Subscribe(string(state.ProductOrchestratorK8sCRDs), r.crdConfigUpdateCallback)
+	r.serviceConf = serviceConf
+	r.rcService = newService
+	r.rcClient = newClient
```
Mathew-Estafanous, line 224: I'm swapping to the new runtime **before** tearing down the old one so there's no window where RC is unavailable. Old runtime cleanup errors are non-fatal since the new runtime is already active.

pkg/remoteconfig/updater.go
```diff
@@ -268,6 +320,76 @@ func getEndpoint(prefix, site string) string {
 	return prefix + defaultSite
 }
 
+func (r *RemoteConfigUpdater) Subscribe(product string, fn func(update map[string]state.RawConfig, applyStateCallback func(string, state.ApplyStatus))) {
+	r.lifecycleMu.Lock()
+	defer r.lifecycleMu.Unlock()
+
+	subscription := rcSubscription{product: product, fn: fn}
+	r.subscriptions = append(r.subscriptions, subscription)
```
Mathew-Estafanous, line 328: Subscriptions are recorded before forwarding to the live client so replacement clients can replay subscriptions during rotation to prevent subscription loss.

pkg/remoteconfig/updater.go
```diff
@@ -268,6 +320,76 @@ func getEndpoint(prefix, site string) string {
...
+	}
+
+	r.lifecycleMu.Lock()
+	defer r.lifecycleMu.Unlock()
+
+	if creds.APIKey == r.activeAPIKey {
+		return nil
+	}
+
+	if err := r.startRuntime(creds.APIKey); err != nil {
```
Mathew-Estafanous, line 371: 💭 It's possible the new API key may not have the correct RC scope or is for a different org. Currently we're not doing any validation. **Should we add validation here?** I'm leaning towards no for the sake of simplicity and relying on customers to rotate fully authorized keys.

coignetp (approved):

pkg/remoteconfig/updater.go
coignetp: I'd say no, if the scope is wrong, there will be logs

PR #2984, opened 2026-05-06
feat: Add APM to DatadogInstrumentation CRD

betterengineering:
### What does this PR do?

This commit adds the schema required to utilize the `DatadogInstrumentation` CRD for APM. Relates to https://github.com/DataDog/datadog-operator/pull/2962

##### Example

```yaml
kind: DatadogInstrumentation
metadata:
  name: redis-check
  namespace: cache
spec:
  targetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: example-service
  config:
    apm:
      enabled: true
      ddTraceVersions:
        python: "3"
      ddTraceConfigs:
        - name: "DD_PROFILING_ENABLED"
          value: "true"
```

### Motivation

Two relevant RFCs:
* [RFC: Datadog Instrumentation CRD](https://docs.google.com/document/d/10vxye8x3pIB_rVwM57sz3h00N8uudfAyBskjG6emQkE/edit?usp=sharing)
* [DatadogInstrumentation CRD for Workload-Scoped Product Enablement](https://datadoghq.atlassian.net/wiki/spaces/CONTP/pages/6564659495/DatadogInstrumentation+CRD+for+Workload-Scoped+Product+Enablement)

### Additional Notes

All this adds is the schema. The consumer of this schema is the Datadog Cluster Agent.

### Describe your test plan

Ran:

```
make generate
make manifests
go test ./api/...
bin/darwin-arm64/operator-sdk bundle validate ./bundle
```

Also verified `make generate && make manifests` produces no additional diff after generated files are up to date.

### Checklist


api/datadoghq/v1alpha1/datadoginstrumentation_types.go
```diff
@@ -26,6 +27,27 @@ type DatadogInstrumentationConfig struct {
...
+	// APM configures the APM product through Single Step Instrumentation for the target workload.
+	// +optional
+	APM DatadogInstrumentationAPMConfig `json:"apm,omitempty"`
+}
+
+// DatadogInstrumentationAPMConfig defines workload-scoped APM configuration.
+type DatadogInstrumentationAPMConfig struct {
+	// Enabled turns on APM via Single Step Instrumentation to automatically install the Datadog SDKs for supported
+	// languages with no additional configuration required.
+	Enabled bool `json:"enabled,omitempty"`
```
Mathew-Estafanous, line 40: Is an `enabled` section needed? The APM config already implies this based if the section is added (enabled) or removed (disabled).

betterengineering: The other two fields are optional and have default values. So I thought it would be weird if we did something like:
```yaml
config:
  apm: {}
```

The current design looks like this, and it's what most users would need to do:
```yaml
config:
  apm:
    enabled: true
```

I agree though, this case feels redundent:
```yaml
config:
  apm:
    enabled: true
    ddTraceVersions:
      python: "3"
```

Do you think there is a way around this?

Mathew-Estafanous: Ahh I see. I can't think of anything better to be honest and `apm: {}` is definitely not the right choice. I'm good to keep enabled. 👍🏽

Mathew-Estafanous (approved):
Looks great!

PR #3027, opened 2026-05-18
[CONTP-1569] [CONTINT-5553] Add podCollectionMode field on kubeStateMetricsCore for node-side pod collection

zhuminyi:
### What does this PR do?

  Adds `features.kubeStateMetricsCore.podCollectionMode` options are `default` and `node_kubelet` on the
  v2alpha1 `DatadogAgent` CRD.

```
  kind: DatadogAgent
  metadata:
    name: datadog
    namespace: datadog
  spec:
    features:
      clusterChecks:
        enabled: true
        useClusterChecksRunners: true  
      kubeStateMetricsCore:
        enabled: true
        podCollectionMode: node_kubelet   # THE SINGLE NEW TOGGLE 
```

  When set to `node_kubelet`, the operator:
  - Updates the operator-generated cluster-side KSM ConfigMap (unless the user supplied their own `conf`) by:
    - Setting `pod_collection_mode: cluster_unassigned` and `cluster_aggregates_enabled: true` on the primary instance so it collects unscheduled pods while suppressing cluster-aggregate metrics.
    - Adding a second instance with `pod_collection_mode: cluster_aggregates_only` to collect those aggregate metrics centrally.
  - Generates a separate node-agent ConfigMap containing a pods-only check with `pod_collection_mode: node_kubelet`, `cluster_aggregates_enabled: true`, and `collectors: [pods]`.

  When a user supplies `features.kubeStateMetricsCore.conf`, the operator still creates the node-side check but does not modify the user-owned cluster-side YAML. The user must configure both `cluster_unassigned` and `cluster_aggregates_only` cluster-side instances themselves.

This is equivalent to the following manual configuration:
```yaml
spec:
  features:
    kubeStateMetricsCore:
      enabled: true
      conf:
        configData: |-
          cluster_check: true
          init_config:
          instances:
            - skip_leader_election: true
              cluster_aggregates_enabled: true # silence aggregate metrics
              pod_collection_mode: cluster_unassigned # collect only unscheduled pods
            - skip_leader_election: true
              pod_collection_mode: cluster_aggregates_only # collect aggregate metrics centrally

  override:
    nodeAgent:
      extraConfd:
        configDataMap:
          kubernetes_state_core.yaml: |-
            init_config:
            instances:
              - collectors:
                  - pods
                cluster_aggregates_enabled: true # silence aggregate metrics
                pod_collection_mode: node_kubelet # collect scheduled pods from the node's Kubelet
```

  Result: scheduled-pod metrics are emitted by each node agent locally from the Kubelet (no API-server traffic for pods). Unscheduled pods and every other KSM resource continue to be collected by the primary cluster-side instance, while cluster-aggregate metrics are collected by the dedicated `cluster_aggregates_only` instance. Default behavior is semantically unchanged when the field is unset.

  ### Motivation

  In large clusters, pod metrics dominate the cardinality of the monolithic `kubernetes_state_core`
  cluster check, making it the scaling bottleneck. The agent has supported offloading pod collection to
  node-agents via `pod_collection_mode: node_kubelet` since 7.58, but customers had to assemble two
  coordinated configs manually (`features.kubeStateMetricsCore.conf.configData` for the cluster side,
  `override.nodeAgent.extraConfd` for the node side) and carefully avoid double-collection or losing
  unscheduled-pod metrics as well as live without `.total` aggregate metrics. This field collapses that into a single typed toggle.

  ### Describe your test plan

**Unit tests added in this PR:**

  - `TestKsmCheckConfigPodCollectionOnNode`
    - When enabled, asserts the cluster-side configuration contains:
      - A `cluster_unassigned` instance with the full collector list and
        `cluster_aggregates_enabled: true`.
      - A separate `cluster_aggregates_only` instance without a collector list or
        `cluster_aggregates_enabled`.
    - When disabled, asserts there is only one instance and no
      `pod_collection_mode` is emitted.

  - `Test_ksmFeature_buildKSMCorePodsOnNodeConfigMap`
    - Asserts the node-side ConfigMap metadata and complete parsed YAML shape.
    - Verifies the check has no `cluster_check` or `skip_leader_election` fields.
    - Verifies the pods-only instance contains:
      - `pod_collection_mode: node_kubelet`
      - `cluster_aggregates_enabled: true`
      - `collectors: [pods]`

### Manual end-to-end validation

```bash
# Both ConfigMaps are generated by the operator:
kubectl get cm -n datadog | grep state-metrics
# Expect:
#   <dda>-kube-state-metrics-core-config
#   <dda>-kube-state-metrics-core-pods-on-node-config

# The node agent mounts the node-side check:
kubectl get ds -n datadog datadog-agent \
  -o jsonpath='{.spec.template.spec.containers[?(@.name=="agent")].volumeMounts[?(@.name=="ksm-core-pods-on-node-config")]}'
# Expect: mountPath /etc/datadog-agent/conf.d/kubernetes_state_core.d, readOnly: true

# The node-side KSM check is loaded:
kubectl exec -n datadog ds/datadog-agent -c agent -- agent configcheck \
  | sed -n '/kubernetes_state_core check/,/===/p'
# Expect:
#   pod_collection_mode: node_kubelet
#   cluster_aggregates_enabled: true
#   collectors: [pods]

# The node-side check is running and emitting metrics:
kubectl exec -n datadog ds/datadog-agent -c agent -- agent status \
  | awk '/kubernetes_state_core/,/Status:/' | grep -E "Total Runs|Metric Samples"
# Expect: Total Runs > 0 and Metric Samples > 0 after approximately 30 seconds.

# The cluster-side configuration contains the split collector and aggregator:
kubectl exec -n datadog deploy/datadog-cluster-agent -- agent clusterchecks \
  | sed -n '/===== Checks on .*cluster-checks-runner/,/===== /p' \
  | grep -E "kubernetes_state_core|pod_collection_mode|cluster_aggregates_enabled"
# Expect:
#   cluster_unassigned with cluster_aggregates_enabled: true
#   a separate cluster_aggregates_only instance
```

### Minimum Agent Versions

- Node Agent: v7.82.0+
- Cluster Agent: v7.82.0+
- Cluster Checks Runner: v7.82.0+ when the KSM check runs there


api/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -947,6 +947,30 @@ type KubeStateMetricsCoreFeatureConfig struct {
 	// +optional
 	Conf *CustomConfig `json:"conf,omitempty"`
 
+	// PodCollectionMode controls where the KSM check collects pod metrics.
+	// When set to "node_kubelet" the operator switches the cluster-side instance to
+	// `pod_collection_mode: cluster_unassigned` (collects every other resource plus
```
AlexanderYastrebov: Example uses `pod_collection_mode` but the actual filed is camel-case. Is that OK?

internal/controller/datadogagent/feature/kubernetesstatecore/configmap_test.go
```diff
@@ -237,3 +239,45 @@ instances:
...
+		nodeAgentConfigMapName: "test-kube-state-metrics-core-pods-on-node-config",
+	}
+
+	got := f.buildKSMCorePodsOnNodeConfigMap()
+	assert.Equal(t, "foo", got.Namespace)
+	assert.Equal(t, "test-kube-state-metrics-core-pods-on-node-config", got.Name)
+
+	content, ok := got.Data[ksmCorePodsOnNodeCheckName]
+	if !ok {
+		t.Fatalf("expected key %q in ConfigMap data, got keys: %v", ksmCorePodsOnNodeCheckName, got.Data)
```
AlexanderYastrebov: nit: you could use require.True, require.NoError, etc since the code already uses assert package.

internal/controller/datadogagent/feature/kubernetesstatecore/configmap_test.go
```diff
@@ -237,3 +239,45 @@ instances:
...
+	}
+
+	// Cluster-check field MUST NOT be set: this is a file-provider check meant
+	// to run locally on every node agent, not be dispatched as a cluster check.
+	assert.NotContains(t, content, "cluster_check")
+
+	// Validate the YAML shape — pods-only collector and node_kubelet mode.
+	var parsed struct {
+		Instances []map[string]any `yaml:"instances"`
+	}
```
AlexanderYastrebov: nit: you could parse into map[string]any and then 
```go
expected := map[string]any{
  "instances": []map[string]any{
   ...
   },
}
assert.Equal(t, expected, parsed)
```
This way the test will catch any future changes including unexpected new fields.

AlexanderYastrebov (approved):

joepeeples (approved):

api/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -947,6 +947,37 @@ type KubeStateMetricsCoreFeatureConfig struct {
...
+	// YAML this corresponds to the snake_case check options
+	// `pod_collection_mode: cluster_unassigned` on the cluster side and
+	// `pod_collection_mode: node_kubelet` (with `collectors: [pods]`) on the
+	// node side; this CRD field itself is the camelCase `podCollectionMode`
+	// per Kubernetes API convention.
+	//
+	// Useful in large clusters where a single KSM instance is a bottleneck.
+	// Trade-off: cluster-wide aggregation metrics (kubernetes_state.pod.count,
+	// kubernetes_state.container.cpu_requested.total, and 15 others) become
+	// sliced per host instead of cluster totals; query them with
```
sblumenthal: This is not correct, the metrics full on break, there is no query that you can use to restore functionality

zhuminyi: Fixed with latest commit

internal/controller/datadogagent/feature/kubernetesstatecore/feature.go
```diff
@@ -142,10 +155,74 @@ func (f *ksmFeature) Configure(dda metav1.Object, ddaSpec *v2alpha1.DatadogAgent
...
+				componentsToCheck = append(componentsToCheck, v2alpha1.ClusterAgentComponentName)
+			}
+			for _, comp := range componentsToCheck {
+				ovr, ok := ddaSpec.Override[comp]
+				if !ok || ovr == nil || ovr.Image == nil {
+					continue
+				}
+				agentVersion := common.GetAgentVersionFromImage(*ovr.Image)
+				fallback := true // assume compatible when unparseable
+				if !utils.IsAboveMinVersion(agentVersion, podCollectionOnNodeMinVersion, &fallback) {
```
Mathew-Estafanous: nit: passing `nil` will already default fallback to true. can we use that

internal/controller/datadogagent/feature/kubernetesstatecore/feature.go
```diff
@@ -142,10 +155,74 @@ func (f *ksmFeature) Configure(dda metav1.Object, ddaSpec *v2alpha1.DatadogAgent
...
+					f.logger.Info(
+						"PodCollectionMode=node_kubelet was set alongside features.kubeStateMetricsCore.conf; " +
+							"the operator will deploy the node-side check but will not modify the user-supplied " +
+							"cluster-side config. To avoid double pod collection ensure the cluster-side instance " +
+							"either omits `pods` from `collectors` OR sets `pod_collection_mode: cluster_unassigned`. " +
+							"Note that omitting `collectors` entirely falls back to upstream KSM defaults, which " +
+							"include `pods`. To collect cluster-aggregate metrics, add a dedicated instance with " +
+							"`pod_collection_mode: cluster_aggregates_only` to the cluster-side config yourself; " +
+							"the operator does not add one for user-supplied configs.",
+					)
```
Mathew-Estafanous, line 217: nit: a more concise log message would be preferable, this is kind of long 😅 I'd also bump this up to WARN since this discrepancy could impact metric collection.

Mathew-Estafanous (approved):
A couple small nits but everything else looks good.

internal/controller/datadogagent/feature/kubernetesstatecore/feature.go
justin-lesko: I shortened the message a bit, unfortunately there is no `Warn()` log function - only `Info()` and `Error()`

PR #3035, opened 2026-05-21
FA rollout experiment fixes

levan-m:
### What does this PR do?

Addressing several bugs/issues:
* When changing same config back and forth to rollout, two controller revisions basically get version bump reusing same name/hash. This leads to accumulation of both marker annotations if one config update is rolled back. 
  ```yaml
      operator.datadoghq.com/experiment-promoted: "true"
      operator.datadoghq.com/experiment-rollback: "true"
  ```  
* Again when flipping same config, revisions of two controllerrevs increase while their `creationTimestamp` which is used as proxy for experiment start time, doesn't change. Once these revisions are old enough controller will timeout and rollback.
* If experiment times out, subsequent experiments will fail unless different config change to force new controller rev creation.
* Operator doesn't send task abort or timeout errors to FA (regardless how FA processes those, which now are ignored).

#### Easiest to review Commit by commit

b6ef9e55a   Decouples experiment timing from `ControllerRevision` metadata by adding explicit `StartedAt` field to `ExperimentStatus`. `rev.CreationTimestamp` might be stale if new spec has is same as earlier experiment. This could lead to immediate timeout if spec matched pre-existing controller rev spec.

d202cf708   rehydrate installer state from DatadogAgent on daemon startup to report state with correct task ID instead of empty one forcing FA to send start requests.

6cf24adee  report TaskState_ERROR for the original start task on timeout.

8509c37a6   replaces two annotations (`*experiment-promoted`, `*experiment-rollback`) on controller rev with a single annotation to avoid accidentally accumulating two mutually exclusive annotations and to simplify logic. Bug reproducible by flipping same config back and forth and intentionally failing experiment to force rollback.

5aa1b51cb  similar to 6cf24adee, report `abort` error for the experiment task when DDA is changed during experiment.

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Below args needs to be set on Operator (debug level logging recommended)
```yaml
        - --createControllerRevisions=true
        - --remoteUpdatesEnabled=true
        - --remoteConfigEnabled=true
```
and env vars to configure cluster name and endpoint
```yaml
        - name: DD_CLUSTER_NAME
          value: my-test-cluster
        - name: DD_SITE
          value: datad0g.com          # endpoint for internal testing and api/app key in the same org
```
**Happy path**
1. Apply DDA
2. Go to FA -> find cluster -> Configuration -> Edit -> Deploy
3. Switch to Deployments view in FA see it succeeds.
4. Check the change was applied in cluster.

**Failures**
timeout
2. Watch dda yaml `| yq .status.experiment`, once experiment is running scale Operator to 0 replicas.
3. Wait 15min
4. Scale back to 1
5. Operator marks experiment as `timed_out` and is rolled back. FA retries same experiment eventually failing deployment.
6. try happy path

abort
2. Watch dda yaml `| yq .status.experiment`, once experiment is running scale Operator to 0 replicas.
3. Apply updated DDA in cluster
4. Operator marks experiment as `aborted`, updated DDA should be reconciled in cluster. FA retries same experiment eventually failing deployment.
5. Try happy path.

### Checklist


coignetp (approved):

internal/controller/datadogagent/experiment.go
```diff
@@ -444,8 +471,13 @@ func (r *Reconciler) handleRollback(
 			return nil
 		}
 	}
-	if rev != nil {
-		elapsed := now.Sub(rev.CreationTimestamp.Time)
+	if rev != nil && instance.Status.Experiment.StartedAt != nil {
+		// status.experiment.startedAt is the timeout anchor. rev.CreationTimestamp
```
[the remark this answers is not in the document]
wdhif: Not yet released so not going to happen

pkg/fleet/daemon.go
```diff
@@ -250,6 +261,49 @@ func (d *Daemon) setTaskState(pkgName, taskID string, taskState pbgo.TaskState,
...
+// rehydrateInstallerState seeds the RC installer state from the
+// current DatadogAgent objects on disk. The rcClient's installer state
+// is in-memory only — after a daemon restart it would otherwise report
+// no in-progress experiment even when status.experiment.Phase is still
+// Running. That mismatch causes Fleet Automation to re-send the start
+// task and also makes reconcileTimedOutExperiment a no-op (its guard
+// `pkg.experimentConfigVersion == experiment.ID` never matches).
+//
+// Reads go through the API reader (not the cache) because the informer
+// cache may not be populated yet at the moment Start runs.
```
Mathew-Estafanous, line 273: Instead of doing this, can we use [cache.WaitForCacheSync](https://pkg.go.dev/k8s.io/client-go/tools/cache#WaitForCacheSync) to guarantee the informer is populated?

levan-m: I think using APIReader for direct reads and doing it before  `d.rcClient.Subscribe` makes intent clearer. Technically we could use `cache.WaitForCacheSync` and then read DDA from cache and initialize config version, would be same amount of code but RC subscription would block on whole cache sync vs single DDA list. In practice there is no big tradeoff.
Dropping this and relying only on WaitForCacheSync and informer event triggering config update would remove some code but create race between RC callbacks and informer events.

Mathew-Estafanous (approved):

PR #3043, opened 2026-05-26
[CONTP-1589] Add DDGR force sync period configuration

tbavelier:
### What does this PR do?

Adds `DD_GENERIC_RESOURCE_FORCE_SYNC_PERIOD` to configure the `DatadogGenericResource` force-sync interval, matching the existing monitor/dashboard/SLO behavior. The value is resolved once when the DDGR controller starts, logged at info level, and reused during reconciliation.

The configured value must be at least one minute. Non-integer values, `0`, and negative values are rejected once during controller startup and fall back to the default 60-minute force-sync period.

Also documents DDGR force sync and the existing DDGR namespace watch environment variable.

### Motivation

`DatadogGenericResource` already force-syncs remote resources every 60 minutes, but unlike related Datadog resource controllers, the interval was not configurable. This makes DDGR behavior easier to tune and easier to validate when remote resources drift from Kubernetes source of truth.

Rejecting values below one minute avoids surprising behavior where `0` or negative durations would force an update on every reconciliation loop.

### Additional Notes

`DD_GENERIC_RESOURCE_WATCH_NAMESPACE` already existed and falls back to `WATCH_NAMESPACE`; this PR only documents that behavior.

The reconcile tests construct the DDGR reconciler through `NewReconciler` and then override the resource handlers with mocks, so the production force-sync initialization path is covered without calling real Datadog APIs.

### Minimum Agent Versions

* Agent: N/A
* Cluster Agent: N/A

### Describe your test plan

Automated validation:

```shell
go test -count=1 ./internal/controller/datadoggenericresource
```

The tests cover valid force-sync configuration, invalid string values, `0`, negative values, and an elapsed force-sync period triggering a remote update.

Manual QA:

1. Start the operator with the DDGR controller enabled and an invalid force-sync period:

   ```shell
   DD_GENERIC_RESOURCE_FORCE_SYNC_PERIOD=abc
   ```

   Verify the operator logs contain this error once during controller startup:

   ```json
   {"level":"ERROR","logger":"controllers.DatadogGenericResource","msg":"Invalid value for generic resource force sync period. Defaulting to 60 minutes.","error":"strconv.Atoi: parsing \"abc\": invalid syntax"}
   ```

2. Restart the operator with a valid five-minute force-sync period:

   ```shell
   DD_GENERIC_RESOURCE_FORCE_SYNC_PERIOD=5
   ```

   Verify the operator logs contain this info log during controller startup:

   ```json
   {"level":"INFO","logger":"controllers.DatadogGenericResource","msg":"Setting generic resource force sync period","minutes":5}
   ```

3. Apply the notebook DDGR sample:

   ```shell
   kubectl apply -f examples/datadoggenericresource/notebook-sample.yaml
   ```

   Wait for the resource to be created and record the Datadog notebook ID:

   ```shell
   kubectl get datadoggenericresource ddgr-notebook-sample -o jsonpath='{.status.id}'
   ```

4. In the Datadog UI, open that notebook and make a visible edit, for example change the notebook name from `Example-Notebook` to `Example-Notebook edited from UI`.

5. Wait at least five minutes, then verify the operator force-sync restored the notebook to the Kubernetes spec from `examples/datadoggenericresource/notebook-sample.yaml`, including the original notebook name `Example-Notebook`.

6. Clean up the sample resource:

   ```shell
   kubectl delete -f examples/datadoggenericresource/notebook-sample.yaml
   ```

### Checklist

### DIFF
--- a/internal/controller/datadoggenericresource/controller.go
+++ b/internal/controller/datadoggenericresource/controller.go
@@ -3,6 +3,8 @@ package datadoggenericresource
 import (
 	"context"
 	"fmt"
+	"os"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


internal/controller/datadoggenericresource/controller.go
```diff
@@ -29,32 +31,60 @@ import (
...
+		} else {
+			forceSyncPeriod = time.Duration(forceSyncPeriodInt) * time.Minute
+		}
+	}
+
+	logger.Info("Setting generic resource force sync period", "minutes", int(forceSyncPeriod.Minutes()))
+	return &forceSyncPeriod
+}
+
+func (r *Reconciler) getForceSyncPeriod() time.Duration {
```
gabedos: nit: do we need this getter function if `r.forceSyncPeriod` will always be defined with a valid value?

gabedos (approved):

internal/controller/datadoggenericresource/controller.go
tbavelier: Good catch, it was only used for the test as we construct the reconciler directly instead of using the constructor where it would use the default. Updated the PR to instead use the constructor so it benefits from the default/always set value.

jeff-morgan-dd (approved):

PR #3047, opened 2026-05-27
feat(ddgr): surface Datadog Monitor state in CR status

wdhif:
### What does this PR do?

Surfaces the live Datadog-side state of Monitor resources managed by `DatadogGenericResource` (DDGR), matching the UX `DatadogMonitor` already provides. Three new fields appear on the CR status — `state`, `stateLastUpdateTime`, `stateLastTransitionTime` — populated on a ~60s cadence from the Datadog API. A new `StateSynced` condition signals refresh success/failure (`reason=GetError` when the API call fails; last-known state is preserved).

The design is intentionally type-agnostic at every layer (schema, condition, dispatch, reconcile branch). Only `MonitorHandler.refreshStatusFunc` does Monitor-specific work; the other five handlers are one-line no-ops. When SLO support lands in DDGR, its handler implements the same `refreshStatusFunc`, populates the same three generic fields, and uses the same `StateSynced` condition — zero schema changes required.

### Motivation

This PR closes that gap for Monitor and leaves the door open for SLO without committing to its schema.

### Additional Notes

- **Deliberate feature gap vs `DatadogMonitor`:** `triggeredState` (per-group breakdown — which hosts/services are firing inside a grouped monitor) is **intentionally omitted** from DDGR. It's Monitor-specific and can't be expressed type-agnostically without polluting the generic schema. Users who need per-group visibility should use the dedicated `DatadogMonitor` CRD. If a real need emerges later, a Monitor-only `triggeredState` field can be added at the top level without breaking anything.
- **State-conversion helper is duplicated, not extracted.** `setMonitorStateOnStatus` in `monitors.go` mirrors `convertStateToStatus` in `internal/controller/datadogmonitor/controller.go` by design — a one-time minor DRY violation in exchange for a smaller blast radius (no shared-package refactor that would touch `DatadogMonitor`). Both copies carry a cross-reference comment so future divergence is easy to spot.

### Minimum Agent Versions

No agent-side change. No minimum version bump.

* Agent: no change
* Cluster Agent: no change

### 1. Install Operator and DDGR

```bash
helm repo update datadog
helm install datadog-operator datadog/datadog-operator --devel \
  --set apiKeyExistingSecret=datadog-secret \
  --set appKeyExistingSecret=datadog-secret \
  --set datadogGenericResource.enabled=true \
  --set datadogCRDs.crds.datadogGenericResources=true \
  --set datadogCRDs.crds.datadogMonitors=true
```

### 2. Apply a Monitor `DatadogGenericResource`

```yaml
# ddgr-monitor-state-test.yaml
kind: DatadogGenericResource
metadata:
  name: ddgr-monitor-state-test
spec:
  type: monitor
  jsonSpec: |-
    {
      "name": "CONTP-1686 state test",
      "type": "metric alert",
      "query": "avg(last_5m):avg:system.cpu.user{*} by {host} > 100",
      "message": "state-refresh test — Notify nobody",
      "tags": ["test:contp-1686"],
      "options": {"thresholds": {"critical": 100, "warning": 80}}
    }
```

```bash
kubectl apply -f ddgr-monitor-state-test.yaml
```

### 3. Verify

```bash
kubectl get ddgr -o wide
```

Expected — new `STATE` and `LAST STATE SYNC` columns populated:

```
NAME                      ID          SYNC STATUS   STATE     LAST STATE SYNC        AGE
ddgr-monitor-state-test   287954487   OK            No Data   2026-05-27T09:33:50Z   76s
```

```bash
kubectl describe ddgr ddgr-monitor-state-test
```

Expected — `StateSynced` condition + new `State*` status fields:

```
Status:
  Conditions:
    Message:                   DatadogGenericResource Created
    Reason:                    CreatingGenericResource
    Status:                    True
    Type:                      Created
    Message:                   State refreshed from Datadog
    Reason:                    Synced
    Status:                    True
    Type:                      StateSynced
  State:                       No Data
  State Last Transition Time:  2026-05-27T09:33:50Z
  State Last Update Time:      2026-05-27T09:33:50Z
  Sync Status:                 OK
```

### Checklist

- [x] PR has at least one valid label: `bug`, `enhancement`, `refactoring`, `documentation`, `tooling`, and/or `dependencies` — please add `enhancement` (primary) and optionally `documentation`

[CONTP-1686]: https://datadoghq.atlassian.net/browse/CONTP-1686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/datadog_generic_resource_dev.md
```diff
@@ -23,28 +23,32 @@ The general steps to add a new type can be summarized to:
...
     ```go
-    v1alpha1.<Resource>: &<Resource>Handler{auth: ddClient.Auth, client: ddClient.<Resource>Client},
+    v1alpha1.<Resource>: &<Resource>Handler{client: clients.<Resource>Client},
     ```
 
 ## Post-development tasks
 
 1. Update the supported resources table in [Overview of the DatadogGenericResource controller][2].
 2. Provide an example manifest with the new resource for the [examples][3] folder.
+3. **If your `refreshStatusFunc` implementation populates `status.State`** (i.e. your resource exposes a live Datadog-side state — like Monitors do, but unlike Dashboards/Notebooks):
```
wdhif:
```suggestion
3. **If your `refreshState` implementation populates `status.State`** (i.e. your resource exposes a live Datadog-side state — like Monitors do, but unlike Dashboards/Notebooks):
```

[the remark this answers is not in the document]
wdhif: Applied manually due to GitHub incident.

docs/datadog_generic_resource.md
```diff
@@ -159,6 +164,28 @@ Further example manifests are provided [in the supported resources table](#suppo
...
+| `.status.conditions[type=StateSynced]` | `True` after a successful state refresh; `False` with `reason=GetError` when the most recent refresh failed (last-known `state` is preserved). |
+
+Inspect quickly via:
+
+```shell
+kubectl get datadoggenericresource    # shows state and last state sync columns
+kubectl wait --for=condition=StateSynced datadoggenericresource/<name>
+```
+
+The controller refreshes `state` roughly every 60 seconds during reconciliation. Failures are visible only via the `StateSynced` condition — they do not break the reconcile loop and the last-known `state` is retained until a subsequent refresh succeeds.
```
OliviaShoup, line 185:
```suggestion
The controller refreshes `state` roughly every 60 seconds during reconciliation. Failures are visible only via the `StateSynced` condition. Failures do not break the reconcile loop and the last-known `state` is retained until a subsequent refresh succeeds.
```

OliviaShoup (approved):
thanks for the PR! 🚀  left one suggestion to chop up a long sentence into two

internal/controller/datadoggenericresource/resource_handler.go
```diff
@@ -16,11 +16,23 @@ type CreateResult struct {
 	Creator     string
 }
 
+// ResourceState represents a snapshot of the live Datadog-side state of a resource.
+// The value of State is resource-type-dependent (e.g. for Monitors: OK, Alert, Warn, No Data, ...).
+type ResourceState struct {
+	State string
+}
+
```
tbavelier: I don't think `resourcestate` struct has value, we can simply use `*string` as this state is a string that is shown in the status (and keeping the pointer to differentiate from `""`. I know SLO have a "richer" status (for instance error budget, target, etc.) on top of status but ultimately, this then becomes too specific.
If we think / want to add more to the status, we can always go back to this struct with more fields, but I don't think that's a short term plan

tbavelier (changes requested):
👍 , except `ResourceState` struct IMHO, see comment

internal/controller/datadoggenericresource/resource_handler.go
wdhif: I think that's fair 👍
Fixed in https://github.com/DataDog/datadog-operator/pull/3047/commits/740f7aa5aa19a9db4b561382831c80d46394a9cf

docs/datadog_generic_resource_dev.md
tbavelier, line 1: nit: This file still references `ResourceState` instead of the updated `string` signature

tbavelier (approved):
LGTM, unblocking, but the dev documentation should be updated accordingly

docs/datadog_generic_resource_dev.md
wdhif: Ah good catch! I'll solve it 👍

wdhif: Done!

hmahmood (approved):

PR #3056, opened 2026-05-28
[CASCL-1292] Point the no-private-subnet error at all the install options

L3n41c:
### What does this PR do?

Expands the error message shown by `kubectl datadog autoscaling cluster install` when its default `fargate` mode auto-detects **no private subnet** in the cluster VPC. The message now names all three ways forward instead of only one:

- add a private subnet to the cluster VPC, or
- pass `--fargate-subnets` to specify them explicitly, or
- use `--install-mode=existing-nodes` to deploy on existing nodes instead of Fargate.

### Motivation

During QA of the Fargate mode (CASCL-1292), a tester hit the private-subnet requirement on a VPC with only public subnets. Fargate is the default mode and the requirement is a hard AWS constraint (`AWS::EKS::FargateProfile` only accepts private subnets). The old message mentioned only `--fargate-subnets`, so the tester had to discover on their own that `--install-mode=existing-nodes` installs with no subnet requirement at all.

### Additional Notes

Deliberately minimal: a single error-message change in `GetClusterPrivateSubnets` plus a test update. No behavior change for successful installs.

### Minimum Agent Versions

No minimum Agent/Cluster Agent version required.

### Describe your test plan

- Unit: `TestGetClusterPrivateSubnets` now asserts the no-subnet error mentions the private-subnet requirement, `--fargate-subnets`, and `--install-mode=existing-nodes`.
- `go test ./cmd/kubectl-datadog/autoscaling/...`, `make lint` (0 issues), `go vet`, `gofmt` all pass.

### Checklist

- [x] All commits are signed

### DIFF
--- a/cmd/kubectl-datadog/autoscaling/cluster/common/eks/privatesubnets_test.go
+++ b/cmd/kubectl-datadog/autoscaling/cluster/common/eks/privatesubnets_test.go
@@ -60,7 +60,7 @@ func TestGetClusterPrivateSubnets(t *testing.T) {
 		ec2Err          error
 		expectedPrivate []string
 		expectError     bool
-		errorContains   string
+		errorContains   []string
 	}{
 		{
 			name:           "mix of public and private with explicit associations",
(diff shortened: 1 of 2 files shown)


cmd/kubectl-datadog/autoscaling/cluster/common/eks/privatesubnets_test.go
```diff
@@ -173,8 +173,13 @@ func TestGetClusterPrivateSubnets(t *testing.T) {
 					Routes:       []ec2types.Route{igwRoute()},
 				},
 			},
-			expectError:   true,
-			errorContains: "no private subnet found",
+			expectError: true,
```
AlexanderYastrebov, line 176: nit: `expectError` could be replaced with `len(errorContains) > 0`

AlexanderYastrebov (approved):

PR #3057, opened 2026-05-28
Validate and auto-resolve AWS region before Karpenter operations

L3n41c:
### What does this PR do?

Reconciles the AWS region with the target EKS cluster before the `kubectl datadog autoscaling cluster` commands (install/update/uninstall) build their AWS clients:

- **`AWS_REGION` unset but derivable** from the kubeconfig context ARN → derive the region and reload the AWS config with `config.WithRegion` (so credential providers also pick it up), then proceed with a log notice.
- **`AWS_REGION` unset and not derivable** → clear, actionable error instead of the opaque STS "Missing Region" noise.
- **`AWS_REGION` set but different** from the cluster's region → hard error (`RegionMismatchError`) in all three commands.

### Motivation

QA hit two confusing failures, both rooted in the AWS region:

1. With `AWS_REGION` unset, the only feedback was an opaque, deeply-wrapped STS error: `failed to get AWS caller identity: operation error STS: GetCallerIdentity, failed to resolve service endpoint, endpoint rule error, Invalid Configuration: Missing Region`.
2. With `AWS_REGION` set to the wrong region, operations silently targeted the wrong place — e.g. uninstall printing `Stack ... does not exist, skipping deletion.`

The kubeconfig context for an EKS cluster is an ARN (`arn:aws:eks:<region>:<account>:cluster/<name>`) that already carries the region — the same source [PR #2892](https://github.com/DataDog/datadog-operator/pull/2892) uses for the AWS account-consistency check. This change reuses it for the region, with no extra AWS API call.

### Additional Notes

- Region resolution lives in `clients.Build`, **before** the service clients are constructed, because credential providers (assume-role / web-identity STS clients) capture the region at config-load time — a post-load mutation wouldn't reach them. The config is reloaded with the derived region only in the derive path.
- Refactors the kubeconfig-ARN parse into a shared `getClusterARNFromKubeconfig` helper used by both the account and region extractors. The helper is tightened to only trust EKS cluster ARNs (`Service == "eks"` + `cluster/` resource prefix).
- `RegionMismatchError` mirrors the existing `AccountMismatchError`.

### Minimum Agent Versions

N/A — this is a `kubectl-datadog` plugin change, not an agent change.

### Describe your test plan

- Unit tests added for `resolveRegion` (all branches incl. mismatch via `errors.As`, GovCloud partitions) and `getClusterARNFromKubeconfig` (EKS/GovCloud ARNs, plain names, eksctl FQDNs, non-EKS ARN rejection).
- `go build ./cmd/kubectl-datadog/...`, `go vet`, and the package tests pass; full `make lint` reports 0 issues.
- Manual (kubeconfig context = an EKS ARN):
  - **Unset region** (profile without a region) → logs `AWS region not set; using "us-east-2" from the kubeconfig context.` and proceeds.
  - **Wrong region** (`AWS_REGION=us-west-2` against a `us-east-2` cluster) → fails immediately with `RegionMismatchError` for install, update, and uninstall.
  - **Correct region** → unchanged behavior.

### Checklist

### DIFF
--- a/cmd/kubectl-datadog/autoscaling/cluster/common/clients/clients.go
+++ b/cmd/kubectl-datadog/autoscaling/cluster/common/clients/clients.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"strings"
 
 	awssdk "github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/aws/arn"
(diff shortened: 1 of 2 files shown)


L3n41c:
stale PR

PR #3086, opened 2026-06-04
[CONTP-1609] Auto-inject agent toleration when untaint controller is enabled

adel121:
### What does this PR do?

Updates the datadog agent controller to Auto-inject agent toleration when untaint controller is enabled

Relates to https://github.com/DataDog/datadog-operator/pull/2753

### Motivation

For the untaint controller to work, the agent DaemonSet must tolerate the startup taint (agent.datadoghq.com/not-ready=presence:NoSchedule) — otherwise the agent can never schedule on tainted nodes, the taint is never removed, and all workloads are blocked indefinitely. Without auto-injection this is a silent foot-gun: enabling the flag without manually adding the toleration produces no error but a permanently deadlocked cluster.

When --untaintControllerEnabled=true, the operator automatically injects the toleration into the node agent DaemonSet spec, following the same pattern used by other feature flags that influence agent pod spec assembly (IntrospectionEnabled, DatadogAgentProfileEnabled).

Pass UntaintControllerEnabled through datadogagent.ReconcilerOptions

Inject toleration in the node agent DaemonSet builder (idempotent — no duplicate if user also sets it manually)

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Follow the same testing instructions as https://github.com/DataDog/datadog-operator/pull/2753 , but instead of adding the toleration manually to the datadog agent, ensure the agent gets the toleration automatically.

### Checklist

### DIFF
--- a/docs/untaint_controller.md
+++ b/docs/untaint_controller.md
@@ -49,6 +49,13 @@ args:
   - --untaintControllerEnabled=true
 ```
 
+When this flag is enabled, the operator also injects a toleration for
+`agent.datadoghq.com/not-ready=presence:NoSchedule` into the node Agent
+DaemonSet (or ExtendedDaemonSet) pod template, unless an equivalent toleration
+is already present. That avoids a deadlock where the node stays tainted because
+the Agent pod cannot schedule without the toleration—especially when admission
+webhook auto-injection is not in use.
+
 ## Configuration
 
 All other tuning knobs are environment variables on the operator pod. Values
--- a/docs/untaint_controller.md
+++ b/docs/untaint_controller.md
@@ -52,8 +52,8 @@ args:
 When this flag is enabled, the operator also injects a toleration for
 `agent.datadoghq.com/not-ready=presence:NoSchedule` into the node Agent
 DaemonSet (or ExtendedDaemonSet) pod template, unless an equivalent toleration
-is already present. That avoids a deadlock where the node stays tainted because
-the Agent pod cannot schedule without the toleration—especially when admission
+is already present. This avoids a deadlock where the node stays tainted because
+the Agent pod cannot schedule without the toleration, especially when admission
 webhook auto-injection is not in use.
 
 ## Configuration
--- a/internal/controller/datadogagent/component/agent/untaint_toleration.go
+++ b/internal/controller/datadogagent/component/agent/untaint_toleration.go
@@ -7,30 +7,14 @@ package agent
 
 import (
 	corev1 "k8s.io/api/core/v1"
-)
 
-// Toleration values must stay in sync with internal/controller/untaint_controller.go
-// (agent-not-ready startup taint).
-const (
-	agentNotReadyTaintKey    = "agent.datadoghq.com/not-ready"
-	agentNotReadyTaintValue = "presence"
+	"github.com/DataDog/datadog-operator/pkg/untaint"
 )
-
-// agentNotReadyStartupToleration is the toleration that allows the node Agent to
-// schedule on nodes carrying agent.datadoghq.com/not-ready=presence:NoSchedule.
-func agentNotReadyStartupToleration() corev1.Toleration {
-	return corev1.Toleration{
-		Key:      agentNotReadyTaintKey,
-		Operator: corev1.TolerationOpEqual,
-		Value:    agentNotReadyTaintValue,
-		Effect:   corev1.TaintEffectNoSchedule,
-	}
-}
 
 // podToleratesAgentNotReadyStartup returns true if the pod already tolerates the
 // agent-not-ready NoSchedule taint (exact Equal match or broader Exists on the key).
 func podToleratesAgentNotReadyStartup(tolerations []corev1.Toleration) bool {
-	want := agentNotReadyStartupToleration()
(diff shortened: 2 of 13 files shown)


adel121:
@codex

docs/untaint_controller.md
```diff
@@ -49,6 +49,13 @@ args:
   - --untaintControllerEnabled=true
 ```
 
+When this flag is enabled, the operator also injects a toleration for
+`agent.datadoghq.com/not-ready=presence:NoSchedule` into the node Agent
+DaemonSet (or ExtendedDaemonSet) pod template, unless an equivalent toleration
+is already present. That avoids a deadlock where the node stays tainted because
+the Agent pod cannot schedule without the toleration—especially when admission
+webhook auto-injection is not in use.
```
OliviaShoup, line 57:
```suggestion
DaemonSet (or ExtendedDaemonSet) pod template, unless an equivalent toleration
is already present. This avoids a deadlock where the node stays tainted because
the Agent pod cannot schedule without the toleration, especially when admission
webhook auto-injection is not in use.
```

OliviaShoup (approved):
thanks for the PR! left a comment with a minor suggestion

internal/controller/datadogagent/component/agent/untaint_toleration.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// podToleratesAgentNotReadyStartup reports whether tolerations tolerate the
+// agent-not-ready startup taint, using the same rules as the scheduler/kubelet
+// (corev1.Toleration.ToleratesTaint). Comparison-operator tolerations (Lt/Gt) are
+// ignored unless the cluster enables that feature (we pass false).
+func podToleratesAgentNotReadyStartup(tolerations []corev1.Toleration) bool {
+	taint := untaint.AgentNotReadyTaint()
+	for i := range tolerations {
+		if tolerations[i].ToleratesTaint(klog.Background(), &taint, false) {
```
levan-m: no need to access klog directly, can be passed from the controller.

internal/controller/datadogagentinternal/controller_reconcile_agent.go
```diff
@@ -95,6 +95,10 @@ func (r *Reconciler) reconcileV2Agent(ctx context.Context, requiredComponents fe
 
 		experimental.ApplyExperimentalOverrides(objLogger, ddai, podManagers)
 
+		if r.options.UntaintControllerEnabled {
+			componentagent.EnsureAgentNotReadyStartupToleration(&podManagers.PodTemplateSpec().Spec)
```
levan-m: pass `objLogger` here, it will have relevant attributes. Alternative, pass context check start of the function and create logger inside the function `ctrl.LoggerFrom(ctx).WithValues` with additional attributes.

internal/controller/datadogagent/component/agent/untaint_toleration.go
```diff
@@ -0,0 +1,39 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// EnsureAgentNotReadyStartupToleration appends the agent-not-ready Equal toleration
+// when not already tolerated per Kubernetes toleration matching.
+func EnsureAgentNotReadyStartupToleration(spec *corev1.PodSpec) {
+	if spec == nil {
```
levan-m: nit - redundant nil check since PodSpec is struct in PodTemplateSpec and can't be nil here.

pkg/untaint/taint.go
```diff
@@ -0,0 +1,45 @@
...
+const (
+	// AgentNotReadyTaintKey is the key for the startup taint applied to nodes
+	// until the Datadog Agent is ready (or removed by timeout policy).
+	AgentNotReadyTaintKey = "agent.datadoghq.com/not-ready"
+	// AgentNotReadyTaintValue is the value paired with AgentNotReadyTaintKey.
+	AgentNotReadyTaintValue = "presence"
+)
+
+// AgentNotReadyTaintEffect is the effect for the agent-not-ready startup taint.
+const AgentNotReadyTaintEffect = corev1.TaintEffectNoSchedule
```
levan-m: nit - doesn't have to be public and can be inlined.

internal/controller/datadogagent/controller_v2_test.go
```diff
@@ -248,6 +261,35 @@ func buildClient(t *testing.T, tt testCase, s *runtime.Scheme) client.Client {
...
 
+func TestReconcileDDA_UntaintController_injectsAgentNotReadyToleration(t *testing.T) {
+	const resourcesName = "foo"
+	const resourcesNamespace = "bar"
+	const dsName = "foo-agent"
+	defaultRequeueDuration := 15 * time.Second
+
+	wantTol := untaint.AgentNotReadyEqualToleration()
+	tt := testCase{
+		name: "untaint controller enabled injects agent-not-ready toleration on node agent DS",
```
levan-m, line 272: for completeness would be nice to have case asserting tolerations aren't added when feature is disabled.

adel121:
@levan-m thanks for the review 🙇 

Addressed all your comments!

levan-m (approved):

PR #3108, opened 2026-06-08
Extend introspection and support DDAI code path

levan-m:
### What does this PR do?

CONTP-1577

Change extends [introspection](https://github.com/DataDog/datadog-operator/blob/main/docs/introspection.md)  detection mechanism to allow OOTB support for [control plane monitoring](https://github.com/DataDog/datadog-operator/blob/main/docs/control_plane_monitoring.md) on EKS (non-auto) and Openshift.

Change summary:
* Moves detection logic outside reconciler.
* Adds detection based on node operator is running as primary source, with an optional fallback (iff operator is watching nodes) to listing all nodes in the cluster.
* Persists detection result in DDA status.
* Implements precedence order
   1. User provided `datadoghq.com/provider` annotation.
   2. Non-empty detection result.
   3. Provider persisted in DDA status.
* Adds detection grace period in reconciler (when provider is empty) to prevent unnecessary resource churn.

### Motivation

Since 1.24 users had to disable DDAI code path to use OOTB control plane monitoring, since 1.27 this option isn't available. In addition users needed to set `introspectionEnabled` flag for caching nodes for List operation. This change resolves former and make latter optional.

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan
1. Provision regular EKS cluster
2. Install Operator
3. Apply minimal DDA
4. Verify control plane metrics like `kube_apiserver.*`, `kube_scheduler.*` appear in the app.

### Checklist


internal/controller/datadogagent/controller_reconcile_v2.go
```diff
@@ -198,3 +210,95 @@ func (r *Reconciler) getNodeList(ctx context.Context) ([]corev1.Node, error) {
...
+	}
+	status.ClusterProvider = provider
+
+	var reason, message string
+	switch {
+	case source == clusterProviderSourceUser:
+		reason = clusterProviderSourceUser
+		message = fmt.Sprintf("Cluster provider set to %q by user annotation.", provider)
+	case kubernetes.IsSpecificProvider(provider):
+		reason = clusterProviderReasonDetected
```
Mathew-Estafanous: Why isn't `source == clusterProviderSourceDetected` used here to determine the reason?

internal/controller/datadogagent/controller_reconcile_v2.go
```diff
@@ -198,3 +210,95 @@ func (r *Reconciler) getNodeList(ctx context.Context) ([]corev1.Node, error) {
...
+	statusProvider := instance.Status.ClusterProvider
+
+	// 1. Live detection. Anti-flap: keep a previously *detected* specific provider
+	// over a live default (a transient blip shouldn't tear down config). This guard
+	// applies only to detected values — a user override that was set and then
+	// removed is not pinned, so removing the annotation cleanly returns to detection.
+	if live, ok := detector.Provider(); ok {
+		if wasProviderDetected(instance) && kubernetes.IsSpecificProvider(statusProvider) && !kubernetes.IsSpecificProvider(live) {
+			return statusProvider, clusterProviderSourceDetected, false
+		}
```
Mathew-Estafanous, line 257: ❓ Is it possible for the new live detection to be persistent and not a 'blip?' Say a GKE cluster going from entirely GKE COS nodes to regular ubuntu. If so, wouldn't this gate always choose the status provider over live?

pkg/introspection/detector.go
```diff
@@ -0,0 +1,204 @@
...
+// detect runs the stages in order, returning the first success or nil if all
+// failed this round (leaving any prior result intact). Stage 1 (operator-node read)
+// is authoritative on success, even for the default provider; Stage 2 (node
+// list) is the fallback when Stage 1 can't run or errors.
+func (d *Detector) detect(ctx context.Context) *detection {
+	// Stage 1 — opportunistic keyed read of the operator's own node.
+	if d.nodeName != "" {
+		node := &corev1.Node{}
+		if err := d.apiReader.Get(ctx, types.NamespacedName{Name: d.nodeName}, node); err != nil {
+			d.logger.V(1).Info("operator-node provider read failed; trying cluster-node-list fallback", "node", d.nodeName, "error", err)
```
Mathew-Estafanous, line 172: ❓  Looking at this detection logic, is it possible the node that the operator is running on doesn't tell the full picture? Say on a GKE cluster that has both COS and ubuntu nodes. Depending on where the operator lands, it will 'detect' something different.

internal/controller/datadogagent/controller_reconcile_v2.go
levan-m: First case is `clusterProviderSourceUser`, which effectively leaves only `clusterProviderSourceDetected`. Added the conditional in a77b92bfe3e515c4da73c0a75c5452931c114c51 regardless to make it more obvious.

internal/controller/datadogagent/controller_reconcile_v2.go
levan-m: It was possible as you pointed out with COS example, however this change should only handle cluster/global provider (EKS, Openshift, etc.) user cases. Intention here is correct, once cluster level provider detected, it can't shouldn't become empty, if we detect we are on EKS node label changes shouldn't revert that. Only way to override this is user provided annotation.
Made fix in a77b92bfe3e515c4da73c0a75c5452931c114c51 provider.go to separate cluster and node level provider detection.

pkg/introspection/detector.go
levan-m: this should be answered in the above comment; in this PR we only care global/cluster level changes for use cases like control plane monitoring for EKS and Openshift, autopilot.

Mathew-Estafanous (approved):
provider introspection look good

tbavelier (changes requested):
LGTM overall, but two points:

- This turns introspection "on" by default while it was previously opt-in: this is what we want, but should be called out. It is not immediately clear that introspection can be disabled using `datadoghq.com/provider: ""` annotation on DDA
- As a follow-up to the previous point and that one is why I'm "requesting changes", this breaks OpenShift users that did not have introspection before as we enable control-plane monitoring which mounts a secret in the Agent daemonset, however, this secret needs to be manually copied over (we log the command to run).

With the second point in mind, considering the operator has cluster-wide secrets access, I'm proposing **as part of the control-plane monitoring feature for OpenShift** to do this operation for users: https://github.com/DataDog/datadog-operator/pull/3131 so that way we don't break them and they start benefiting from it automatically. My approach is not perfect, it has tradeoffs (direct api read (not a write) every reconciliation to avoid the error below), I'm open to alternatives (e.g. add this namespace in bundles that we deploy on OpenShift / do something else)

```json
{"level":"INFO","ts":"2026-06-12T09:46:10.796Z","logger":"controllers.DatadogAgentInternal","msg":"Unable to copy OpenShift etcd metric client secret automatically","namespace":"system","name":"datadog","reconcileID":"dd20b327-ef67-45a2-88f4-91e849090cf0","kind":"DatadogAgentInternal","namespace":"openshift-etcd-operator","name":"etcd-metric-client","error":"unable to get: openshift-etcd-operator/etcd-metric-client because of unknown namespace for the cache"}
```

tbavelier (approved):
Conflict with main to resolve, but LGTM!

khewonc (approved):

tbavelier (approved):

PR #3131, opened 2026-06-12
Copy OpenShift etcd metric secret

tbavelier:
## What

PoC for automatically copying the OpenShift etcd metric client Secret when OpenShift control plane monitoring is enabled.

- Pass an uncached API reader into feature construction from the DatadogAgentInternal reconciler.
- Read `openshift-etcd-operator/etcd-metric-client` directly from the API server.
- Add a managed `etcd-metric-client` Secret in the DatadogAgent namespace through the existing dependency store.
- Keep the manual-copy log fallback when the source Secret cannot be read.

## Why

OpenShift etcd monitoring mounts `etcd-metric-client` in the DatadogAgent namespace. Previously the operator only logged an `oc get ... | oc apply ...` command for users to copy it manually. This secret is necessary otherwise the pod fails to initialize. Now that introspection is enabled by default through target PR, it would "break" users that do not expect/use control-plane monitoring.

The operator ClusterRole already allows core Secret `get/list/watch/create/update/patch/delete`, so it has the required RBAC to perform this copy.

## Alternatives considered

- Watch `openshift-etcd-operator` by default: rejected for this PoC because it broadens the controller cache to a sensitive platform namespace, keeps platform Secrets in memory, and ties this behavior to bundle/install packaging.
- Use the cached client: rejected because default installs do not watch `openshift-etcd-operator`, producing `unknown namespace for the cache`.
- Use a direct API server read: chosen for the source Secret because this is a narrow one-off read outside the watched namespace set. The copied target Secret still uses the normal dependency store for ownership, apply, and cleanup.

## Follow-ups before productionizing

This PoC performs the direct source Secret read during each relevant DDAI reconciliation. Better production options include:

- Read only when the target Secret is missing or differs.
- Add a small TTL/cache around the source Secret content.
- Move copying into a dedicated helper/reconciler with status/error reporting.
- Consider a narrow watch only if it can target the exact source object without broadly caching the namespace.

## Validation

- Tested on my OpenShift cluster

### DIFF
--- a/internal/controller/datadogagent/feature/controlplanemonitoring/feature.go
+++ b/internal/controller/datadogagent/feature/controlplanemonitoring/feature.go
@@ -136,7 +136,7 @@ func (f *controlPlaneMonitoringFeature) copyOpenShiftEtcdSecret(managers feature
 	defer cancel()
 	if err := f.client.Get(ctx, sourceKey, source); err != nil {
 		f.logger.Info("Unable to copy OpenShift etcd metric client secret automatically", "namespace", sourceKey.Namespace, "name", sourceKey.Name, "error", err)
-		return false
+		return f.keepExistingOpenShiftEtcdSecret(managers)
 	}
 
 	target := &corev1.Secret{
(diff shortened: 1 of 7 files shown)


internal/controller/datadogagent/feature/controlplanemonitoring/feature.go
```diff
@@ -109,6 +116,43 @@ func (f *controlPlaneMonitoringFeature) ManageDependencies(managers feature.Reso
...
+	// Read the source Secret directly from the API server instead of relying on
+	// the controller cache: default operator installs do not watch the
+	// openshift-etcd-operator namespace, and broadening the cache would keep
+	// sensitive platform Secrets in memory just for this one copy.
+	source := &corev1.Secret{}
+	sourceKey := types.NamespacedName{
+		Namespace: etcdCertsSourceNamespace,
+		Name:      etcdCertsSecretName,
+	}
+	if err := f.client.Get(context.TODO(), sourceKey, source); err != nil {
```
levan-m:
```suggestion
	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	defer cancel()
	if err := f.client.Get(ctx, sourceKey, source); err != nil {
```
let's add a timeout since it's direct api server read.

internal/controller/datadogagent/feature/controlplanemonitoring/feature.go
```diff
@@ -6,12 +6,16 @@
 package controlplanemonitoring
 
 import (
+	"context"
 	"fmt"
+	"maps"
```
levan-m, line 11:
```suggestion
	"maps"
	"time"
```
for context change

levan-m (approved):

internal/controller/datadogagent/feature/controlplanemonitoring/feature.go
```diff
@@ -109,6 +117,45 @@ func (f *controlPlaneMonitoringFeature) ManageDependencies(managers feature.Reso
...
+	source := &corev1.Secret{}
+	sourceKey := types.NamespacedName{
+		Namespace: etcdCertsSourceNamespace,
+		Name:      etcdCertsSecretName,
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	if err := f.client.Get(ctx, sourceKey, source); err != nil {
+		f.logger.Info("Unable to copy OpenShift etcd metric client secret automatically", "namespace", sourceKey.Namespace, "name", sourceKey.Name, "error", err)
+		return false
```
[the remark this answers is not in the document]
tbavelier: fixed in 7c146b06f7e0d93e31d97b6a675838dd14dbfd4c

PR #3135, opened 2026-06-12
never let PodTemplateSpec rewrite Host Profiler seccomp init container

theomagellan:
### What does this PR do?

- Fixes `host-profiler-seccomp-setup` init container using the wrong image. `override.PodTemplateSpec` rewrites all init container images to `spec.override.nodeAgent.image`, but `host-profiler-seccomp-setup` must use the profiler image to copy its seccomp profile, which the agent image doesn't have.
The fix skips `host-profiler-seccomp-setup` in the init container rewrite loop.

- Reverts the workaround introduced in #3061 (`resolveHostProfilerImage` reading `spec.override.nodeAgent.image`), which was trying to keep the hash and init container image consistent with each other around the wrong image.

### Motivation

`host-profiler-seccomp-setup` was failing on nightly because it was running from the agent image and trying to copy a file that doesn't exist there.

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan
On test cluster using main:
1. set `spec.override.nodeAgent.image`
2. observed the init container with the wrong image

Repeated these steps with fix included to confirm.

### Checklist


theomagellan:
@codex review

internal/controller/datadogagent/override/podtemplatespec.go
```diff
@@ -83,6 +83,11 @@ func PodTemplateSpec(logger logr.Logger, manager feature.PodTemplateManagers, ov
 		}
 
 		for i, initContainer := range manager.PodTemplateSpec().Spec.InitContainers {
+			// host-profiler-seccomp-setup copies a seccomp profile JSON baked into the profiler image, not the agent image
+			// until the profiler is bundled in the agent.
+			if apicommon.AgentContainerName(initContainer.Name) == apicommon.HostProfilerSeccompSetupContainerName {
+				continue
```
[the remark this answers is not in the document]
theomagellan: The Host Profiler is only supported through annotations with the `image-override-config`. `spec.override.nodeAgent.image.pullPolicy` doesn't apply to the annotation.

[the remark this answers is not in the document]
theomagellan: `spec.override.nodeAgent.image` doesn't point at a `full replacement image` since the Host Profiler is not yet bundled in the agent image; the Host Profiler IS ONLY enabled by using its override annotation which has a higher precedence than `spec.override.nodeAgent.image`.
If the annotation wasn't set, we would be using the profiler's default image, which is the agent's image so it should also be correct for when the profiler is bundled in the agent's image.
This would need to be checked again when time comes, Codex is hard to follow here so I did my best to cover our bases.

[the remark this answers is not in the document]
levan-m: @theomagellan I wonder if this implies that `host-profiler` image override should be skipped similar to host profiler init container. [This](https://github.com/DataDog/datadog-operator/blob/aed954126ecbc443ef77ceb79e793fb905e588f5/internal/controller/datadogagent/feature/hostprofiler/feature.go#L199-L218) will resolve image based on the annotation but if `host-profiler` image override is present resolved image name will be wrong.

levan-m: fyi, [global](https://github.com/DataDog/datadog-operator/blob/aed954126ecbc443ef77ceb79e793fb905e588f5/internal/controller/datadogagent/global/global.go#L314) settings may modify image registry and suffix

levan-m (approved):

internal/controller/datadogagent/override/podtemplatespec.go
theomagellan: 👋 Thank you for this! I added skip checks there as well. See [74695b6](https://github.com/DataDog/datadog-operator/pull/3135/commits/74695b65008b3a03c8884d3e324425f657061301)

internal/controller/datadogagent/override/podtemplatespec.go
```diff
@@ -73,6 +73,10 @@ func PodTemplateSpec(logger logr.Logger, manager feature.PodTemplateManagers, ov
 						JMXEnabled: false,
 					}
 					manager.PodTemplateSpec().Spec.Containers[i].Image = images.OverrideAgentImage(container.Image, otelOverride)
+				} else if containerName == apicommon.HostProfiler {
+					// The Host Profiler only relies on its override annotation image as it is in preview and not yet bundled in
```
Gandem: Would it be worth updating these comments?

theomagellan: Yes! Thanks for bringing these up.

Gandem (approved):

levan-m (approved):

PR #3144, opened 2026-06-15
Align host profiler seccomp image overrides

Gandem:
## Summary

Stacked on #3135. This follows up on the host-profiler seccomp image override review feedback:

- use the same experimental image override semantics when selecting the host-profiler seccomp init image/profile
- keep `host-profiler` and `host-profiler-seccomp-setup` images from being rewritten by `spec.override.nodeAgent.image`
- still apply `spec.override.nodeAgent.image.pullPolicy` to those containers
- keep global registry/FIPS image handling for host-profiler so short-form experimental overrides preserve the expected base registry/tag
- add focused coverage for image alignment through the real mutation order

## Validation

```bash
GOMODCACHE=/tmp/datadog-operator-gomodcache \
GOCACHE=/tmp/datadog-operator-gocache \
go test ./internal/controller/datadogagent/feature/hostprofiler \
        ./internal/controller/datadogagent/override \
        ./internal/controller/datadogagent/global \
        ./internal/controller/datadogagent/experimental
```

### DIFF
--- a/internal/controller/datadogagent/global/global.go
+++ b/internal/controller/datadogagent/global/global.go
@@ -13,7 +13,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	apicommon "github.com/DataDog/datadog-operator/api/datadoghq/common"
 	"github.com/DataDog/datadog-operator/api/datadoghq/v2alpha1"
 	apiutils "github.com/DataDog/datadog-operator/api/utils"
 	"github.com/DataDog/datadog-operator/internal/controller/datadogagent/common"
@@ -313,10 +312,6 @@ func applyGlobalSettings(logger logr.Logger, manager feature.PodTemplateManagers
 func updateContainerImages(config *v2alpha1.GlobalConfig, podTemplateManager feature.PodTemplateManagers) {
 	image := &images.Image{}
 	for i, container := range podTemplateManager.PodTemplateSpec().Spec.Containers {
-		if apicommon.AgentContainerName(container.Name) == apicommon.HostProfiler {
-			// do not override host-profiler image
-			continue
-		}
 		image = images.FromString(container.Image).
 			WithRegistry(*config.Registry).
 			WithFIPS(*config.UseFIPSAgent)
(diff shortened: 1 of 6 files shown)


internal/controller/datadogagent/global/global.go
```diff
@@ -313,10 +312,6 @@ func applyGlobalSettings(logger logr.Logger, manager feature.PodTemplateManagers
 func updateContainerImages(config *v2alpha1.GlobalConfig, podTemplateManager feature.PodTemplateManagers) {
 	image := &images.Image{}
 	for i, container := range podTemplateManager.PodTemplateSpec().Spec.Containers {
-		if apicommon.AgentContainerName(container.Name) == apicommon.HostProfiler {
-			// do not override host-profiler image
-			continue
-		}
```
theomagellan, line 319: Wouldn't that mean getting overwritten by line 319?

internal/controller/datadogagent/global/global.go
```diff
@@ -325,10 +320,6 @@ func updateContainerImages(config *v2alpha1.GlobalConfig, podTemplateManager fea
 	}
 
 	for i, container := range podTemplateManager.PodTemplateSpec().Spec.InitContainers {
-		if apicommon.AgentContainerName(container.Name) == apicommon.HostProfilerSeccompSetupContainerName {
-			// do not override host-profiler-seccomp-setup initContainer image
-			continue
-		}
```
theomagellan, line 331: Same as above

internal/controller/datadogagent/global/global.go
Gandem: I wonder whether we want to be overridden here to respect the global registry overrides. e.g. if the customer is using gcr, ecr, etc. It looks like the annotation will always have a higher priority afterwards if I'm not mistaken.

internal/controller/datadogagent/override/podtemplatespec.go
```diff
@@ -73,10 +73,7 @@ func PodTemplateSpec(logger logr.Logger, manager feature.PodTemplateManagers, ov
 						JMXEnabled: false,
 					}
 					manager.PodTemplateSpec().Spec.Containers[i].Image = images.OverrideAgentImage(container.Image, otelOverride)
-				} else if containerName == apicommon.HostProfiler {
-					// The Host Profiler only relies on its override annotation image
-					continue
-				} else {
+				} else if containerName != apicommon.HostProfiler {
```
theomagellan, line 76: Not doing a `continue` here would mean having the Host Profiler's ImagePullPolicy be modified, do we want this?

Gandem: I think so, the pull policy should align with the rest of the agent components. I'm not seeing a reason not do this, is there any?

theomagellan: Not really, just wanted to make sure we were sure about which agent image configuration knobs apply to us or not. SGTM

PR #3146, opened 2026-06-15
Fix digest-pinned image references being truncated by FromString

carloscastrojumo:
### What does this PR do?

Fixes parsing of digest-pinned image references in `pkg/images`. Since v1.26.0, `FromString` splits the last image path component on every colon and keeps only the first two parts, so an override such as

```yaml
override:
  clusterAgent:
    image:
      name: "eu.gcr.io/datadoghq/cluster-agent:7.80.0@sha256:55504934ebb295875d09c551e9f4e9f0e8540d1c274cde643a04c26dcb3b1a8a"
```

loses the digest hex and is rendered into pods as `eu.gcr.io/datadoghq/cluster-agent:7.80.0@sha256`, which is not a valid image reference. The cluster agent, cluster checks runners, and any rescheduled node agent pods then fail with `InvalidImageName`.

This change splits the digest off before the tag is parsed, carries it through `Image` and `ToString`, and makes an override of the image identity (name or tag) replace any previous digest instead of carrying a stale one over. Digest-only references (`registry/name@sha256:...`) now parse correctly too.

### Motivation

We pin agent images by digest. Upgrading the operator from 1.23.1 to 1.27.1 broke every cluster that applied the upgrade: the regression came in with the tag suffix parsing rewrite in #2605 (first released in v1.26.0). Before that, the parser split on the first colon only, so tag plus digest references survived. Backports to v1.27 and v1.28 would be appreciated.

### Additional Notes

None.

### Minimum Agent Versions

No minimum Agent or Cluster Agent version changes.

* Agent: N/A
* Cluster Agent: N/A

### Describe your test plan

Added unit tests covering digest parsing in `Test_FromString`, `Test_ToString`, `Test_OverrideAgentImage` (including the production repro and digest clearing on tag override) and `Test_AssembleImage`.

- `go test ./pkg/images/...`
- `go test ./internal/controller/datadogagent/override/... ./internal/controller/datadogagent/global/...`

### Checklist

### DIFF
--- a/pkg/images/images.go
+++ b/pkg/images/images.go
@@ -75,6 +75,7 @@ type Image struct {
 	registry string
 	name     string
 	tag      string
+	digest   string
 	isJMX    bool
 	isFIPS   bool
 	isFull   bool
@@ -114,6 +115,14 @@ func (i *Image) WithName(name string) *Image {
 		return i
 	}
 	i.name = name
+	return i
+}
+
+func (i *Image) WithDigest(digest string) *Image {
+	if digest == "" {
+		return i
+	}
+	i.digest = digest
 	return i
 }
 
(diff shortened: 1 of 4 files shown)


davidxia:
Thanks, this is affecting me as well. Looking forward to the release of this fix.

PR #3159, opened 2026-06-18
fix: include version in the operator's Kubernetes client User-Agent

wdhif:
### What does this PR do?

Appends the operator version to the User-Agent of the operator's Kubernetes client. The `rest.Config` backing the controller-runtime manager now sets `UserAgent = "datadog-operator/" + version.Version` (set before `ctrl.NewManager`, so the manager's cache and clients inherit it), instead of a static `"datadog-operator"` with no version.

### Motivation

The operator's client sent a static `datadog-operator` User-Agent with no version, so API-server audit logs and managed control-plane telemetry (GKE Cloud Audit Logs `callerSuppliedUserAgent`, EKS audit logs) could not identify which operator version was issuing calls. Adding the version makes the caller identifiable. Related to FRCONS-815, where the Cluster Agent had the analogous problem (reported as `v0.0.0`).

### Additional Notes

N/A

### Minimum Agent Versions

* Agent: N/A
* Cluster Agent: N/A

### Describe your test plan

- Build the operator image and deploy it to a cluster.
- Confirm the API server observes the versioned User-Agent:
  - kube-apiserver audit log → `userAgent` field, or
  - GKE/EKS control-plane (Cloud) audit logs → `callerSuppliedUserAgent`.
  - Expect `datadog-operator/<version>` instead of bare `datadog-operator`.
- Cross-check the running version via the operator's `--version` output / startup logs, which already print `version.Version`.

### Checklist

- [x] All commits are signed (see: [signing commits][1]) — verified (SSH-signed)

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -282,7 +282,7 @@ func run(opts *options) error {
 	}
 
 	restConfig := ctrl.GetConfigOrDie()
-	restConfig.UserAgent = "datadog-operator"
+	restConfig.UserAgent = "datadog-operator/" + version.Version
 	mgr, err := ctrl.NewManager(restConfig, ctrl.Options{
 		Scheme:                     scheme,
 		Metrics:                    metricsServerOptions,


code-coverage.datadog.yml
```diff
@@ -2,6 +2,7 @@ schema-version: v1
 ignore:
   - "test/"
   - "**/zz_generated.*.go"
+  - "cmd/main.go"
```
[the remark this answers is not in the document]
wdhif: Removed the second commit 👍

cmd/main.go
```diff
@@ -282,7 +282,7 @@ func run(opts *options) error {
 	}
 
 	restConfig := ctrl.GetConfigOrDie()
-	restConfig.UserAgent = "datadog-operator"
+	restConfig.UserAgent = "datadog-operator/" + version.Version
```
tbavelier, line 285: nit: There's a very small risk of getting an empty/malformed `version.Version` in custom builds (someone explicitly providing `VERSION=<malformed>`, e.g. with control characters), so maybe we could add a sanitizer here, but I'm not sure if it's really worth it, as non custom builds will have the proper operator version, and even custom builds will do using the git commit sha, so it can only happen by explicitly crafting a malformed VERSION and providing it.

tbavelier (approved):

cmd/main.go
wdhif: Yeah I think it's not worth it since it would only happens if someone would explicitly malform the version 👍

PR #3162, opened 2026-06-18
[CASCL-1386] (3/11) Display the eviction plan and confirmation prompt

L3n41c:
### What does this PR do?

Implements the human-readable plan display (`printPlan`) and the interactive `y/N` confirmation prompt.

### Motivation

PR #3026 is too large to review as a single change. This is **part 3 of 11** of a stack that splits it into small, individually-reviewable pieces that each build and pass tests on their own. See #3026 for the full feature context and end-to-end QA.

### Additional Notes

**Stacked PR — the base branch is `lenaic/CASCL-1386-evict-02-buildplan`, _not_ `main`**, so the diff shows only this step's change. Implements logic the skeleton (part 1) left stubbed. The code is taken verbatim from #3026 (rebased onto `main`); there are no logic changes versus #3026.

### Minimum Agent Versions

### Describe your test plan

### Checklist

- [x] All commits are signed


cmd/kubectl-datadog/autoscaling/cluster/evict/prompt.go
```diff
@@ -1,15 +1,58 @@
...
+	if ensurePDBs {
+		fmt.Fprintln(streams.Out, "  • Remove the temporary PodDisruptionBudgets created above")
+	}
+	fmt.Fprintln(streams.Out, color.YellowString("\n⚠ This will drain workloads and terminate the underlying instances of non-Datadog node groups."))
+	fmt.Fprintln(streams.Out, color.YellowString("  Verify the Datadog Karpenter NodePool has enough capacity headroom for the migrated pods."))
 }
 
 func promptConfirmation(streams genericclioptions.IOStreams) bool {
-	panic("TODO: promptConfirmation — implemented in PR https://github.com/DataDog/datadog-operator/pull/3162")
+	fmt.Fprint(streams.Out, "\nContinue? (y/N): ")
```
AlexanderYastrebov, line 54: Note: here, to make it strict, it could be a loop in case answer is not one of the supported. Also maybe require answer strictly match one of `yes`, `no` and retry on anything else.

AlexanderYastrebov (approved):

PR #3163, opened 2026-06-18
[CASCL-1386] (4/11) Add evict-legacy-nodes preflight warnings

L3n41c:
### What does this PR do?

Implements the non-blocking preflight warnings (e.g. user Karpenter NodePool weight conflicts) shown before eviction.

### Motivation

PR #3026 is too large to review as a single change. This is **part 4 of 11** of a stack that splits it into small, individually-reviewable pieces that each build and pass tests on their own. See #3026 for the full feature context and end-to-end QA.

### Additional Notes

**Stacked PR — the base branch is `lenaic/CASCL-1386-evict-03-prompt`, _not_ `main`**, so the diff shows only this step's change. Implements logic the skeleton (part 1) left stubbed. The code is taken verbatim from #3026 (rebased onto `main`); there are no logic changes versus #3026.

### Minimum Agent Versions

### Describe your test plan

### Checklist

- [x] All commits are signed


cmd/kubectl-datadog/autoscaling/cluster/evict/preflight.go
```diff
@@ -2,11 +2,75 @@ package evict
...
+	}
+
+	for _, name := range userNPNames {
+		w, ok := weightByName[name]
+		if !ok {
+			continue
+		}
+		if w >= ddMaxWeight {
+			fmt.Fprintln(streams.ErrOut, color.YellowString(
+				"⚠ NodePool %q has spec.weight=%d ≥ Datadog NodePool max weight=%d; evicted pods may be re-scheduled onto a freshly provisioned node from the same user NodePool. Consider raising the Datadog NodePool weight before retrying.",
```
AlexanderYastrebov: How do they increase Datadog NodePool weight? Shouldn't it be automanaged?

AlexanderYastrebov (approved):

cmd/kubectl-datadog/autoscaling/cluster/evict/preflight.go
L3n41c: This check is indeed an extra-care to cover edge cases that shouldn’t happen in nominal cases.
Indeed, my understanding of the `cluster-agent` code is that Datadog-managed node pools always have a weight higher than the targeted node pool: https://github.com/DataDog/datadog-agent/blob/7a5097d31a904a3578f8f108b7c7ae361540f732/pkg/clusteragent/autoscaling/cluster/model/node_pool_internal.go#L154

cmd/kubectl-datadog/autoscaling/cluster/evict/preflight.go
```diff
@@ -2,11 +2,75 @@ package evict
...
+// max, freshly evicted pods may be re-scheduled onto a new node from the
+// same user NodePool, defeating the migration.
+//
+// This is a non-blocking pre-flight warning: it flags a situation that may
+// surprise the operator after the fact but does not prevent eviction. Pure
+// best-effort — failures here are logged but do not abort the run.
+func warnKarpenterWeightConflicts(ctx context.Context, streams genericclioptions.IOStreams, ctrlClient client.Client, targets []Target) {
+	userNPNames := lo.FilterMap(targets, func(t Target, _ int) (string, bool) {
+		return t.Entity, t.Manager == clusterinfo.NodeManagerKarpenter
+	})
```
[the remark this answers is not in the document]
L3n41c: Addressed in `3abc97db`. `warnKarpenterWeightConflicts` no longer derives the user-NodePool set from the eviction targets: it now lists every NodePool in the cluster and warns for any non-Datadog NodePool whose `spec.weight` ≥ the max weight among Datadog-managed NodePools. This covers the cases you flagged — a non-Karpenter target (e.g. `--target=asg/foo`) and user NodePools with no current nodes (absent from the eviction plan) — since Karpenter arbitrates provisioning across all NodePools by weight regardless of what we are draining.

L3n41c:
@codex review

PR #3174, opened 2026-06-18
[CASCL-1386] (7/11) Evict EKS managed node groups

L3n41c:
### What does this PR do?

Implements EKS managed node group eviction (scale-to-zero then wait; EKS drains pods server-side, so it needs no client-side drain primitives).

### Motivation

PR #3026 is too large to review as a single change. This is **part 7 of 11** of a stack that splits it into small, individually-reviewable pieces that each build and pass tests on their own. See #3026 for the full feature context and end-to-end QA.

### Additional Notes

**Stacked PR — the base branch is `lenaic/CASCL-1386-evict-06-ec2-instance-id-helper`, _not_ `main`**, so the diff shows only this step's change. Implements an evictor the skeleton (part 1) left stubbed. The code is taken verbatim from #3026 (rebased onto `main`); there are no logic changes versus #3026.

### Minimum Agent Versions

### Describe your test plan

### Checklist

- [x] All commits are signed


cmd/kubectl-datadog/autoscaling/cluster/evict/eks_mng.go
```diff
@@ -3,18 +3,64 @@ package evict
...
+		log.Printf("[dry-run] would scale EKS node group %s/%s to min=desired=0 (max preserved) and wait for EKS to drain", clusterName, nodegroupName)
+		return nil
+	}
+	if _, err := eksAPI.UpdateNodegroupConfig(ctx, &eks.UpdateNodegroupConfigInput{
+		ClusterName:   awssdk.String(clusterName),
+		NodegroupName: awssdk.String(nodegroupName),
+		ScalingConfig: &ekstypes.NodegroupScalingConfig{
+			MinSize:     awssdk.Int32(0),
+			DesiredSize: awssdk.Int32(0),
+		},
```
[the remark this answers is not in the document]
hmahmood: Is this correct? I guess we are draining the nodes before this code is called? Maybe note that in the comments here, or add a check that ensures nodes are empty.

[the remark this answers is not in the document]
L3n41c: Addressed in 409d6326.

`evictEKSManagedNodeGroup` now cordons and drains every node through the Kubernetes Eviction API (PDB-aware) **before** calling `UpdateNodegroupConfig` to scale the group to zero, so the temporary PDBs are honored.

cmd/kubectl-datadog/autoscaling/cluster/evict/eks_mng.go
```diff
@@ -3,18 +3,84 @@ package evict
...
 	UpdateNodegroupConfig(ctx context.Context, in *eks.UpdateNodegroupConfigInput, opts ...func(*eks.Options)) (*eks.UpdateNodegroupConfigOutput, error)
 }
 
-func evictEKSManagedNodeGroup(ctx context.Context, eksAPI EKSManagedNodeGroupAPI, clientset kubernetes.Interface, clusterName, nodegroupName string, drainOpts nodeDrainOptions) error {
-	panic("TODO: evictEKSManagedNodeGroup — implemented in PR https://github.com/DataDog/datadog-operator/pull/3174")
+func evictEKSManagedNodeGroup(ctx context.Context, eksAPI EKSManagedNodeGroupAPI, clientset kubernetes.Interface, clusterName, nodegroupName string, nodes []string, drainOpts nodeDrainOptions) error {
+	cordoned, errs := cordonNodes(ctx, clientset, nodes, drainOpts.DryRun)
+	for _, node := range cordoned {
+		if err := drainNode(ctx, clientset, node.Name, drainOpts); err != nil {
+			errs = append(errs, fmt.Errorf("drain node %s: %w", node.Name, err))
```
AlexanderYastrebov, line 30: Maybe use errors.Join here.

AlexanderYastrebov (approved):

cmd/kubectl-datadog/autoscaling/cluster/evict/eks_mng.go
L3n41c: The error slice folding is done just below: https://github.com/DataDog/datadog-operator/pull/3174/changes#diff-6b899156345b292b05d4951862034449f4a5413e7b2506f88fb821db60bf90bdR37

cmd/kubectl-datadog/autoscaling/cluster/evict/eks_mng.go
```diff
@@ -3,18 +3,84 @@ package evict
...
-	DescribeNodegroup(ctx context.Context, in *eks.DescribeNodegroupInput, opts ...func(*eks.Options)) (*eks.DescribeNodegroupOutput, error)
 	UpdateNodegroupConfig(ctx context.Context, in *eks.UpdateNodegroupConfigInput, opts ...func(*eks.Options)) (*eks.UpdateNodegroupConfigOutput, error)
 }
 
-func evictEKSManagedNodeGroup(ctx context.Context, eksAPI EKSManagedNodeGroupAPI, clientset kubernetes.Interface, clusterName, nodegroupName string, drainOpts nodeDrainOptions) error {
-	panic("TODO: evictEKSManagedNodeGroup — implemented in PR https://github.com/DataDog/datadog-operator/pull/3174")
+func evictEKSManagedNodeGroup(ctx context.Context, eksAPI EKSManagedNodeGroupAPI, clientset kubernetes.Interface, clusterName, nodegroupName string, nodes []string, drainOpts nodeDrainOptions) error {
+	cordoned, errs := cordonNodes(ctx, clientset, nodes, drainOpts.DryRun)
+	for _, node := range cordoned {
+		if err := drainNode(ctx, clientset, node.Name, drainOpts); err != nil {
```
AlexanderYastrebov, line 29: Does it make sense instead of cordoning all and then draining each have a primitive that cordones+drains a single node?

L3n41c: If we cordon and then drain node by node, then the evicted pods might be scheduled on other nodes of the same node group that will be drained just few seconds later.
This is inefficient.
Cordoning all the nodes of a node group first it a way to guarantee that evicted pods cannot land on nodes that are about to be drained just after.

AlexanderYastrebov: Right, indeed.

PR #3175, opened 2026-06-18
[CASCL-1386] (8/11) Add node drain primitives and evict ASG nodes

L3n41c:
### What does this PR do?

Introduces the shared node-drain primitives — node cordoning (`cordonNodes`), pod eviction/draining (`drainNode`) and the pod-eligibility predicates (`shouldSkipEviction` …) — together with their first consumer, ASG eviction (cordon, drain, terminate the instances, lock the ASG to 0).

### Motivation

PR #3026 is too large to review as a single change. This is **part 8 of 11** of a stack that splits it into small, individually-reviewable pieces that each build and pass tests on their own. See #3026 for the full feature context and end-to-end QA.

### Additional Notes

**Stacked PR — the base branch is `lenaic/CASCL-1386-evict-07-eks-mng`, _not_ `main`**, so the diff shows only this step's change. The drain primitives are bundled with the ASG evictor on purpose: `golangci-lint`'s `unused` linter rejects an unexported helper that has no caller in the same commit, so the primitives land in the same PR as the first evictor that calls them. Parts 9–10 reuse them. The code is taken verbatim from #3026 (rebased onto `main`); there are no logic changes versus #3026.

### Minimum Agent Versions

### Describe your test plan

### Checklist

- [x] All commits are signed

### DIFF
--- a/cmd/kubectl-datadog/autoscaling/cluster/evict/cordon.go
+++ b/cmd/kubectl-datadog/autoscaling/cluster/evict/cordon.go
@@ -1,0 +1,69 @@
+package evict
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/util/retry"
+)
+
+// cordonNodes marks every node in the group Unschedulable up front, before any
+// of them is drained. A pod evicted from one node is then never rescheduled
+// onto another node of the same group that is itself about to be drained.
+//
+// It returns the Node objects that are now cordoned and safe to drain, so the
+// caller can read their immutable fields (e.g. providerID) without a second
+// Get. A node that is already gone is skipped silently (absent from both
+// return values); a node that fails to cordon is recorded in errs and left out
+// of cordoned so the caller never drains a node that can still receive pods.
+func cordonNodes(ctx context.Context, clientset kubernetes.Interface, nodes []string, dryRun bool) (cordoned []*corev1.Node, errs []error) {
+	for _, nodeName := range nodes {
+		if node, err := cordonNode(ctx, clientset, nodeName, dryRun); err != nil {
+			errs = append(errs, fmt.Errorf("cordon node %s: %w", nodeName, err))
+		} else if node != nil {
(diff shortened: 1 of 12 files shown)


cmd/kubectl-datadog/autoscaling/cluster/evict/cordon.go
```diff
@@ -0,0 +1,71 @@
...
+// rescheduled onto another node of the same group that is itself about to be
+// drained.
+//
+// It returns the Node objects that are now cordoned and safe to drain, so the
+// caller can read their immutable fields (e.g. providerID) without a second
+// Get. A node that is already gone is skipped silently (absent from both
+// return values); a node that fails to cordon is recorded in errs and left out
+// of cordoned so the caller never drains a node that can still receive pods.
+func cordonNodes(ctx context.Context, clientset kubernetes.Interface, nodes []string, dryRun bool) (cordoned []*corev1.Node, errs []error) {
+	for _, nodeName := range nodes {
```
hmahmood, line 25: Maybe use https://pkg.go.dev/golang.org/x/sync/errgroup here to parallelize/collect errors?

L3n41c: The benefit of parallelizing the cordoning in term of wall clock time would be minimal because cordoning the nodes is blazing fast compared to draining their pods.
And I’d prefer to keep nodes draining sequential in this first iteration to:
* minimize the disruption caused to the customer’s workload,
* minimize the stress caused to the API server.
Thanks to the PDB protection, it should theoretically be possible to parallelize the drain of all the nodes, but I’m a bit afraid of the load it might generate on the API server.

Using `errgroup` would also change the way errors are managed.
Current behavior is best-effort: if cordoning a single node fails, we register that error to surface it later, but we keep on cordoning the remaining nodes.
`errgroup` behavior is more fail-early: it makes all the goroutines abort as soon as one of them returns an error.
In order to avoid too many back and forth with the user, I think it’s better to clean up as much nodes as possible before asking the user to manually fix something and relaunch the script.

AlexanderYastrebov (approved):

cmd/kubectl-datadog/autoscaling/cluster/evict/cordon.go
AlexanderYastrebov: > errgroup behavior is more fail-early

You can code worker to not return error but, yeah, I agree its better start simple and then optimize the bottlenecks.

hmahmood (approved):

PR #3176, opened 2026-06-18
[CASCL-1386] (9/11) Evict standalone EC2 nodes

L3n41c:
### What does this PR do?

Implements standalone EC2 node eviction: cordon, drain and terminate the instances (reusing the drain primitives from part 8).

### Motivation

PR #3026 is too large to review as a single change. This is **part 9 of 11** of a stack that splits it into small, individually-reviewable pieces that each build and pass tests on their own. See #3026 for the full feature context and end-to-end QA.

### Additional Notes

**Stacked PR — the base branch is `lenaic/CASCL-1386-evict-08-drain-and-asg`, _not_ `main`**, so the diff shows only this step's change. Reuses the drain primitives introduced in part 8. The code is taken verbatim from #3026 (rebased onto `main`); there are no logic changes versus #3026.

### Minimum Agent Versions

### Describe your test plan

### Checklist

- [x] All commits are signed

### DIFF
--- a/cmd/kubectl-datadog/autoscaling/cluster/evict/standalone.go
+++ b/cmd/kubectl-datadog/autoscaling/cluster/evict/standalone.go
@@ -2,9 +2,14 @@ package evict
 
 import (
 	"context"
+	"errors"
+	"fmt"
+	"log"
 
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	"k8s.io/client-go/kubernetes"
+
+	commonaws "github.com/DataDog/datadog-operator/cmd/kubectl-datadog/autoscaling/cluster/common/aws"
 )
 
 type EC2API interface {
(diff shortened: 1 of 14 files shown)


cmd/kubectl-datadog/autoscaling/cluster/evict/standalone.go
```diff
@@ -2,15 +2,48 @@ package evict
...
 }
 
 func evictStandalone(ctx context.Context, clientset kubernetes.Interface, ec2API EC2API, nodes []string, drainOpts nodeDrainOptions) error {
-	panic("TODO: evictStandalone — implemented in PR https://github.com/DataDog/datadog-operator/pull/3176")
+	cordoned, errs := cordonNodes(ctx, clientset, nodes, drainOpts.DryRun)
+	for _, node := range cordoned {
+		nodeName := node.Name
+		id, hasInstanceID := commonaws.ExtractEC2InstanceID(node)
+		if !hasInstanceID {
+			log.Printf("Warning: node %s has unexpected providerID %q; cannot terminate the underlying instance", nodeName, node.Spec.ProviderID)
```
AlexanderYastrebov: nit: maybe move this after the drain since it does not need either id? Otherwise it will show warning and then hand on long-running drain...

AlexanderYastrebov (approved):

cmd/kubectl-datadog/autoscaling/cluster/evict/standalone.go
L3n41c: Definitively makes a lot of sense. I’ve just moved that part in 34876a8cffa179afb3b6a89791d70e18c40e8aec.

PR #3177, opened 2026-06-18
[CASCL-1386] (10/11) Evict Karpenter user NodePools

L3n41c:
### What does this PR do?

Implements Karpenter user NodePool eviction: cordon and drain (Karpenter reclaims the emptied instances).

### Motivation

PR #3026 is too large to review as a single change. This is **part 10 of 11** of a stack that splits it into small, individually-reviewable pieces that each build and pass tests on their own. See #3026 for the full feature context and end-to-end QA.

### Additional Notes

**Stacked PR — the base branch is `lenaic/CASCL-1386-evict-09-standalone`, _not_ `main`**, so the diff shows only this step's change. Reuses the drain primitives introduced in part 8. The code is taken verbatim from #3026 (rebased onto `main`); there are no logic changes versus #3026.

### Minimum Agent Versions

### Describe your test plan

### Checklist

- [x] All commits are signed

### DIFF
--- a/cmd/kubectl-datadog/autoscaling/cluster/evict/karpenter_user.go
+++ b/cmd/kubectl-datadog/autoscaling/cluster/evict/karpenter_user.go
@@ -2,10 +2,22 @@ package evict
 
 import (
 	"context"
+	"errors"
+	"fmt"
+	"log"
 
 	"k8s.io/client-go/kubernetes"
 )
 
 func evictKarpenterUserNodePool(ctx context.Context, clientset kubernetes.Interface, nodePoolName string, nodes []string, drainOpts nodeDrainOptions) error {
-	panic("TODO: evictKarpenterUserNodePool — implemented in PR https://github.com/DataDog/datadog-operator/pull/3177")
+	cordoned, errs := cordonNodes(ctx, clientset, nodes, drainOpts.DryRun)
+	for _, node := range cordoned {
+		if err := drainNode(ctx, clientset, node.Name, drainOpts); err != nil {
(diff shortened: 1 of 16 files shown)


cmd/kubectl-datadog/autoscaling/cluster/evict/karpenter_user.go
```diff
@@ -2,10 +2,22 @@ package evict
...
 
 func evictKarpenterUserNodePool(ctx context.Context, clientset kubernetes.Interface, nodePoolName string, nodes []string, drainOpts nodeDrainOptions) error {
-	panic("TODO: evictKarpenterUserNodePool — implemented in PR https://github.com/DataDog/datadog-operator/pull/3177")
+	cordoned, errs := cordonNodes(ctx, clientset, nodes, drainOpts.DryRun)
+	for _, node := range cordoned {
+		if err := drainNode(ctx, clientset, node.Name, drainOpts); err != nil {
+			errs = append(errs, fmt.Errorf("drain node %s: %w", node.Name, err))
+		}
+	}
+	if !drainOpts.DryRun && len(errs) == 0 {
```
AlexanderYastrebov, line 19: Should this log rather be at the call site?

AlexanderYastrebov (approved):

cmd/kubectl-datadog/autoscaling/cluster/evict/karpenter_user.go
L3n41c: Good point.
The “policy” that should be applied to all node group types is to:
* in case of success, log directly what was done inside the `evictXXXXX` function.
* in case of failure, return the errors and let the caller log them.

PR #3179, opened 2026-06-18
fix podstartup

levan-m:
### What does this PR do?

If Operator can't read the etcd secret agent pod gets stuck in `ContainerCreating`. To avoid we check if secret exists before mounting. Unmounted secret leads to check failure but everything else work fine.

### Motivation

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

To reproduce the failure
```
k scale deploy etcd-operator -n openshift-etcd-operator --replicas 0
k delete secret etcd-metric-client -n openshift-etcd-operator
```

### Checklist

### DIFF
--- a/internal/controller/datadogagent/feature/controlplanemonitoring/feature.go
+++ b/internal/controller/datadogagent/feature/controlplanemonitoring/feature.go
@@ -13,7 +13,6 @@ import (
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
@@ -50,7 +49,6 @@ type controlPlaneMonitoringFeature struct {
 	eksConfigMapName       string
 	client                 client.Reader
 
-	etcdSecretChecked bool
 	etcdSecretPresent bool
 }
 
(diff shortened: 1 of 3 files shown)


internal/controller/datadogagent/feature/controlplanemonitoring/feature.go
```diff
@@ -306,30 +310,63 @@ func (f *controlPlaneMonitoringFeature) ManageSingleContainerNodeAgent(managers
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	key := types.NamespacedName{Namespace: f.owner.GetNamespace(), Name: etcdCertsSecretName}
+	if err := f.client.Get(ctx, key, &corev1.Secret{}); err != nil {
+		if !apierrors.IsNotFound(err) {
+			f.logger.V(1).Info("Unable to verify OpenShift etcd metric client secret; skipping etcd cert mount", "namespace", key.Namespace, "name", key.Name, "error", err)
+		}
+		return false
+	}
```
tbavelier: This whole loop uses the direct api read the same way `copyOpenShiftEtcdSecret` does. However, unlike the copy function, this is not needed I feel like because we're checking existence within the "store"/the owner namespace that we are watching. We can possibly check the store or add this flag in the existing helpers that already go through the direct api call: e.g. https://github.com/DataDog/datadog-operator/pull/3180

tbavelier (changes requested):
LGTM overall, is indeed a useful defensive measure. Unit tests are failing also before merging
I don't think we need a new direct api call but could be wrong. Proposing to add the flag as part of existing helpers or something else, what do you think ?

PR #3183, opened 2026-06-22
Add shared dependencies

khewonc:
### What does this PR do?

Add shared dependencies. Dependencies that can be configured by both DDA and DDAIs and are 1 per cluster are "shared" dependencies since we can't create 1 per DDA/DDAI. These are things like local service for APM/DSD and eventually seccomp profile configmap

### Motivation

https://datadoghq.atlassian.net/browse/CONTP-1762

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Create a DDA and DAP with hostports set to different values:
```yaml
# dda
    features:
      apm:
        enabled: true
        hostPortConfig:
          enabled: true
          hostPort: 8126
```

```yaml
# dap
    features:
      apm:
        enabled: true
        hostPortConfig:
          enabled: true
          hostPort: 8127
```

The DDA value `8126` should be in the resulting service (`kubectl get svc <local-agent-svc-name>`). In the operator logs, you will see this error: `local Agent Service port xxxx conflicts with existing port` where xxxx is the conflicting port

### Checklist

### DIFF
--- a/internal/controller/datadogagent/dependencies.go
+++ b/internal/controller/datadogagent/dependencies.go
@@ -7,13 +7,11 @@ package datadogagent
 
 import (
 	"context"
-	"errors"
 	"fmt"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
-	ctrl "sigs.k8s.io/controller-runtime"
 
 	v1alpha1 "github.com/DataDog/datadog-operator/api/datadoghq/v1alpha1"
 	"github.com/DataDog/datadog-operator/api/datadoghq/v2alpha1"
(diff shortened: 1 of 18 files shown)


internal/controller/datadogagent/dependencies.go
```diff
@@ -93,6 +104,31 @@ func (r *Reconciler) manageDDADependenciesWithDDAI(ctx context.Context, logger l
...
+	for _, ddai := range ddais {
+		if err := feature.ApplyDDASharedDependencies(dda, &dda.Spec, ddai, &ddai.Spec, managers); err != nil {
+			if errors.Is(err, merger.ErrServicePortConflict) {
+				logger.Info(
+					"Ignoring conflicting DDA shared Service port",
+					"datadogagentinternal_namespace", ddai.Namespace,
+					"datadogagentinternal_name", ddai.Name,
+					"reason", err.Error(),
+				)
+				continue
```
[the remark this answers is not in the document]
khewonc: Instead of this, we validate hostport to see if profiles conflict. If the profile conflicts, we mark as invalid and don't apply it. Thus only non-conflicting profiles should have dependencies applied. If there is still a conflict, we error and prevent further reconciliation

internal/controller/datadogagent/feature/apm/feature.go
levan-m, line 44: I guess the asymmetry with otel, dsd, otlp feature is due to the fact that we don't allow feature overrides for those in DAP right? once we do we will also add overlays and validation before DDAI creation?

levan-m (approved):

internal/controller/datadogagent/feature/apm/feature.go
khewonc: Yes, that's right

PR #3185, opened 2026-06-23
Do not set process-agent AppArmor for NPM fallback

tbavelier:
### What does this PR do?

Removes the `container.apparmor.security.beta.kubernetes.io/process-agent: unconfined` annotation added by the NPM/USM non-direct-send fallback path.

`process-agent` is still created and configured when CNM/USM direct-send is disabled or unavailable, but only `system-probe` keeps its existing AppArmor annotation.

### Motivation

The process-agent AppArmor annotation was introduced while enabling CNM direct-send by default: https://github.com/DataDog/datadog-operator/pull/3045. That changed the legacy fallback rendering even though this annotation was not previously emitted and the Helm chart does not add an AppArmor profile for `process-agent`.

On GKE Autopilot, the extra annotation causes the Datadog daemonset to be rejected by the WorkloadAllowlist because the allowlist does not permit a container-level AppArmor profile for `process-agent`.

### Additional Notes

This is intentionally not an Autopilot-specific workaround. It restores the previous operator behavior for process-agent while keeping the existing system-probe AppArmor behavior unchanged.

### Minimum Agent Versions

* Agent: N/A
* Cluster Agent: N/A

### Describe your test plan

```console
go test ./internal/controller/datadogagent/feature/npm ./internal/controller/datadogagent/feature/usm ./internal/controller/datadogagent/feature/...
```

### Checklist

[1]: https://docs.github.com/en/authentication/managing-commits/signing-commits

### DIFF
--- a/internal/controller/datadogagent/feature/npm/feature.go
+++ b/internal/controller/datadogagent/feature/npm/feature.go
@@ -112,9 +112,6 @@ func (f *npmFeature) ManageNodeAgent(managers feature.PodTemplateManagers) error
 
 	// annotations
 	managers.Annotation().AddAnnotation(common.SystemProbeAppArmorAnnotationKey, common.SystemProbeAppArmorAnnotationValue)
-	if !f.directSend {
-		managers.Annotation().AddAnnotation(common.AppArmorAnnotationKey+"/"+string(apicommon.ProcessAgentContainerName), "unconfined")
-	}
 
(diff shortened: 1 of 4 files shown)


Janaeq (approved):
LGTM!

brycekahle (approved):

khewonc (approved):

PR #3205, opened 2026-06-30
[CONTP-1614] Add e2e tests for untaint controller

adel121:
### What does this PR do?

Adds an end-to-end test suite for the operator's Untaint controller. The suite spins up a kind cluster with a pre-tainted worker node (carrying the `agent.datadoghq.com/not-ready` startup taint), deploys the operator with the untaint controller enabled, and verifies the full onboarding flow: a workload pinned to the tainted node stays `Pending`, and only schedules once the taint is removed.

It covers two modes:

Agent-only (`--untaintControllerEnabled=true`): the taint is removed once the node Agent is Ready.
Wait-for-CSI (`--untaintControllerWaitForCSIDriver=true`): the taint persists until both the node Agent and the CSI node-server are Ready.
Each mode has a local kind entrypoint (for dev) and an AWS kind-on-VM entrypoint (for CI).

Supporting changes:

Adds `WithKindWorkerNodes` to the operator's kind provisioner so worker nodes can be labeled and pre-tainted at cluster creation, and bumps the e2e-framework / fakeintake dependency to the version that exposes it.
Adds a dedicated, blocking `e2e_untaint` CI job (and a `make e2e-untaint-tests` target) pinned to a single recent Kubernetes version.

### Motivation

Validate the taint-based onboarding flow on a real cluster rather than relying only on unit tests, so we can catch regressions in the operator behavior that actually matters to users: workloads should not schedule onto a node until the Agent (and optionally the CSI node-server) is ready there.

### Additional Notes

* The CI job runs automatically only when untaint-related code or the suite changes; otherwise it is a manual button. It runs both the agent-only and wait-for-CSI AWS variants sequentially, pinned to K8s 1.30 (the controller targets 1.27+).

* Both flows were validated locally against datadog/operator:main and pass. The AWS kind-on-VM variants share the same suite logic and differ only in the provisioner, so the first CI run is the real check of the AWS provisioning path.

* Local runs use one kind cluster at a time (the two local variants share host ports), so tear down one before running the other.

### Describe your test plan

E2E tests.

### Checklist


adel121:
@codex

adel121:
@codex

adel121:
@codex

.gitlab-ci.yml
```diff
@@ -487,6 +487,60 @@ e2e_autoscaling:
...
+    # Run automatically on changes to untaint controller / suite code
+    - if: $CI_COMMIT_BRANCH
+      changes:
+        paths:
+          - "internal/controller/untaint_controller.go"
+          - "internal/controller/datadogagent/component/agent/untaint_toleration.go"
+          - "internal/controller/datadogcsidriver/**/*"
+          - "pkg/untaint/**/*"
+          - "test/e2e/tests/untaint_suite/**/*"
+          - "test/e2e/provisioners/kind.go"
```
levan-m, line 535:
```suggestion
          - "test/e2e/provisioners/kind.go"
        compare_to: "refs/heads/main"                 
```
job wasn't triggered by the CI and had to trigger manually. Unless that was deliberate, we are likely missing ^^

.gitlab-ci.yml
```diff
@@ -487,6 +487,60 @@ e2e_autoscaling:
...
+    # trigger_e2e_operator_image (.on_run_e2e_base); without it the catch-all
+    # manual rule below would add this job on *.md-only pipelines while its
+    # required image job is skipped, which GitLab rejects as a missing need.
+    - if: $CI_COMMIT_BRANCH
+      changes:
+        paths:
+          - "*.md"
+        compare_to: "refs/heads/main"
+      when: never
+    # Disable on Conductor-triggered jobs (ex: nightly) — must be first so it wins
```
levan-m: either comment is stale or this should be first?

.gitlab-ci.yml
```diff
@@ -487,6 +487,60 @@ e2e_autoscaling:
...
+    TARGET_IMAGE: $E2E_DOCKER_REGISTRY:${CI_PIPELINE_ID}-${CI_COMMIT_SHORT_SHA}
+  rules:
+    # Skip on docs-only changes. This mirrors the skip rule on
+    # trigger_e2e_operator_image (.on_run_e2e_base); without it the catch-all
+    # manual rule below would add this job on *.md-only pipelines while its
+    # required image job is skipped, which GitLab rejects as a missing need.
+    - if: $CI_COMMIT_BRANCH
+      changes:
+        paths:
+          - "*.md"
```
levan-m, line 517: iirc this will skip PRs which have mix of markdowns and code changes, maybe we can drop this similar to other e2e tests.

adel121: The problem is that the e2e test needs the image to be built, which is triggered in a previous job (`trigger_e2e_operator_image`) that has the same `never` rule. 

If we drop this as you suggested, the job will fail whenever a change mixes both `md` and non- `md` files. 

Added a comment in the code to clarify this. 

I think this should still be fixed, but it is a problem in the `trigger_e2e_operator_image`, I prefer having a separate PR to fix it because it may impact other pipeline jobs.

levan-m (approved):

PR #3207, opened 2026-07-01
[CASCL-1386] (6½/11) Implement shared node cordon and drain primitives

L3n41c:
### What does this PR do?

Introduces the shared node-cordon and pod-eviction/drain primitives used by the `evict-legacy-nodes` command's per-manager evictors (EKS managed node groups, ASG, Karpenter, standalone):

- `cordonNodes`/`cordonNode` — idempotent, conflict-retrying cordon that treats an already-gone node as a silent skip and returns the cordoned `Node` objects.
- `drainNode` — evicts every evictable pod (skipping mirror, DaemonSet, completed and terminating pods), then waits for the node to empty.
- `evictPodWithRetry` — PDB-aware eviction that retries on 429 and treats 404 as success.
- `listPodsOnNode`, `waitForNodeEmpty`, and the pod-classification predicates (`podOccupiesNode`/`shouldSkipEviction`/`isMirrorPod`/`isDaemonSetPod`/`isCompleted`).

### Motivation

Every per-manager evictor needs these primitives. Extracting them into a dedicated PR at the base of the stack keeps each downstream PR focused on its manager-specific logic and lets them all build on a single, reviewed implementation.

### Additional Notes

- Part of the **CASCL-1386** stacked series. This PR sits at the base; the per-manager evictors (starting with EKS managed node groups) rebase on top and become the first production callers.
- golangci-lint runs with `[run] tests = false`, so unit-test usage does not count toward the `unused` linter and there is no production caller yet in this PR. A small, documented `var ( _ = drainNode; _ = cordonNodes )` block keeps `unused` quiet; it is removed by the next PR in the stack when `evictEKSManagedNodeGroup` becomes the first caller.

### Minimum Agent Versions

N/A — `kubectl-datadog` plugin only; no Datadog Agent or Cluster Agent change.

### Describe your test plan

Unit tests (`go test ./cmd/kubectl-datadog/autoscaling/cluster/evict/`) cover: cordon (idempotent / conflict-retry / already-gone / dry-run), pod eviction (429 retry, 404 success, non-retryable error, 429 exhaustion, zero-interval guard), node-empty polling (empty / becomes-empty / timeout / list-error), pod classification, and the `spec.nodeName` field-selector scoping of `listPodsOnNode`. `make lint` is clean.

### Checklist

### DIFF
--- a/cmd/kubectl-datadog/autoscaling/cluster/evict/cordon.go
+++ b/cmd/kubectl-datadog/autoscaling/cluster/evict/cordon.go
@@ -1,0 +1,69 @@
+package evict
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/util/retry"
+)
+
+// cordonNodes marks every node in the group Unschedulable up front, before any
+// of them is drained. A pod evicted from one node is then never rescheduled
+// onto another node of the same group that is itself about to be drained.
+//
+// It returns the Node objects that are now cordoned and safe to drain, so the
+// caller can read their immutable fields (e.g. providerID) without a second
+// Get. A node that is already gone is skipped silently (absent from both
+// return values); a node that fails to cordon is recorded in errs and left out
+// of cordoned so the caller never drains a node that can still receive pods.
+func cordonNodes(ctx context.Context, clientset kubernetes.Interface, nodes []string, dryRun bool) (cordoned []*corev1.Node, errs []error) {
+	for _, nodeName := range nodes {
+		if node, err := cordonNode(ctx, clientset, nodeName, dryRun); err != nil {
+			errs = append(errs, fmt.Errorf("cordon node %s: %w", nodeName, err))
+		} else if node != nil {
+			cordoned = append(cordoned, node)
+		}
+	}
+	return cordoned, errs
+}
+
+// cordonNode marks the node Unschedulable and returns the resulting Node so the
+// caller can read immutable fields (e.g. providerID) without a second Get. It
+// returns (nil, nil) when the node is already gone: there is nothing to
+// schedule onto a deleted node, and a re-run rediscovers any surviving nodes.
+// The Get + mutate + Update is wrapped in RetryOnConflict to survive races
+// against kubelet, the scheduler, and any other controller that mutates Node
+// objects.
+func cordonNode(ctx context.Context, clientset kubernetes.Interface, name string, dryRun bool) (*corev1.Node, error) {
+	var cordoned *corev1.Node
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		node, err := clientset.CoreV1().Nodes().Get(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return nil // node already gone: nothing to schedule on it
+			}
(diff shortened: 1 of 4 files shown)


cmd/kubectl-datadog/autoscaling/cluster/evict/cordon.go
```diff
@@ -0,0 +1,69 @@
...
+// of cordoned so the caller never drains a node that can still receive pods.
+func cordonNodes(ctx context.Context, clientset kubernetes.Interface, nodes []string, dryRun bool) (cordoned []*corev1.Node, errs []error) {
+	for _, nodeName := range nodes {
+		if node, err := cordonNode(ctx, clientset, nodeName, dryRun); err != nil {
+			errs = append(errs, fmt.Errorf("cordon node %s: %w", nodeName, err))
+		} else if node != nil {
+			cordoned = append(cordoned, node)
+		}
+	}
+	return cordoned, errs
```
AlexanderYastrebov, line 32: Maybe use https://pkg.go.dev/errors#Join here?

cmd/kubectl-datadog/autoscaling/cluster/evict/cordon_test.go
```diff
@@ -0,0 +1,134 @@
...
+		// updateNotFound makes the Update return NotFound, simulating the node
+		// being deleted between the Get and the Update.
+		updateNotFound bool
+		dryRun         bool
+		// wantUpdateCalls is the minimum number of Update invocations
+		// expected on the Nodes endpoint.
+		wantMinUpdateCalls int
+		wantUnschedulable  bool
+		// wantNilNode expects cordonNode to return a nil Node (the node is gone).
+		wantNilNode bool
```
AlexanderYastrebov, line 35: Note: for test suites which vary greatly in setup/behaviour it might be better to use nested t.Runs instead of table tests.

cmd/kubectl-datadog/autoscaling/cluster/evict/evict_pods.go
```diff
@@ -9,3 +27,149 @@ type nodeDrainOptions struct {
...
+	}
+	pods, err := listPodsOnNode(ctx, clientset, nodeName)
+	if err != nil {
+		return fmt.Errorf("failed to list pods on node %s: %w", nodeName, err)
+	}
+	for _, p := range pods {
+		if shouldSkipEviction(&p) {
+			continue
+		}
+		if err := evictPodWithRetry(ctx, clientset, &p, opts.EvictionTimeout, opts.PollInterval); err != nil {
```
AlexanderYastrebov, line 63: Note: here it would add up retry latency on each pod with pdb, the alternative approach could be to make several passes over the pod list to amortize that.

AlexanderYastrebov (approved):

cmd/kubectl-datadog/autoscaling/cluster/evict/cordon.go
L3n41c: It’s actually convenient to keep a slice at this point as it allows to append other errors at the call site like here: https://github.com/DataDog/datadog-operator/pull/3174/changes#diff-6b899156345b292b05d4951862034449f4a5413e7b2506f88fb821db60bf90bdR27-R38
Thanks to that, we can have, at the end of the cordon+drain sequence a slice of errors with exactly one error per node which couldn’t be either cordoned or drained.
It can potentially be interesting in the future to know precisely how many nodes failed to be cordoned or drained.

cmd/kubectl-datadog/autoscaling/cluster/evict/evict_pods.go
L3n41c: Agreed that the current algorithm is very naive and slow.
Processing node by node is also sub-optimal.
In a future version, I’d like to have a queue of “pods to evict” that would allow to constantly have `x` pods being evicted at a time even if they are spread on different nodes.

PR #3213, opened 2026-07-02
Add GKE COS NVIDIA library mounts

gjulianm:
### What does this PR do?

Mounts GKE COS NVIDIA driver libraries into the Agent containers when GPU monitoring runs on GKE COS nodes.

### Motivation

GPU monitoring needs these libraries on GKE COS, where NVIDIA driver libraries live under `/home/kubernetes/bin/nvidia/lib64`.

### Additional Notes

Mirrors DataDog/helm-charts#2764 for the Operator.

The mount is applied for the `gke-cos` provider. The operator auto-detects this from the node label `cloud.google.com/gke-os-distribution=cos`, but it can also be forced with the annotation on the DatadogAgent:

```yaml
metadata:
  annotations:
    datadoghq.com/provider: gke-cos
```

### Minimum Agent Versions

* Agent: N/A
* Cluster Agent: N/A

### Describe your test plan

`go test ./internal/controller/datadogagent/feature/gpu`

Validated end-to-end on a GKE COS test cluster with a Tesla T4 node:
- provider forced via the `datadoghq.com/provider: gke-cos` annotation
- generated DaemonSet mounts `/home/kubernetes/bin/nvidia/lib64` at `/host/run/nvidia/driver/usr/lib/x86_64-linux-gnu` on the core-agent and system-probe containers
- system-probe finds the NVML library at the mounted path
- with `features.gpu.privilegedMode: true`, `features.gpu.patchCgroupPermissions: true` and `override.nodeAgent.hostPID: true` (which should be there by default with GPU monitoring enabled, will be addressed in a separate PR), the core-agent `gpu` check reports `[OK]` and discovers `GPU 0: Tesla T4`

### Checklist

### DIFF
--- a/internal/controller/datadogagent/feature/gpu/feature.go
+++ b/internal/controller/datadogagent/feature/gpu/feature.go
@@ -15,6 +15,8 @@ import (
 	"github.com/DataDog/datadog-operator/internal/controller/datadogagent/component/agent"
 	"github.com/DataDog/datadog-operator/internal/controller/datadogagent/feature"
 	"github.com/DataDog/datadog-operator/internal/controller/datadogagent/object/volume"
+	"github.com/DataDog/datadog-operator/internal/controller/datadogagent/providercaps"
+	"github.com/DataDog/datadog-operator/pkg/kubernetes"
 )
 
 func init() {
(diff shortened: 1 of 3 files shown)


internal/controller/datadogagent/feature/gpu/feature.go
```diff
@@ -42,6 +44,33 @@ func (f *gpuFeature) ID() feature.IDType {
...
+// NodeAgentProviderCapabilities returns provider-conditional pod-template
+// mutations for the node agent. On GKE COS, the NVIDIA driver libraries are not
+// available at the standard path used by the nvidia-container-runtime; mount
+// them from the host location (/home/kubernetes/bin/nvidia/lib64) so the agent
+// (and system-probe, when privileged) can load them.
+func (f *gpuFeature) NodeAgentProviderCapabilities() providercaps.ProviderCapabilityMap {
+	containers := []apicommon.AgentContainerName{apicommon.CoreAgentContainerName}
+	if f.isPrivilegedModeEnabled {
+		containers = append(containers, apicommon.SystemProbeContainerName)
+	}
```
mbertrone, line 56: nit (non-blocking): this container-list build duplicates the identical block in `Configure` (`feature.go:83-86`), where `f.isPrivilegedModeEnabled` gates appending `SystemProbe` the same way. No issue today, but the two can silently drift if the privileged-mode container set ever changes. A small `func (f *gpuFeature) targetContainers() []apicommon.AgentContainerName` shared by both would keep them in lockstep; the current form reads fine if you'd rather leave it.

mbertrone (approved):
LGTM! One non-blocking nit inline.

tbavelier (approved):

PR #3220, opened 2026-07-03
Enable hostPID for the GPU feature

gjulianm:
### What does this PR do?

The Node Agent pod now runs in the host PID namespace when the GPU feature is enabled.

### Motivation

GPU monitoring watches host processes, which requires the host PID namespace. It is also a prerequisite for `patchCgroupPermissions`: the cgroup permissions patch resolves the host pid 1 cgroup and is a no-op without it. Unlike the Helm chart, the operator does not enable `hostPID` by default, so this was previously missing.

### Minimum Agent Versions

- Agent: N/A
- Cluster Agent: N/A

### Describe your test plan

1. Deploy DDA with gpuM enabled
2. Verify hostPID is set for Agent pod

### Checklist

- [x] All commits are signed (see: [signing commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits))

### DIFF
--- a/internal/controller/datadogagent/feature/gpu/feature.go
+++ b/internal/controller/datadogagent/feature/gpu/feature.go
@@ -200,6 +200,11 @@ func (f *gpuFeature) configureCgroupPermissions(managers feature.PodTemplateMana
 // ManageNodeAgent allows a feature to configure the Node Agent's corev1.PodTemplateSpec
 // It should do nothing if the feature doesn't need to configure it.
 func (f *gpuFeature) ManageNodeAgent(managers feature.PodTemplateManagers) error {
+	// GPU monitoring watches host processes, so it needs the host PID namespace.
+	// It is also required when patchCgroupPermissions is enabled, as the cgroup
+	// permissions patch resolves the host pid 1 cgroup.
+	managers.PodTemplateSpec().Spec.HostPID = true
+
 	// env var to enable the GPU core check
 	enableCoreCheckEnvVar := &corev1.EnvVar{
 		Name:  DDEnableGPUMonitoringCheckEnvVar,
--- a/internal/controller/datadogagent/feature/gpu/feature_test.go
+++ b/internal/controller/datadogagent/feature/gpu/feature_test.go
@@ -230,6 +230,9 @@ func Test_GPUMonitoringFeature_Configure(t *testing.T) {
 		systemProbeEnvVars := mgr.EnvVarMgr.EnvVarsByC[apicommon.SystemProbeContainerName]
 		assert.ElementsMatch(t, systemProbeEnvVars, wantSystemProbeEnvVars)
 
+		// GPU monitoring requires the host PID namespace
+		assert.True(t, mgr.PodTemplateSpec().Spec.HostPID, "HostPID should be enabled for GPU monitoring")
+
 		// Check runtime class
 		if expectedRuntimeClass == "" {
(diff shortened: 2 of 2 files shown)


tbavelier (approved):

internal/controller/datadogagent/feature/gpu/feature.go
```diff
@@ -200,6 +200,11 @@ func (f *gpuFeature) configureCgroupPermissions(managers feature.PodTemplateMana
 // ManageNodeAgent allows a feature to configure the Node Agent's corev1.PodTemplateSpec
 // It should do nothing if the feature doesn't need to configure it.
 func (f *gpuFeature) ManageNodeAgent(managers feature.PodTemplateManagers) error {
+	// GPU monitoring watches host processes, so it needs the host PID namespace.
+	// It is also required when patchCgroupPermissions is enabled, as the cgroup
+	// permissions patch resolves the host pid 1 cgroup.
+	managers.PodTemplateSpec().Spec.HostPID = true
```
mbertrone, line 206: question: `HostPID` is set unconditionally here, so it also lands on the core-check-only path (`privilegedMode: false`), where the pod runs just the core agent and is otherwise unprivileged. `Configure` requires only `CoreAgentContainerName` on that path (`feature.go:53`), so this line is what puts an otherwise-unprivileged pod into the host PID namespace. The comment and the new `GPUCoreCheckOnlyWantFunc` assertion make clear that's intended, so this is confirm-your-intent, not a defect: does the NVML core check read host processes when it isn't privileged? If not, gating on `f.isPrivilegedModeEnabled || f.patchCgroupPermissions` would keep the non-privileged path out of the host PID namespace.

internal/controller/datadogagent/feature/gpu/feature_test.go
```diff
@@ -314,6 +317,9 @@ func Test_GPUMonitoringFeature_Configure(t *testing.T) {
 		systemProbeEnvVars := mgr.EnvVarMgr.EnvVarsByC[apicommon.SystemProbeContainerName]
 		assert.Empty(t, systemProbeEnvVars, "System Probe should not have env vars when privileged mode is disabled")
 
+		// GPU monitoring requires the host PID namespace, even in core-check only mode
+		assert.True(t, mgr.PodTemplateSpec().Spec.HostPID, "HostPID should be enabled for GPU monitoring")
```
mbertrone, line 321: nit (non-blocking): the motivation calls out `patchCgroupPermissions` as the main driver for `hostPID`, yet `GPUMonitoringWithCgroupPermissionsWantFunc` (L331) is the one enabled path that doesn't assert it. It's covered transitively (that case is privileged and the flag is set unconditionally), so this is purely cosmetic. Dropping the same one-line assert there would pin the behavior right next to the feature it most directly unblocks.

mbertrone (approved):
LGTM — non-blocking question + nit inline.

internal/controller/datadogagent/feature/gpu/feature.go
gjulianm: Yes, that's by design. The core agent also reads processes so we need to have hostPID there.

internal/controller/datadogagent/feature/gpu/feature_test.go
gjulianm: Right now I don't think it's necessary, as the tests cover the feature well without conditions.

PR #3240, opened 2026-07-09
Clean registry migration and force GCR on GKE Autopilot if unset

tbavelier:
### What does this PR do?

* c50d2ec3c Defaults Agent images to `registry.datadoghq.com` and removes the `DD_REGISTRY_OVERRIDE_*` migration path.
* 4485a2edb Forces GCR for GKE Autopilot unless an explicit GCR, EU GCR, or Asia GCR registry is configured.
* 3672d18e3 remove useless test matrix since the switch is simpler now

### Motivation

* Clean up migration env vars: registry has been live for a while now without issues
* Ensure we can deploy on Autopilot overriding the default/user mistake

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

* Verify images are pulled from `registry.datadoghq.com` by default when deploying on `kind`
* On a GKE Autopilot cluster, set no registry and ensure images are pulled from `gcr.io/datadoghq`.

### Checklist


pkg/images/images.go
```diff
@@ -60,6 +60,14 @@ const (
 	DefaultDdotCollectorImageName string = "ddot-collector"
 )
 
+// IsGCRRegistry reports whether registry points to a Datadog GCR registry,
+// including the known regional GCR registries.
+func IsGCRRegistry(registry string) bool {
```
gabedos, line 65: potential nit: https://docs.cloud.google.com/artifact-registry/docs/transition/gcr-repositories?utm_source=chatgpt.com#gcr-domain-support

`us.gcr.io` also exists and could be added here? Also, maybe we can make this more resilient by using regex if we expect the gcr.io in the name? This will make it so that no specific Operator version is pinned down to the registries that existed when the Operator was released.

tbavelier: I initially went indeed with something that only looks for `gcr.io` string inside repositories, but felt like it was cleaner like this. We (Datadog) only use gcr.io/datadoghq, eu.gcr.io/datadoghq and asia one. If we add a new GCR registry, it will need to be present in operator code anyway, so this function can then be updated imho. What do you think ?

gabedos: Yeah I think this is fine! I'm split 50/50 either way on it

gabedos (approved):
Defaulting to `registry.datadoghq.com` alongside specific gke autopilot overrides lgtm

PR #3264, opened 2026-07-16
[CONTP-1884] Migrate AppArmor annotations to appArmorProfile

tbavelier:
### What does this PR do?

Migrate deprecated container-scoped AppArmor annotations to `securityContext.appArmorProfile` on Kubernetes 1.30 and later.

The completed Pod template is finalized after features and overrides are applied, but before workload hashing and submission. On Kubernetes versions older than 1.30, the existing annotations are preserved for compatibility. The finalizer supports `unconfined`, `runtime/default`, and localhost profiles, and preserves an explicitly configured `appArmorProfile` field.

### Motivation

Kubernetes 1.30 deprecates `container.apparmor.security.beta.kubernetes.io/<container>` annotations in favor of the structured `appArmorProfile` field. The annotation path will no longer be supported starting with Kubernetes 1.37 (refer to discussion on https://github.com/kubernetes/enhancements/issues/24)

This removes the deprecation warning on supported clusters while retaining compatibility with older Kubernetes versions.

### Additional Notes

I considered directly setting `securityContext.appArmorProfile` in every feature that currently requests an AppArmor profile, instead of using the annotation as an intermediate representation and finalizing it at the workload boundary.

Pros of the direct approach:

- Each feature would express its final security-context intent locally.
- Generated templates would never contain the deprecated annotation internally.
- The relationship between a feature and its AppArmor requirements would be more explicit.

Cons of the direct approach:

- AppArmor is currently emitted by several features as well as container overrides and raw Pod annotation overrides; updating only feature code would be incomplete.
- A correct implementation would need to thread `PlatformInfo` through `PodTemplateManagers` or the security-context manager, including its fake implementations and tests.
- Duplicating Kubernetes-version branching across producers would increase the risk of omissions and future behavioral drift.
- Raw annotation overrides would still require a centralized compatibility path.

The centralized Pod-template finalizer keeps the Kubernetes-version policy in one place, covers current and future annotation producers, and gives an explicitly configured `appArmorProfile` field precedence.

### Minimum Agent Versions

No minimum Datadog Agent or Cluster Agent version is required.

* Agent: none
* Cluster Agent: none

### Describe your test plan

Automated validation:

- `go test ./internal/controller/datadogagent/common ./pkg/kubernetes`
- `go test ./internal/controller/datadogagent ./internal/controller/datadogagentinternal`
- `go test ./internal/controller/testutils/renderer -run 'TestRender_AppArmorProfileVersionGate|TestRender_MinimalDDA'`

QA instructions:

1. Deploy the Operator on a Kubernetes 1.30+ cluster.
2. Deploy a DatadogAgent with a system-probe feature enabled, for example:

   ```yaml
   spec:
     features:
       npm:
         enabled: true
   ```

3. Inspect an Agent Pod created from that DatadogAgent.
4. Verify that the `system-probe` container has `securityContext.appArmorProfile.type: Unconfined`.
5. Verify that the Pod no longer has the `container.apparmor.security.beta.kubernetes.io/system-probe` annotation (can verify the annotation is present once again if deploying previous/stable operator version)

### Checklist

### DIFF
--- a/internal/controller/datadogagent/common/apparmor.go
+++ b/internal/controller/datadogagent/common/apparmor.go
@@ -1,0 +1,75 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2026-present Datadog, Inc.
+
+package common
+
+import (
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/ptr"
+
+	"github.com/DataDog/datadog-operator/pkg/kubernetes"
+)
+
+// FinalizeAppArmorProfile applies Kubernetes-version compatibility to AppArmor
+// settings on a completed Pod template. It must run after features and overrides
+// have finished mutating the template, but before the workload is hashed and sent
+// to the API server.
+//
+// On Kubernetes 1.30 and later, it moves deprecated container-scoped AppArmor
+// annotations to securityContext.appArmorProfile. Older Kubernetes versions keep
+// the annotations. An explicitly configured appArmorProfile field takes precedence
+// over an annotation.
+func FinalizeAppArmorProfile(podTemplate *corev1.PodTemplateSpec, platformInfo kubernetes.PlatformInfo) {
+	if podTemplate == nil || !platformInfo.SupportsAppArmorProfile() {
+		return
+	}
+
+	for annotation, value := range podTemplate.Annotations {
+		containerName, found := strings.CutPrefix(annotation, AppArmorAnnotationKey+"/")
+		if !found {
+			continue
+		}
+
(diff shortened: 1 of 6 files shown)


internal/controller/datadogagent/common/apparmor.go
```diff
@@ -0,0 +1,76 @@
...
+		if !found {
+			continue
+		}
+
+		setAppArmorProfile(podTemplate.Spec.Containers, containerName, value)
+		setAppArmorProfile(podTemplate.Spec.InitContainers, containerName, value)
+		delete(podTemplate.Annotations, annotation)
+	}
+
+	if len(podTemplate.Annotations) == 0 {
```
levan-m: I think this should be dropped so this function doesn't set it to nil  for downstream code.

tbavelier: Good call, fixed with ebc531694029ef756ebff5528faa36c9ef55a3b1

levan-m (approved):

PR #3268, opened 2026-07-16
[TON-742] Support managed DatadogAgent creation and removal

fanny-jiang:
### What does this PR do?

Adds opt-in support for managed DatadogAgent creation and removal. The Operator reads lifecycle requests from a ConfigMap, creates or removes the DatadogAgent and Windows AgentProfile resources, and reports progress and completion through Remote Config.

The new flow coordinates with existing Fleet configuration updates so the two operations do not overwrite each other's state.

### Motivation

After the EKS instrumenter installs the Operator EKS add-on, it needs a reliable way to create the DatadogAgent and confirm that removal has completed before removing the add-on.

### Additional Notes

The feature is disabled by default. EKS is the first supported provider, while the managed-DatadogAgent flow remains provider-neutral and extensible. Existing DatadogAgent and DatadogAgentInternal reconciliation behavior is unchanged.

### Minimum Agent Versions

* Agent: N/A
* Cluster Agent: N/A

### Describe your test plan

- `CGO_ENABLED=0 go test ./cmd ./pkg/config ./pkg/fleet ./pkg/remoteconfig`
- `CGO_ENABLED=0 make lint`

### Checklist


fanny-jiang:
@codex review

cmd/main.go
```diff
@@ -424,22 +428,59 @@ func run(opts *options) error {
...
+			case <-mgr.Elected():
+			}
 			if rcErr := rcUpdater.Setup(creds); rcErr != nil {
 				setupErrorf(setupLog, rcErr, "Unable to set up Remote Config service")
 				return
 			}
-
 			if opts.remoteUpdatesEnabled {
-				if rcErr := setupFleetDaemon(setupLog, mgr, rcUpdater.Client(), opts.createControllerRevisions && opts.datadogAgentEnabled); rcErr != nil {
+				if rcErr := setupFleetDaemon(setupLog, mgr, rcUpdater.Client(), opts.createControllerRevisions && opts.datadogAgentEnabled, managedAgentInstallationIdentity, managedAgentInstallationEnabled); rcErr != nil {
```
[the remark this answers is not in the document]
fanny-jiang: Addressed in [c9712cd](https://github.com/DataDog/datadog-operator/pull/3268/commits/c9712cddd38014e92cd304234a4ee75a6deeb463)

fanny-jiang:
@codex review

pkg/fleet/managed_agent_installation.go
```diff
@@ -0,0 +1,856 @@
...
+		return nil
+	}
+	var global map[string]json.RawMessage
+	if err := json.Unmarshal(rawGlobal, &global); err != nil {
+		return fmt.Errorf("invalid config.spec.global: %w", err)
+	}
+	if _, ok := global["credentials"]; ok {
+		return fmt.Errorf("config must not contain spec.global.credentials")
+	}
+	for _, field := range []string{"clusterAgentToken", "clusterAgentTokenSecret"} {
```
gpalmz: `encoding/json` matches struct fields case-insensitively, so a config with capitalized keys (e.g. `{"Spec":{"Global":{"ClusterAgentToken":"..."}}}`) slips past this case-sensitive map scan and still decodes into the typed struct. `buildFleetDatadogAgentSpec` overwrites `Credentials` but not `ClusterAgentToken`/`ClusterAgentTokenSecret`, so those would survive into the created DDA. Since it's already decoded into a typed struct, could we reject on the decoded fields instead (non-nil `Global.Credentials`/`ClusterAgentToken`/`ClusterAgentTokenSecret`)? That closes the casing gap and drops the duplicate parse.

pkg/fleet/daemon_worker.go
```diff
@@ -183,6 +193,12 @@ func evaluatePendingTask(snapshot ddaStatusSnapshot, task pendingOperation) (boo
 // between, the annotations may be cleaned up later, but the task result is not
 // lost.
 func (d *Daemon) finishPendingOperation(ctx context.Context, task pendingOperation, resultErr error) {
+	if resultErr == nil && task.intent == pendingIntentPromote {
+		if err := d.persistManagedAgentInstallationStableConfig(ctx, task.nsn, task.experimentID, task.resultVersion); err != nil {
+			resultErr = fmt.Errorf("persist promoted managed Agent installation config: %w", err)
```
gpalmz, line 198: If the reconciler has already promoted the experiment but `persistManagedAgentInstallationStableConfig` fails hard (not a conflict), this flips `resultErr` and reports the promote as failed to RC, even though the cluster is in the promoted phase - so Fleet may re-drive a successful operation. Is the re-drive idempotent here, or would it be safer to report the promote DONE and retry the config-persist separately (best-effort/async) rather than gating the task result on it? Same coupling on the `daemon_operations.go` fast path.

pkg/fleet/managed_agent_installation.go
```diff
@@ -0,0 +1,856 @@
...
+	}
+	if !apiequality.Semantic.DeepDerivative(*desired, dda.Spec) {
+		return &stateDoesntMatchError{msg: fmt.Sprintf("DatadogAgent %s/%s spec does not match Fleet config %q", dda.Namespace, dda.Name, configID)}
+	}
+	liveHash, err := fleetDatadogAgentSpecHash(&dda.Spec)
+	if err != nil {
+		return fmt.Errorf("hash live DatadogAgent spec: %w", err)
+	}
+	if dda.Annotations[fleetConfigHashAnnotation] != liveHash {
+		return &stateDoesntMatchError{msg: fmt.Sprintf("DatadogAgent %s/%s spec changed after Fleet config %q was accepted", dda.Namespace, dda.Name, configID)}
```
gpalmz: Small durability edge: if the process dies after `Create` but before `recordFleetDatadogAgentSpecHash` commits, the DDA is left `partial` with `fleetConfigHashAnnotation = hash(preSpec)` while the live spec is the post-admission spec. On retry, this `annotation != liveHash` check fails terminally ("spec changed after Fleet config was accepted") and the install never self-heals. This only bites if a DDA `.spec` is mutated on admission/defaulting - is that possible here? If so, could the replay path re-record the accepted hash (like the create path does) when the spec still passes `DeepDerivative`, rather than hard-failing?

pkg/fleet/daemon.go
```diff
@@ -288,45 +354,195 @@ func (d *Daemon) setTaskState(pkgName, taskID string, taskState pbgo.TaskState,
...
+				fleetOwned.Name,
+				nsn.Namespace,
+				nsn.Name,
+			)
+		}
+		fleetOwned = &nsn
+	}
+	if fleetOwned != nil && unmanaged != nil {
+		return &stateDoesntMatchError{msg: fmt.Sprintf(
+			"Fleet-managed DatadogAgent %s/%s coexists with unmanaged DatadogAgent %s/%s",
```
gpalmz: `ddas.Items` is pre-filtered to the canonical target above, so the multiple-owned / active-experiment / owned+unmanaged-coexistence branches here can't fire (`activeExperiment` is also unused). Not a deletion risk, but it means the coexistence guard doesn't actually guard. Classify before the filter, or drop the dead branches?

pkg/fleet/managed_agent_installation_intent.go
```diff
@@ -0,0 +1,514 @@
...
+
+func (d *Daemon) requestManagedAgentInstallationRetryAfter() {
+	d.requestManagedAgentInstallationRetryAfterDelay(managedAgentInstallationRetryInterval)
+}
+
+func (d *Daemon) requestManagedAgentInstallationRetryAfterDelay(delay time.Duration) {
+	if d.managedAgentInstallationUpdates == nil {
+		return
+	}
+	time.AfterFunc(delay, d.requestManagedAgentInstallationRetry)
```
gpalmz, line 213: `time.AfterFunc` here isn't tracked or stopped on `ctx.Done()`, so after shutdown the timer still fires into `requestManagedAgentInstallationRetry` (whose consumer has exited). Harmless per-timer, but they can accumulate across leader churn - worth gating the callback on a context check or stopping the timers on `Stop`.

gpalmz (approved):
Really solid, thoroughly tested PR. Approving. Left a few comments - a couple are MED questions worth a look (the credential-guard casing bypass, the promote-reported-as-failed coupling, and the partial-stuck edge), the rest are minor.

pkg/fleet/managed_agent_installation_intent.go
fanny-jiang: The callback only performs a non-blocking send, and all delays are at most 8 seconds; manager shutdown terminates the process. I'd prefer to keep this scoped and defer timer tracking.

pkg/fleet/managed_agent_installation_intent.go
```diff
@@ -0,0 +1,514 @@
...
+	Bootstrap      managedAgentInstallationBootstrap    `json:"bootstrap"`
+}
+
+type managedAgentInstallationIntentSnapshot struct {
+	raw             []byte
+	resourceVersion string
+}
+
+func (d *Daemon) installManagedAgentInstallationIntentForwarder(ctx context.Context) error {
+	informer, err := d.cache.GetInformer(ctx, &corev1.ConfigMap{})
```
zhuminyi, line 83: Managed install can be enabled while the cache watches a different namespace. The feature hard-codes managed resources to datadog-agent in [pkg/fleet/managed_agent_installation.go:52-56](https://github.com/datadog/datadog-operator/blob/30d3e87cfa6e840d843365382792aba7902e6b5d/pkg/fleet/managed_agent_installation.go#L52-L56), but cache namespaces still come from DD_AGENT_WATCH_NAMESPACE/WATCH_NAMESPACE in [pkg/config/config.go:245-249](https://github.com/datadog/datadog-operator/blob/30d3e87cfa6e840d843365382792aba7902e6b5d/pkg/config/config.go#L245-L249). The intent/credential informers in [pkg/fleet/managed_agent_installation_intent.go:87-119](https://github.com/datadog/datadog-operator/blob/30d3e87cfa6e840d843365382792aba7902e6b5d/pkg/fleet/managed_agent_installation_intent.go#L87-L119) will miss updates if that namespace is not watched

fanny-jiang: Good catch — the managed installation namespace is now included in the relevant cache watches, including custom watch configurations.

pkg/fleet/daemon_worker.go
fanny-jiang: Yes — retrying is safe here. If promotion already completed, we retry saving the stable config and then mark the task done.

pkg/fleet/daemon.go
fanny-jiang: Addressed — I removed the dead coexistence branches and simplified this to classify the canonical target directly.

pkg/fleet/managed_agent_installation.go
fanny-jiang: Addressed — interrupted creates now recover by recording the accepted live spec hash before continuing.

pkg/fleet/managed_agent_installation.go
fanny-jiang: Addressed — credential validation now checks the decoded typed fields, so casing variants are rejected too.

pkg/fleet/managed_agent_installation_windows_profile.go
```diff
@@ -0,0 +1,122 @@
...
+				fleetTargetIDLabel:                         d.managedAgentInstallationIdentity.TargetID(),
+			},
+			Annotations: map[string]string{
+				kubernetes.ProviderAnnotationKey: kubernetes.WindowsProvider,
+			},
+			OwnerReferences: []metav1.OwnerReference{controllerOwnerReference(
+				v2alpha1.GroupVersion.String(), "DatadogAgent", dda.Name, dda.UID,
+			)},
+		},
+		Spec: v1alpha1.DatadogAgentProfileSpec{
```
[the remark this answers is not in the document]
fanny-jiang: Addressed — the generated Windows profile now includes an empty config and passes profile validation.

pkg/fleet/managed_agent_installation_command.go
```diff
@@ -0,0 +1,226 @@
...
+	d.managedAgentInstallationOperationID = command.Intent.OperationID
+	d.managedAgentInstallationCancel = cancel
+	d.managedAgentInstallationDone = make(chan struct{})
+	d.setTaskState(packageDatadogOperator, command.Intent.OperationID, pbgo.TaskState_RUNNING, nil)
+	d.taskMu.Unlock()
+	d.transitionMu.Unlock()
+
+	if err := d.writeManagedAgentInstallationState(taskCtx, managedAgentInstallationStateFromCommand(command, pbgo.TaskState_RUNNING, nil)); err != nil {
+		d.finishManagedAgentInstallationTask(command.Intent.OperationID)
+		return fmt.Errorf("persist accepted managed Agent installation intent: %w", err)
```
rahulkaukuntla, line 108: This clears `managedAgentInstallationActive` but not `managedAgentInstallationTaskReserved`, and doesn't call `requestManagedAgentInstallationRetryAfter()` like the other failure paths do. If the initial state write fails transiently, the reserved flag stays true forever and `handleTask` will reject every future Fleet task until the process restarts.

pkg/fleet/managed_agent_installation.go
```diff
@@ -0,0 +1,876 @@
...
+	if err := d.validateManagedAgentInstallationWindowsProfileExists(ctx, observed); err != nil {
+		return d.retainFleetDatadogAgentPartial(ctx, command, dda.UID, err)
+	}
+
+	ctrl.LoggerFrom(ctx).Info("Created Fleet-managed DatadogAgent", "namespace", dda.Namespace, "name", dda.Name, "config", configID)
+	return nil
+}
+
+func (d *Daemon) retainFleetDatadogAgentPartial(ctx context.Context, command managedAgentInstallationCommand, uid types.UID, cause error) error {
+	d.setPackageConfigVersions(packageDatadogOperator, fleetPartialConfigVersionPrefix+command.Intent.OperationID, "")
```
rahulkaukuntla: This sets the config version to `command.Intent.OperationID` before we've confirmed that's the config actually owning the live resource — if `markFleetDatadogAgentPartial` below fails, we return with this possibly-wrong value still recorded instead of the real `configID`. Consider only setting it once, after `configID` is resolved.

rahulkaukuntla (changes requested):

pkg/fleet/managed_agent_installation_command.go
fanny-jiang: Good catch — fixed. We now release the reservation and retry when the state write fails.

pkg/fleet/managed_agent_installation.go
fanny-jiang: Good catch — fixed. We now update the config version only after confirming the live config ID.

rahulkaukuntla (approved):

zhuminyi (approved):

PR #3302, opened 2026-07-23
fix(gpu): force disable eBPF probes

betterengineering:
### What does this PR do?

In privileged mode the GPU eBPF probes are now force-disabled (DD_GPU_MONITORING_ENABLE_EBPF_PROBES=false) on both the core agent and system-probe; privileged mode is retained for cgroup permission patching. Clusters that still need the probes can re-enable them by setting DD_GPU_MONITORING_ENABLE_EBPF_PROBES=true on those containers under spec.override.nodeAgent.containers — overrides are applied after feature config and env vars merge last-writer-wins by name, so the override wins with no duplicate env var.

### Motivation

The eBPF probes for GPU Monitoring are deprecated. Some customers still have the previously recommended setting gpu.privilegedMode enabled. These probes have caused crashes on NVIDIA GB300 Grace Blackwell Ultra machines. We want to disable the probes by default moving forward so they no longer cause problems. We still need to support `privilegedMode` in some environments for cgroup permission patching.

### Additional Notes

We've had two customer cases of this and we want to prevent further cases.

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

The additional unit tests.

### Checklist

### DIFF
--- a/internal/controller/datadogagent/feature/gpu/feature.go
+++ b/internal/controller/datadogagent/feature/gpu/feature.go
@@ -127,6 +127,13 @@ func configureSystemProbe(managers feature.PodTemplateManagers) {
 	// add the env var to the core agent as well, to prevent config mismatches in runtime
 	managers.EnvVar().AddEnvVarToContainer(apicommon.CoreAgentContainerName, enableSPEnvVar)
 
+	// In privileged mode the eBPF probes are disabled, as GPU monitoring relies on
+	// the privileged host access rather than the eBPF probes.
+	managers.EnvVar().AddEnvVarToContainer(apicommon.SystemProbeContainerName, &corev1.EnvVar{
+		Name:  DDEnableEBPFProbesEnvVar,
+		Value: "false",
+	})
+
 	// annotations
 	managers.Annotation().AddAnnotation(common.SystemProbeAppArmorAnnotationKey, common.SystemProbeAppArmorAnnotationValue)
 
--- a/internal/controller/datadogagent/feature/gpu/feature.go
+++ b/internal/controller/datadogagent/feature/gpu/feature.go
@@ -129,10 +129,18 @@ func configureSystemProbe(managers feature.PodTemplateManagers) {
 
 	// In privileged mode the eBPF probes are disabled, as GPU monitoring relies on
 	// the privileged host access rather than the eBPF probes.
-	managers.EnvVar().AddEnvVarToContainer(apicommon.SystemProbeContainerName, &corev1.EnvVar{
+	disableEBPFProbesEnvVar := &corev1.EnvVar{
 		Name:  DDEnableEBPFProbesEnvVar,
 		Value: "false",
-	})
+	}
+
+	// disable the eBPF probes in system-probe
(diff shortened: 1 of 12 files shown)


internal/controller/datadogagent/feature/gpu/feature.go
```diff
@@ -127,6 +127,21 @@ func configureSystemProbe(managers feature.PodTemplateManagers) {
 	// add the env var to the core agent as well, to prevent config mismatches in runtime
 	managers.EnvVar().AddEnvVarToContainer(apicommon.CoreAgentContainerName, enableSPEnvVar)
 
+	// In privileged mode the eBPF probes are disabled, as GPU monitoring relies on
+	// the privileged host access rather than the eBPF probes.
+	disableEBPFProbesEnvVar := &corev1.EnvVar{
+		Name:  DDEnableEBPFProbesEnvVar,
+		Value: "false",
```
mbertrone, line 137: suggestion (non-blocking): this force-disables the probes, which makes the `privilegedMode` field doc stale. The marker at `api/datadoghq/v2alpha1/datadogagent_types.go:721` still reads "PrivilegedMode enables GPU Probe module in System Probe" — accurate for the module, but the probes it implies are now off by default. That comment is the source for both generated docs (`docs/configuration_public.md:235`, `docs/configuration.v2alpha1.md:136`), so they drift with it. Update the marker to note the eBPF probes are disabled by default and privileged mode is retained for cgroup permission patching, then run `make generate && make manifests` so both generated docs move in lockstep.

internal/controller/datadogagent/feature/gpu/feature.go
```diff
@@ -127,6 +127,21 @@ func configureSystemProbe(managers feature.PodTemplateManagers) {
...
+	}
+
+	// disable the eBPF probes in system-probe
+	managers.EnvVar().AddEnvVarToContainer(apicommon.SystemProbeContainerName, disableEBPFProbesEnvVar)
+
+	// add the env var to the core agent as well, so both containers agree that the
+	// eBPF probes are disabled and the core GPU check does not poll a system-probe
+	// module that has the probes turned off
+	managers.EnvVar().AddEnvVarToContainer(apicommon.CoreAgentContainerName, disableEBPFProbesEnvVar)
```
mbertrone, line 146: suggestion (non-blocking): this hard-codes the probes off with no first-class opt-out, which is the right default given the GB300 crashes. Worth documenting the escape hatch that already exists: feature config runs before `spec.override`, and env merge is last-writer-wins by name (`OverrideCurrentEnvVarMergeFunction` in `merger/envvars.go`), so a user who still needs the probes can set `DD_GPU_MONITORING_ENABLE_EBPF_PROBES=true` via `spec.override.nodeAgent.containers` and it wins cleanly, with no duplicate env var. A line in the release note / docs saying probes are now disabled by default in privileged mode and can be re-enabled that way would save a support round-trip.

mbertrone (approved):

OliviaShoup (approved):
thanks for the PR!

levan-m (approved):

PR #3310, opened 2026-07-29
DatadogGenericResource: Don't advance LastForceSyncTime on failed force-sync attempts

glasser:
### What does this PR do?

status.LastForceSyncTime was stamped unconditionally as soon as a periodic force-sync was attempted, before the resulting create/update was known to succeed. On a failed sync (e.g. a transient 504 from the Datadog API), this made the controller wait out a full forceSyncPeriod (60m by default) before retrying, instead of retrying on the fast error-requeue cadence used everywhere else in this reconciler. create()/update() already stamp LastForceSyncTime themselves on success, so the redundant early stamp is simply removed.

### Motivation

We observed this in practice on DatadogGenericResource monitors that received 504s from Datadog; waiting an hour (or doing a manual edit to the jsonSpec) was our only workaround. We would prefer a retry like with other errors.

### Additional Notes

The research and authorship of the PR was done via Claude, but to the degree that I was able to familiarize myself with the codebase, I agree with its analysis.

This may affect resources other than DatadogGenericResource but we have fully migrated to DGR because the others are deprecated.

### Minimum Agent Versions

No, this code structure has been there for a while.

### Describe your test plan

Automated test added.

### Checklist

### DIFF
--- a/internal/controller/datadoggenericresource/controller.go
+++ b/internal/controller/datadoggenericresource/controller.go
@@ -159,10 +159,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, instance *v1alpha1.DatadogGe
 				if strings.Contains(err.Error(), ctrutils.NotFoundString) {
 					shouldCreate = true
 				}
+				// Do not advance LastForceSyncTime here: leaving it stale lets the next
+				// reconcile retry the force sync instead of waiting out a full forceSyncPeriod.
+				// create()/update() advance it themselves once the sync actually succeeds.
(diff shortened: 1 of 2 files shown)


tbavelier:
Thank you for the contribution @glasser ! As our CI doesn't run from forks, I created https://github.com/DataDog/datadog-operator/pull/3312 from your commit to preserve authorship. This will ship with version 1.30 in a few weeks

tbavelier (approved):

PR #3335, opened 2026-08-06
Gate remote agent registry on vsock by resolved agent version

rahulkaukuntla:
### What does this PR do?

Instead of unconditionally dropping the `DD_REMOTE_AGENT_REGISTRY_ENABLED=false` override on vsock clusters, only drop it when the node agent's resolved image tag is `7.83+`.

### Motivation

RAR's AF_VSOCK support (DataDog/datadog-agent#54474) only ships from agent 7.83. Since customers can upgrade the operator and agent versions independently, there's no point at which the operator can assume the agent is 7.83+ — so the override must stay version-gated rather than removed outright, or older agents on vsock clusters would crash-loop.

### Describe how you validated your changes

Added test cases covering both below- and at-minimum-version scenarios; `go test ./internal/controller/datadogagent/...` passes.

### DIFF
--- a/internal/controller/datadogagent/global/agent.go
+++ b/internal/controller/datadogagent/global/agent.go
@@ -38,8 +38,7 @@ func applyNodeAgentResources(manager feature.PodTemplateManagers, ddaSpec *v2alp
 		if override, ok := ddaSpec.Override[v2alpha1.NodeAgentComponentName]; ok && override.Image != nil {
 			agentVersion = common.GetAgentVersionFromImage(*override.Image)
 		}
-		defaultIfVersionUnknown := false
-		if !utils.IsAboveMinVersion(agentVersion, remoteAgentRegistryVSockMinVersion, &defaultIfVersionUnknown) {
+		if !utils.IsAboveMinVersion(agentVersion, remoteAgentRegistryVSockMinVersion, new(false)) {
 			manager.EnvVar().AddEnvVar(&corev1.EnvVar{
 				Name:  DDRemoteAgentRegistryEnabled,
 				Value: "false",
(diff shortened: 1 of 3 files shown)


internal/controller/datadogagent/global/agent.go
```diff
@@ -27,12 +27,6 @@ func applyNodeAgentResources(manager feature.PodTemplateManagers, ddaSpec *v2alp
 			Value: "host",
 		})
```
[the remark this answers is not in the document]
rahulkaukuntla: Fixed in 23d966d7 — the override is now version-gated on the resolved node agent image tag (kept when <7.83, dropped when >=7.83) instead of removed unconditionally. Thanks for catching this.

internal/controller/datadogagent/global/agent.go
```diff
@@ -27,11 +34,17 @@ func applyNodeAgentResources(manager feature.PodTemplateManagers, ddaSpec *v2alp
...
-		manager.EnvVar().AddEnvVar(&corev1.EnvVar{
-			Name:  DDRemoteAgentRegistryEnabled,
-			Value: "false",
-		})
+		agentVersion := images.AgentLatestVersion
+		if override, ok := ddaSpec.Override[v2alpha1.NodeAgentComponentName]; ok && override.Image != nil {
+			agentVersion = common.GetAgentVersionFromImage(*override.Image)
+		}
+		defaultIfVersionUnknown := false
+		if !utils.IsAboveMinVersion(agentVersion, remoteAgentRegistryVSockMinVersion, &defaultIfVersionUnknown) {
```
Mathew-Estafanous: nit: you can use `new(false)` to get a pointer

Mathew-Estafanous (approved):
one small nit, but everything else looks good

internal/controller/datadogagent/global/agent.go
rahulkaukuntla: sounds good

PR #3337, opened 2026-08-10
feat(appsec): migrate injector config to spec.features.appsec.injector and add GKE Gateway support

eliottness:
### What does this PR do?

**We are about to mark appsec injector GA**

Migrates AppSec proxy injector configuration from `agent.datadoghq.com/appsec.*` annotations to a first-class CRD surface at `spec.features.appsec.injector.*`, and adds GKE Gateway support.

- **New CRD surface** mirroring the Helm `datadog.appsec.injector.*` layout 1:1, plus a net-new `injector.gke.gatewayClasses`.
- **Per-field precedence** — annotations are parsed first, then the CRD overlays field by field: the CRD wins where set, annotations fill what is unset.
- **GKE Gateway handover** — `gke-gateway` proxy value, `gatewayClasses` → `DD_APPSEC_PROXY_GKE_GATEWAY_CLASSES`, `mode: external` required when `gke-gateway` is explicitly listed in `proxies`, and a cluster-agent >= 7.82.0 gate.
- **RBAC** — `networking.gke.io/gcptrafficextensions` (get/create/delete) and `events` gains `patch`, granted to both the operator's own ClusterRole and the cluster-agent's.
- **Deprecation** — a `V(0)` log on every reconcile when any `appsec.*` annotation is present, plus a `docs/deprecated_configs.md` entry mapping all 18 annotations to their CRD paths.

### Motivation

The `appsec.*` annotations were a stop-gap. Moving to the CRD gives users schema validation, discoverability in the generated docs, and parity with how every other feature is configured. The GKE Gateway work also needed a configuration surface that annotations could not express cleanly (`gatewayClasses` is a list and is CRD-only by design).

### Additional Notes

- **Fully backward compatible.** With no `spec.features.appsec` block present, behavior is unchanged. `TestFromAnnotations` still passes with its original cases and expectations untouched — only its call site moved from the removed `FromAnnotations` shim to `parseAnnotations(x, nil)` + `Validate()`.
- **"Set" semantics**: non-nil for pointers, `len > 0` for slices. An empty CRD list therefore does *not* clear an annotation-set value. This asymmetry is intentional and documented on the `proxies` field.
- **Malformed annotations on CRD-set fields are skipped rather than errored**, so the CRD wins even over an unparseable annotation. A malformed annotation on a field the CRD does *not* set still errors, preserving annotation-only strictness.
- `zz_generated.deepcopy.go` shows ~44 deletions that are purely an **import-alias reassignment** (core/v1 `corev1` → `v1`, meta/v1 `v1` → `metav1`), triggered by the new `corev1.ResourceRequirements` field landing earlier in declaration order. Verified equivalent by an alias-normalized diff: exactly 2 lines removed, both old import statements, zero semantic code lost.
- `appsec:` appears in the `datadogagentprofiles` CRD schema because `DatadogFeatures` is embedded by three CRDs. AppSec is **not** profile-overridable — that guardrail is the runtime `datadogAgentProfileFeatureAllowlist`, which is unchanged and still rejects it.
- GKE behavioral caveats (create-only with no drift reconciliation, no ownerReferences, ~5-7 min teardown lag, the `managed-by` ownership guard, and the `appsec.datadoghq.com/enabled=false` opt-out) are authored as leaf Go doc-comments and flow into `docs/configuration.v2alpha1.md`.
- Follow-up (pre-existing, not introduced here): a disabled or version-gated feature is dropped before `ManageDependencies`, so stale-dependency cleanup can remove the AppSec ClusterRole before the cluster-agent finishes its GCPTrafficExtension teardown. This affects all 7 features that use the same disabled-path convention and wants an operator-wide draining state.

### Minimum Agent Versions

* Agent: n/a — no node-agent changes
* Cluster Agent: **v7.82.0** for GKE Gateway injection (new). Existing gates unchanged: v7.76.0 for AppSec injection, v7.79.0 for ingress-nginx.

Note: the default cluster-agent image is currently 7.81.1, so GKE Gateway injection is gated off by default until the image is pinned to >= 7.82.0.

### Describe your test plan

`go test ./...` → **106 ok / 0 FAIL**. `make lint` → **0 issues**. `make generate manifests` is idempotent (empty `git status --porcelain` afterwards).

New coverage in `internal/controller/datadogagent/feature/appsec/`:

- **`config_test.go`** — 19 per-field CRD-wins rows (one per destination `Config` field); annotation fallback; nil injector and every nested-nil combination; empty-slice-does-not-clear; partial `resources`; `claims` dropped; deterministic first-error ordering; and each of the 4 parse-skip predicates paired with a skip-mirror case proving the skip is not over-broad.
- **`feature_test.go`** — CRD-only enable; CRD-over-annotation precedence asserted on the *emitted env var* rather than just internal state; annotation-only regression; deprecation-log firing (including when disabled); and a GKE version-gate matched pair (7.81.0 → not configured, 7.82.0 → configured) built from one shared fixture so the legs differ only in the image tag. The negative leg additionally asserts `isEnabled()`, `Validate() == nil`, and `requiresGKESupport()` so it cannot pass for the wrong reason.
- **`rbac_test.go`** — events verbs are exactly `create` + `patch`; the GKE rule is exactly `get` / `create` / `delete`.

Manual QA: there is no cluster or HTTP surface at this layer — the real user surface is *a DatadogAgent goes in, and cluster-agent env vars + `RequiredComponents` + a ClusterRole rule set come out*. Real DatadogAgents were driven through `feature.BuildFeatures` (the production entry point) for CRD-only, precedence, both malformed-annotation classes, the version-gate pair, and `getRBACPolicyRules()`, capturing observed values rather than pass/fail.

### Checklist

- [ ] All commits are signed (see: [signing commits][1]) — **not yet.** `commit.gpgsign=true` is set locally but `user.signingkey` is unset with `gpg.format=ssh`, so signing silently no-op'd across all 15 commits. To be re-signed before this leaves draft.


eliottness:
@codex review

api/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -174,6 +180,137 @@ type APMFeatureConfig struct {
...
+
+	// Controls automatic proxy detection.
+	// Default: true
+	// +optional
+	AutoDetect *bool `json:"autoDetect,omitempty"`
+
+	// Lists proxies for AppSec injection.
+	// Default: []
+	// +optional
+	Proxies []string `json:"proxies,omitempty"`
```
RomainMuller, line 211: `proxies` has a closed value set, but this field accepts arbitrary strings at admission. A typo therefore passes `kubectl apply`, then fails `Config.Validate()` and makes `Configure()` drop the entire AppSec feature with only an Operator log message. `mode` immediately below already validates the same class of constraint at admission. Could we add item-level enum validation (`+kubebuilder:validation:items:Enum=envoy-gateway;istio;istio-gateway;ingress-nginx;gke-gateway`) or use a named enum type? Runtime validation can remain as defense in depth.

api/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -174,6 +180,137 @@ type APMFeatureConfig struct {
...
+
+	// Sets the sidecar image tag.
+	// When unset, the tag defaults to a value determined by the cluster-agent image in use.
+	// +optional
+	ImageTag *string `json:"imageTag,omitempty"`
+
+	// Sets the sidecar port.
+	// Default: 8080
+	// +optional
+	Port *int32 `json:"port,omitempty"`
```
RomainMuller, line 283: Could we add admission-time bounds to the new numeric fields? As written, values such as `sidecar.port: 0` pass CRD validation but later cause `Configure()` to drop the whole AppSec feature, while a negative `bodyParsingSizeLimit` is forwarded to the Cluster Agent. Suggested constraints are `1..65535` for the sidecar and health ports, `0..65535` for the processor port if explicit zero is intentionally treated as unset, and `>=0` for the body parsing limit. This would make invalid security configuration fail visibly at apply time instead of disabling the feature during reconciliation.

docs/deprecated_configs.md
```diff
@@ -30,3 +31,32 @@ The `networkStats` field in the ServiceDiscovery feature configuration was remov
...
 
 This field can be removed from your `DatadogAgent` spec with no behavior change.
+
+### appsec.* annotations
+
+The `agent.datadoghq.com/appsec.*` annotations are deprecated in `v1.30` and will be removed in `v1.32`. They have been migrated to the `DatadogAgent` CRD spec under `spec.features.appsec.injector`.
+
+#### Migration Path
+
+Migrate your Kubernetes annotations to the `spec.features.appsec.injector` configuration in your `DatadogAgent` spec:
```
RomainMuller, line 62: Could we document the mixed annotation/CRD precedence here? Both sources are merged per field; the CRD wins only where considered set, and an empty `proxies: []` does **not** clear an annotation-provided proxy list. During migration, a user can therefore believe explicit injection was disabled while the annotation keeps it active. A malformed annotation also still rejects the feature when the corresponding CRD field is unset. Please add a short precedence subsection, and repeat the empty-list caveat on the `proxies` API field itself since `kubectl explain ...injector.proxies` currently only shows `Default: []`.

internal/controller/datadogagent/feature/appsec/rbac.go
```diff
@@ -85,6 +86,11 @@ func getRBACPolicyRules() []rbacv1.PolicyRule {
 				rbac.CreateVerb,
 			},
 		},
+		{
+			APIGroups: []string{rbac.NetworkingGKEAPIGroup},
+			Resources: []string{rbac.GCPTrafficExtensionsResource},
+			Verbs:     []string{rbac.GetVerb, rbac.CreateVerb, rbac.DeleteVerb},
```
RomainMuller, line 92: This grants every AppSec-enabled Cluster Agent cluster-wide `create` and `delete` access to a traffic-interception resource, including configurations that do not explicitly select GKE. I understand the rule may need to remain broad while `autoDetect` is enabled by default, and the Operator needs equivalent authority to grant it, but the documented ownership protection is enforced by Agent code rather than RBAC. Has this unconditional privilege expansion received explicit security/product approval? If it is intentional, could we record the rationale and blast radius in the feature documentation? It may also be possible to omit the rule only in the unambiguous case where auto-detection is explicitly disabled and the configured proxies exclude GKE.

eliottness: The idea of having the operator do the auto-detection was considered in the original design a year ago with container-platform but was abandoned due to its complexity and the fact that the current setup is not incompatible with moving the detection away from the cluster-agent. It was just easier to implement because the operator would only need to mirror the work done in the helm-chart where auto-detection is forced to happen in the cluster-agent anyway

jeff-morgan-dd (approved):

api/datadoghq/v2alpha1/datadogagent_types.go
eliottness: Done

docs/deprecated_configs.md
eliottness: Done

api/datadoghq/v2alpha1/datadogagent_types.go
eliottness: Done

RomainMuller (approved):

api/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -70,6 +70,12 @@ type DatadogFeatures struct {
 	APM *APMFeatureConfig `json:"apm,omitempty"`
 	// ASM (Application Security Management) configuration.
 	ASM *ASMFeatureConfig `json:"asm,omitempty"`
+	// Appsec configuration.
+	// This block supersedes the deprecated `agent.datadoghq.com/appsec.*` annotations. Resolution is
+	// per field: a field set here wins, and a field left unset falls back to its annotation value if
+	// one is present. An empty list here does not clear an annotation-set value.
```
Mathew-Estafanous: nit: no need for this comment. other configs have migrated from annotation to CRD without this comment

internal/controller/datadogagent/feature/appsec/feature.go
```diff
@@ -73,15 +74,59 @@ func isAboveMinVersion(ddaSpec *v2alpha1.DatadogAgentSpec) bool {
...
 		return feature.RequiredComponents{}
 	}
 
+	// Merge before validating: a CRD field can rescue configuration that the annotations
+	// alone would not satisfy. mergeInjectorConfig is a no-op when inj is nil.
+	cfg = mergeInjectorConfig(cfg, inj)
+	if validateErr := cfg.Validate(); validateErr != nil {
+		f.logger.Error(validateErr, "invalid AppSec configuration")
+		return feature.RequiredComponents{}
+	}
```
Mathew-Estafanous: All this logic is overly complicated. If the purpose is to use the CRD spec as the primary configuration of appsec then have the CRD config simply override the annotation. This complicated annotation x CRD merging logic doesn't add much value but make it more confusing to end users.

internal/controller/datadogagent/feature/appsec/feature_test.go
```diff
@@ -397,6 +498,148 @@ func TestAppsecFeature(t *testing.T) {
...
+			DDA:           gkeGatewayClassesDDA(nil),
+			WantConfigure: true,
+			ClusterAgent: assertEnv(
+				envVar{name: DDAppsecProxyEnabled, value: "true", present: true},
+				envVar{name: DDClusterAgentAppsecInjectorEnabled, value: "true", present: true},
+				envVar{name: DDAppsecProxyGKEGatewayClasses, present: false},
+			),
+		},
+		{
+			// (a) CRD-only minimal enable: no annotation whatsoever, a single CRD field.
```
Mathew-Estafanous: nit: this comment and the other comments added to the test table are distracting.

internal/controller/datadogagent/feature/appsec/feature_test.go
```diff
@@ -593,6 +836,822 @@ func TestAppsecFeatureManageClusterAgentEnabled(t *testing.T) {
...
+				"annotation %q is not covered by the deprecation warning", key)
+		})
+	}
+}
+
+// TestConfigureLogsAnnotationDeprecation pins that the deprecation warning is the very
+// first thing Configure does. It has to fire even when the feature is disabled, when the
+// cluster-agent is too old, and when the spec is nil, because those users are exactly
+// the ones who still need to hear about the migration. Every early return sits after it.
+func TestConfigureLogsAnnotationDeprecation(t *testing.T) {
```
Mathew-Estafanous, line 1015: nit: this test isn't really helpful, we don't test other similar logs.

internal/controller/datadogagent/feature/appsec/feature_test.go
```diff
@@ -593,6 +836,822 @@ func TestAppsecFeatureManageClusterAgentEnabled(t *testing.T) {
...
+			}
+		})
+	}
+}
+
+// TestConfigureAnnotationParsePrecedence pins the CRD-aware parse skip. A malformed
+// annotation whose field the CRD already sets must not fail Configure, because the parsed
+// value would have been discarded by the merge anyway. A malformed annotation on a field
+// the CRD leaves unset must still fail, preserving annotation-only strictness.
+func TestConfigureAnnotationParsePrecedence(t *testing.T) {
```
Mathew-Estafanous: Related to my other comment, this is why I simple CRD > annotation override is a better option. Trying to merge + include precedence is overly complex and easily prone to errors.

internal/controller/datadogagent/feature/appsec/feature_test.go
```diff
@@ -593,6 +836,822 @@ func TestAppsecFeatureManageClusterAgentEnabled(t *testing.T) {
...
+			if tt.assertCfg != nil {
+				tt.assertCfg(t, f.config)
+			}
+		})
+	}
+}
+
+// TestConfigureCRDWinsOverAnnotations pins per-field precedence end to end through
+// Configure: whenever both sources set a field, f.config holds the CRD value.
+func TestConfigureCRDWinsOverAnnotations(t *testing.T) {
```
Mathew-Estafanous, line 1400: Same thought as above

api/datadoghq/v2alpha1/datadogagent_types.go
```diff
@@ -174,6 +180,152 @@ type APMFeatureConfig struct {
...
+type AppsecInjectorSidecarConfig struct {
+	// Sets the sidecar image.
+	// Default: ghcr.io/datadog/dd-trace-go/service-extensions-callout
+	// +optional
+	Image *string `json:"image,omitempty"`
+
+	// Sets the sidecar image tag.
+	// When unset, the tag defaults to a value determined by the cluster-agent image in use.
+	// +optional
+	ImageTag *string `json:"imageTag,omitempty"`
```
khewonc, line 276: Can this be reused? https://github.com/DataDog/datadog-operator/blob/9720c44495c19732b54a695d84ff48cd2e3b67ab/api/datadoghq/v2alpha1/datadogagent_types.go#L1683

internal/controller/datadogagent/feature/appsec/feature.go
eliottness: What do you recommend then? I can't have current users of the annotations upgrade the operator and now their is disabled. I was planning to send them a mail saying this is deprecated and remove this code in the next version of the operator

internal/controller/datadogagent/feature/appsec/feature_test.go
eliottness: Again, the implicit decision you are leaning to is to abandon the 120 orgs that are currently using annotations. Remove annotation is planned but this need to be a proper transition. Sorry if this adds complexity. Let me know if you have something in between in mind

internal/controller/datadogagent/feature/appsec/feature_test.go
eliottness: Keeping this one: the deprecation warning is the only signal annotation users get before removal in v1.32, and it's emitted before the early returns on purpose so disabled and version-gated users still see it — this test is what guards that placement.

internal/controller/datadogagent/feature/appsec/feature_test.go
eliottness: Same answer as the thread above — happy to simplify if we can find an in-between that doesn't silently drop the orgs currently on annotations.

api/datadoghq/v2alpha1/datadogagent_types.go
eliottness: Not reusing it here: this sidecar is rendered by the cluster-agent, not the operator, so we only forward `image`/`imageTag` as two env vars and never call `AssembleImage` — `AgentImageConfig` would publish `jmxEnabled`/`pullPolicy`/`pullSecrets` that we silently ignore, and its non-pointer `Name`/`Tag` would break the non-nil-pointer "is set" rule the annotation fallback relies on.

internal/controller/datadogagent/feature/appsec/feature.go
Mathew-Estafanous: You should do what ADP and other products do. Keep support for annotations as is but if a customer switches on appsec from the CRD, then it should completely override the annotation instead of this merging logic where you're trying to merge the two configurations together. It's confusing both in code and for users if they end up in this hybrid annotation & CR configuration. It should be one or the other.

internal/controller/datadogagent/feature/appsec/feature_test.go
Mathew-Estafanous: No you shouldn't remove annotation support outright. You should support both **separately** so the entire config is defined as either a CRD config or annotation, not both. Then in a few versions you can mark the annotation as deprecated then remove annotations completely.

Feel free to look at ADP's feature and how they've done it.

internal/controller/datadogagent/feature/appsec/feature.go
eliottness: @Mathew-Estafanous  Checked ADP before answering: IsDataPlaneEnabled in feature/utils/utils.go:117 does exactly what this PR does: CRD wins when set, annotation is the fallback, per field. ADP looks simpler because it has one annotation (adp-enabled) mapping to one bool, so per-field and whole-block collapse to the same thing; its only other CRD field is CRD-only and can't read annotations at all. AppSec has 18 annotations across nested structs, which is where the two models actually diverge. If you want the hybrid state gone I'd rather make it loud than silent. That is: reject at validation when both sources are present, but silently dropping 17 annotation-set fields because someone set injector.enabled: true is the failure mode I'm trying to avoid on a security feature.

internal/controller/datadogagent/feature/appsec/feature_test.go
eliottness: Thanks for the clarification, but ADP only has one annotation so override and separation are the same for them, I can't find any other places where annotations are used as a all-or-none fallback. Have I missed one ?

internal/controller/datadogagent/feature/appsec/feature.go
Mathew-Estafanous: I don't see this as a 'failure mode.' Any annotation based configuration is inherently a beta/preview capability and is expected to stop working at any point. The CRD spec is the intended configuration path so customers switching over to the spec should do so in its entirety. Say you keep the merging logic, what happens when it comes time to remove the annotation? This 'failure mode' can *potentially* happen in that case too, but since it's an annotation we treat it as an acceptable risk.

On your point about ADP. The `IsDataPlaneEnabled` method explicitly outlines the CRD > annotation precedence that I'm referring to. To me, appsec having even more fields makes this even more of a reason to have one or the other, not this hybrid merging logic.

tbavelier: Agreed with Mathew here (and our whole team too), we don't believe in a deep merging logic other than simply enabling the feature or not. Annotations are meant to be explicitly configured as part of private, and removed to migrate to the fully fledged feature when it's ready/the CRD won't be changing as much as it could during the annotation part, and customers using the preview should be informed of the upcoming change to take the "breaking" action.

eliottness: Conceded — done in 6aade983. The CRD and the annotations are mutually exclusive now: when `spec.features.appsec.injector` is present it defines the whole configuration and no annotation is read, including for fields it leaves unset; when it's absent the annotations are the only source and any malformed one still rejects the feature. The exclusivity is symmetric, so an annotation no longer completes a CRD that `mode: external` leaves invalid either. `mergeInjectorConfig` is gone, along with the CRD-aware parse skips it needed — net -754 lines.

One addition on top of your suggestion: since a partial migration now silently drops annotation values and this CRD has no validating webhook to reject it at apply time, I added a `DeprecatedConfigInUse` status condition so the deprecation is visible on the object rather than only in operator logs. It only exists while a deprecated surface is in use and is removed once the migration completes, and the check list is generic so other deprecations can reuse it.

internal/controller/datadogagent/feature/appsec/feature_test.go
eliottness: Conceded — see the thread above; 6aade983 drops the merge entirely in favour of all-or-none.

internal/controller/datadogagent/feature/appsec/feature_test.go
eliottness: Same — conceded in 6aade983.

docs/deprecated_configs.md
eliottness: @RomainMuller heads-up: this reverses the precedence model you reviewed here. Following @Mathew-Estafanous and @tbavelier, the CRD and the annotations are now **mutually exclusive** rather than merged per field — 6aade983.

Concretely, the section you asked for is gone: there is no per-field resolution to document any more, and the `proxies: []` caveat no longer applies because there is no cross-source merge for an empty list to fail to clear (I dropped it from the API field too). `docs/deprecated_configs.md` now documents all-or-none, warns that a partial migration silently drops annotation values, and adds a `DeprecatedConfigInUse` status condition so the deprecation is visible via `kubectl describe` rather than only in operator logs. Flagging since you approved on the old semantics.

Mathew-Estafanous (approved):
looks good! thanks for addressing my feedback.

PR #3374, opened 2026-08-20
fix(clusterchecksrunner): preserve packaged conf.d assets (e.g. SNMP …

frank-spano:
…profiles) on CLC runners (CONS-8516)

### What does this PR do?

Adds an init container to the cluster-checks-runner that seeds the conf.d overlay with the agent image's packaged check assets (e.g. SNMP default_profiles), then removes only *.yaml.default so default core checks still don't run. This preserves out-of-the-box profiles on Operator CLC runners—aligning with Helm #1594—so SNMP/NDM cluster checks can autodetect profiles without a manual init-container workaround.

### Motivation

https://datadoghq.atlassian.net/browse/CONS-8516

### Additional Notes

Manually tested - 

Ran the patched operator locally (go run ./cmd/main.go) against a minikube cluster and applied a DatadogAgent with features.clusterChecks.useClusterChecksRunners: true plus an SNMP cluster check via clusterAgent.extraConfd. On the resulting cluster-checks-runner pod I confirmed the new init-copy-check-assets init container ran and that /etc/datadog-agent/conf.d/snmp.d/default_profiles is now populated (e.g. _arista.yaml, _aruba-base.yaml, …), while find /etc/datadog-agent/conf.d -name '*.yaml.default' returns 0 — so packaged profiles are preserved and default core checks are still pruned. In agent status, the previous profile errors (failed to read profile dir … no such file or directory / no profiles found for sysObjectID) are gone; the SNMP check now proceeds past profile autodetection and only fails on device reachability (request timeout), which is expected since there's no real SNMP device at the test IP. This confirms the fix restores OOTB SNMP profiles on Operator CLC runners. Unit tests and the render golden files were also updated and pass.

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist


internal/controller/testutils/renderer/testdata/golden/comprehensive-baseline.golden.yaml
```diff
@@ -2358,6 +2358,25 @@ spec:
         - mountPath: /checks.d
           name: checksd
           readOnly: true
+      - args:
+        - |-
```
levan-m, line 2363: This will lead to errors when simply upgrading on GKE Autopilot so first [allowlist](https://github.com/ddoghq/datadog-gke-workload-allowlist/blob/main/datadog/datadog-datadog-daemonset-exemption-v1.0.5.yaml#L638-L641) needs to be updated.

frank-spano: @levan-m The init-copy-check-assets init container only lands on the cluster-checks-runner Deployment, not the node agent DaemonSet. That Deployment sets no host namespaces, so it doesn't match the datadog-datadog-daemonset-exemption allowlist's hostPID: true criteria, and there's no CLC-runner WorkloadAllowlist. The CLC runner already runs a bash init-config init container on Autopilot today; this new one is the same shape (agent image, readOnlyRootFilesystem, allowPrivilegeEscalation: false, emptyDir only), so it should run under Autopilot's default policy without an exemption. Did you have a specific rule in mind that it would trip, or a cluster where you saw the rejection? Happy to add it to the allowlist + bump the synchronizer version if you can confirm it's actually required.

levan-m: sorry, somehow assumed with was daemonset. Deployments aren't affected by allowlist.

levan-m (approved):

PR #3376, opened 2026-08-20
Add cluster_id tag in remote service callback

levan-m:
### What does this PR do?

CONTP-2018

### Motivation

Make FA experiment setup simpler by making`cluster_name` optional so users don't have to set `DD_CLUSTER_NAME` in operator deployment.  

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

Enable FA experiments without setting `DD_CLUSTER_NAME` in Operator deployment and confirm config editing is enabled in FA ( dd-source #61699).

### Checklist

### DIFF
--- a/pkg/remoteconfig/updater.go
+++ b/pkg/remoteconfig/updater.go
@@ -239,7 +239,7 @@ func (r *RemoteConfigUpdater) Start(apiKey string, site string, clusterName stri
 		"",
 		r.serviceConf.baseRawURL,
 		r.serviceConf.hostname,
-		func() []string { return []string{"cluster_name:" + r.serviceConf.clusterName} },
+		func() []string { return r.clusterTags(context.Background()) },
 		r.serviceConf.telemetryReporter,
 		r.serviceConf.agentVersion,
 		service.WithAPIKey(apiKey),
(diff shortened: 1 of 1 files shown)


tbavelier (approved):

pkg/remoteconfig/updater.go
```diff
@@ -239,7 +239,7 @@ func (r *RemoteConfigUpdater) Start(apiKey string, site string, clusterName stri
 		"",
 		r.serviceConf.baseRawURL,
 		r.serviceConf.hostname,
-		func() []string { return []string{"cluster_name:" + r.serviceConf.clusterName} },
+		func() []string { return r.clusterTags(context.Background()) },
```
khewonc, line 242: non-blocking nit for a future pr: we should look into passing context into setup and start methods to use here so this can't become blocking if the request is slow/hanging

khewonc (approved):

pkg/remoteconfig/updater.go
levan-m: agree, how we handle context in various .Start function is all over the place. Will create a card.

PR #3400, opened 2026-08-25
Empty objectives are fine for DPA manifest

ADKosm:
Remove `MinItems=1` validation from the Objectives field on both DatadogPodAutoscaler and DatadogPodAutoscalerClusterProfile. This allows users to define a DPA with an empty objectives array, which is needed when all scaling configuration comes from the custom-recommender annotation settings or vertical-only autoscalers.


vboulineau:
I thought we already did it, make sure the Cluster Agent requires an objective for horizontal or multi-dim.

Note: you need to use the proper versions of tools to generate (wipe `bin` and use the Makefile).

vboulineau (approved):

config/rbac/role.yaml
```diff
@@ -105,6 +105,9 @@ rules:
   - apiservices
   verbs:
   - '*'
+  - deletecollection
```
vboulineau: Unrelated change?

ADKosm: oh my, yeah, some mess after generate I suppose 🤔 
I will clean up shortly

ADKosm:
@vboulineau for now, we agreed with @clamoriniere that we will rely on default of 75% on CPU set by backend - https://dd.slack.com/archives/C08FUDH30EA/p1787778072416079?thread_ts=1787755494.894189&cid=C08FUDH30EA
We can reconsider this later

levan-m (approved):

PR #3403, opened 2026-08-26
fix(cspm): set compliance config env vars on all node agent containers

lebauce:
### What does this PR do?

Add DD_COMPLIANCE_CONFIG_ENABLED, DD_COMPLIANCE_CONFIG_CHECK_INTERVAL, DD_COMPLIANCE_CONFIG_HOST_BENCHMARKS_ENABLED and
DD_COMPLIANCE_CONFIG_RUN_IN_SYSTEM_PROBE to every container of the pod so that the Core Agent snapshot carries the real values.

HOST_ROOT stays on the target container only: it is not a DD_ config setting, so config streaming never touches it, and it is only meaningful where the host root volume is mounted.

### Motivation

When config streaming is enabled, the consumer container drops its own env layer and replays the Core Agent's config snapshot. The CSPM feature only set the compliance settings on the container actually running the checks, so the Core Agent's snapshot carried the defaults and streamed them back: compliance_config.run_in_system_probe reverted to false, system-probe skipped StartCompliance, and every datadog.security_agent.compliance.* metric disappeared without any error being logged.

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist

### DIFF
--- a/internal/controller/datadogagent/feature/cspm/feature.go
+++ b/internal/controller/datadogagent/feature/cspm/feature.go
@@ -313,12 +313,17 @@ func (f *cspmFeature) ManageNodeAgent(managers feature.PodTemplateManagers) erro
 	VolMgr.AddVolume(&groupVol)
 
 	// env vars
+	// The compliance settings are added to every container of the pod, not only to the one
+	// actually running the checks. When config streaming is enabled, the consumer container
+	// drops its own env layer and replays the Core Agent's config snapshot; if the Core Agent
+	// doesn't carry these settings, they silently revert to their default values.
 	enabledEnvVar := &corev1.EnvVar{
 		Name:  DDComplianceConfigEnabled,
 		Value: "true",
 	}
 	managers.EnvVar().AddEnvVarToContainers([]apicommon.AgentContainerName{apicommon.CoreAgentContainerName, targetContainer}, enabledEnvVar)
 
+	// HOST_ROOT is not a config setting, it only makes sense where the host root volume is mounted.
 	hostRootEnvVar := &corev1.EnvVar{
 		Name:  common.DDHostRootEnvVar,
 		Value: common.HostRootMountPath,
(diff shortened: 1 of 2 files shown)


internal/controller/datadogagent/feature/cspm/feature.go
```diff
@@ -313,12 +313,17 @@ func (f *cspmFeature) ManageNodeAgent(managers feature.PodTemplateManagers) erro
...
+	// The compliance settings are added to every container of the pod, not only to the one
+	// actually running the checks. When config streaming is enabled, the consumer container
+	// drops its own env layer and replays the Core Agent's config snapshot; if the Core Agent
+	// doesn't carry these settings, they silently revert to their default values.
 	enabledEnvVar := &corev1.EnvVar{
 		Name:  DDComplianceConfigEnabled,
 		Value: "true",
 	}
-	managers.EnvVar().AddEnvVarToContainers([]apicommon.AgentContainerName{apicommon.CoreAgentContainerName, targetContainer}, enabledEnvVar)
+	managers.EnvVar().AddEnvVar(enabledEnvVar)
```
tbavelier: This will increase the Agent pod size which is already quite big, by adding the env var to un-related containers such as trace-agent. Can we instead always add the env var to core Agent / wherever it's needed both when configstream is enabled and disabled ?

lebauce: Yes, I can do that but I'm not sure to understand why it would increase the Agent pod size.

tbavelier: Sorry I was not clear, I meant the pod size in ETCD, not the binary size. The whole pod spec is saved/stored in ETCD as a "string" and on large clusters with thousands of Agent pods, every tiny bit adds up (each env var on each container) and it also improves readability to not have un-related env vars on un-related containers.

lebauce: Got it, thanks for the clarification. I updated the PR

tbavelier (approved):
Looks good for cont-platform side, @DataDog/agent-security to validate feature wise this is expected/proper (just want to make sure we don't run CSPM twice, both on core Agent and system-probe in some weird scenarios)

spikat (approved):

YoannGh (approved):
Do we need to do the same for `DD_RUNTIME_SECURITY_CONFIG_*` env vars?

PR #3412, opened 2026-08-31
Fix concurrent access to enabled features in metrics forwarder

carlosrodfern:
### What does this PR do?

The metrics forwarding loop could iterate over `EnabledFeatures` while a reconciliation updated the map, causing a fatal concurrent map iteration and map write error.

This commit clones the map under a read lock and makes the existing code iterate over the snapshot, avoiding concurrent access without holding the lock while sending metrics.

### Motivation

Observed `concurrent map iteration and map write` errors causing the operator to crash.

### Additional Notes

This fix doesn't fix the potential of map value slices concurrent modifications, however, that scenario is currently not present in the code as the values are set as whole slices and guarded by the `setEnabledFeatures(..)` function.

While browsing the code, I noticed that the `EnabledFeatures` map only holds one key, which is the `id` of this forwarder. This id is set only at creation time and the no longer changed. This creates an opportunity for simplification here, by turning `EnabledFeatures` into a slice instead.

### Describe your test plan

### Checklist

### DIFF
--- a/pkg/controller/utils/datadog/metrics_forwarder.go
+++ b/pkg/controller/utils/datadog/metrics_forwarder.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"hash/fnv"
+	"maps"
 	"os"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


tbavelier (approved):

tbavelier:
Thank you for the contribution!

PR #3413, opened 2026-09-01
Fix concurrent access to enabled features in metrics forwarder

tbavelier:
## Summary

Apply the fix from #3412 to snapshot `EnabledFeatures` under a read lock before iterating and sending feature metrics. This prevents concurrent map iteration/write crashes without holding the lock during network calls.

## Validation

- `make build`
- `make ci-test`
- `make update-golang`
- All commands completed successfully with no git diffs.


tbavelier:
Closing, was only done to trigger CI on the contributor commit

PR #3429, opened 2026-09-03
Drop Retries for Permanent Failures on Resource-Creating Controllers

### LINKED ISSUE #3422: DatadogMonitor/DatadogSLO controllers retry permanently-invalid (4xx) CRs every 5s forever, contributing to goroutine growth and operator crash loops
### Pre-submission Checklist

### Operator version

Reproduced on both `v1.26.0` and, after upgrading, `v1.29.0` (latest stable at the time) — so the version bump alone did not fix it.

Corresponding chart versions `2.22.0` and `2.25.1`.

### Bug Report

**TL;DR:** `DatadogMonitor` and `DatadogSLO` reconciles that fail with a permanent (4xx-style) validation error get requeued on the exact same fixed, non-backing-off `defaultErrRequeuePeriod = 5 * time.Second` as a transient (5xx/network) error. There's no distinction between "will never succeed without a spec change" and "worth retrying soon." Over a few hours this correlates with steady goroutine growth that eventually breaches the `maximumGoroutines` health-check guard, `/healthz` starts returning 500, kubelet kills the container on the failed liveness probe, and the operator pod crash-loops — degrading every other `DatadogMonitor`/`DatadogSLO` on the cluster, not just the misconfigured ones.

**What happened:**

We had a `datadog-operator` pod stuck in a liveness-probe crash loop. Kubernetes events showed the classic cycle:

```
Unhealthy: Liveness probe failed: HTTP probe failed with statuscode: 500
Killing: ... failed liveness probe
BackOff
```

and `kubernetes_state.container.restarts` climbed steadily over time rather than staying flat.

Tracing it back through our logs, we found a handful of `DatadogMonitor`/`DatadogSLO` custom resources that were failing validation on *every single reconcile attempt* — one `DatadogMonitor` missing a required tag, one `DatadogSLO` with an invalid field type, and one `DatadogSLO` with invalid query syntax. Each produced its own error (`error creating SLO`, `error validating monitor`, etc.) roughly every 5-10 seconds, continuously, for hours. None of these CRs could ever succeed without an actual spec fix — but the controller kept retrying at the same fixed 5s cadence as it would for a transient API/network hiccup.

We fixed those three CRs, which stopped our immediate crash loop, but the underlying mechanism that turned "a few misconfigured CRs" into "the whole shared operator pod crash-loops for every Monitor/SLO on the cluster" is a general operator behavior, not specific to those CRs' content.

**Root cause:**

Both the `DatadogMonitor` and `DatadogSLO` controllers requeue on *any* reconcile error — including a permanent validation failure that can never succeed without a spec change — using a hardcoded, non-backing-off period:

[`internal/controller/datadogmonitor/controller.go#L47`](https://github.com/DataDog/datadog-operator/blob/3e86172a64a54e4129781464e43dc69c5bd08a5e/internal/controller/datadogmonitor/controller.go#L47):
```go
defaultErrRequeuePeriod        = 5 * time.Second
```

applied identically regardless of error type at multiple call sites, e.g. [`controller.go#L257-L263`](https://github.com/DataDog/datadog-operator/blob/3e86172a64a54e4129781464e43dc69c5bd08a5e/internal/controller/datadogmonitor/controller.go#L257-L263):
```go
if err = r.create(auth, logger, instance, newStatus, now, instanceSpecHash); err != nil {
    logger.Error(err, "error creating monitor")
    result.RequeueAfter = defaultErrRequeuePeriod
}
} else {
    err = fmt.Errorf("monitor type %v not supported", instance.Spec.Type)
    logger.Error(err, "error creating monitor")
    result.RequeueAfter = defaultErrRequeuePeriod
}
```

and the same pattern in [`internal/controller/datadogslo/controller.go#L43`](https://github.com/DataDog/datadog-operator/blob/3e86172a64a54e4129781464e43dc69c5bd08a5e/internal/controller/datadogslo/controller.go#L43) / [`controller.go#L177-L179`](https://github.com/DataDog/datadog-operator/blob/3e86172a64a54e4129781464e43dc69c5bd08a5e/internal/controller/datadogslo/controller.go#L177-L179):
```go
if err != nil {
    result.RequeueAfter = defaultErrRequeuePeriod
}
```

There's no classification of the underlying Datadog API error (4xx permanent vs. 5xx/network transient) anywhere in this path, and no exponential backoff — a permanently broken CR gets hammered every 5 seconds, forever, until someone notices and fixes its spec.

**Related issues:**

- [#2803](https://github.com/DataDog/datadog-operator/issues/2803) hit the same downstream symptom (goroutine growth → `maximumGoroutines` breach → liveness-probe crash loop) from a different source (a metrics-forwarder goroutine leak, since fixed in `81fef5635d40242330cffbfb4715a11bc4db1eae`). We reproduced on operator versions that post-date that fix, so this looks like a distinct contributor to the same failure class rather than a regression of it — most likely the tight, indefinite 5s retry loop described above.
- [#2816](https://github.com/DataDog/datadog-operator/issues/2816) independently flagged that the fixed periodic requeue interval causes reconcile-queue problems at CR scale, which lines up with this being a recurring pain point in the requeue-period design rather than a one-off.

**Suggested direction** (not attaching a PR, since we're not certain which approach the team prefers):

- Classify Datadog API client errors as permanent (4xx) vs. transient (5xx/network).
- For permanent errors, apply exponential backoff, or stop auto-retrying entirely until the CR's spec hash changes, instead of the fixed 5s loop forever.
- Surface a clear terminal status/condition and a Kubernetes Event for a permanently-invalid CR, so cluster operators can find and alert on it before it degrades the shared operator pod for every other CR on the cluster.

Happy to share further logs or data if it would help narrow this down further.

gabedos:
### What does this PR do?

Whenever resource creating controllers (Monitors/SLOs/Dashboards/DDGR) reconciles that fail with a permanent (4xx) API error now back off to the existing force-sync cadence (default 60m) instead of retrying every 5s forever. Transient errors (5xx, network, 429) keep the fast 5s retry.

**4xx** responses are **client-side errors**. The request itself was invalid (bad query, bad field, etc.), so retrying the exact same request isn't going to produce a different result. It's a permanent failure until the spec changes, so hammering it every 5s is for nothing.

### Motivation

Fixes #3422: a handful of misconfigured `DatadogMonitor`/`DatadogSLO` resources could get retried every 5s indefinitely which is noisy.

### Additional Notes

In the future, we could consider adding k8s events to better track these instances of CRD that perma fail.

### Describe your test plan

Deploy Operator with a malformed CRD and witness only 1 error log:

```
kind: DatadogMonitor
metadata:
  name: permanent-error-test
  namespace: system
spec:
  query: "this is not a valid datadog monitor query at all !!!"
  type: "metric alert"
  name: "Permanent error test monitor"
  message: "This monitor intentionally has an invalid query to test permanent-error backoff."
  tags:
    - env:staging
```

```
k logs datadog-operator-manager-6d6c7bb4fb-c5vsj -n system | grep controllers.DatadogMonitor
{"level":"INFO","ts":"2026-09-03T16:13:11.974Z","logger":"controllers.DatadogMonitor","msg":"Setting monitor requeue period","duration":"1m0s"}
{"level":"INFO","ts":"2026-09-03T16:14:24.465Z","logger":"controllers.DatadogMonitor","msg":"Reconciling DatadogMonitor","datadogmonitor":{"name":"permanent-error-test","namespace":"system"}}
{"level":"INFO","ts":"2026-09-03T16:14:24.465Z","logger":"controllers.DatadogMonitor","msg":"Object does not have a finalizer; adding finalizer","datadogmonitor":{"name":"permanent-error-test","namespace":"system"},"kind":"DatadogMonitor","datadogID":"0","finalizername":"finalizer.monitor.datadoghq.com"}
{"level":"INFO","ts":"2026-09-03T16:14:24.472Z","logger":"controllers.DatadogMonitor","msg":"Reconciling DatadogMonitor","datadogmonitor":{"name":"permanent-error-test","namespace":"system"}}
{"level":"INFO","ts":"2026-09-03T16:14:24.475Z","logger":"controllers.DatadogMonitor","msg":"Added required tags","datadogmonitor":{"name":"permanent-error-test","namespace":"system"},"Monitor Namespace":"system","Monitor Name":"permanent-error-test","Monitor ID":0}
{"level":"INFO","ts":"2026-09-03T16:14:24.479Z","logger":"controllers.DatadogMonitor","msg":"Reconciling DatadogMonitor","datadogmonitor":{"name":"permanent-error-test","namespace":"system"}}
{"level":"ERROR","ts":"2026-09-03T16:14:24.622Z","logger":"controllers.DatadogMonitor","msg":"error creating monitor","datadogmonitor":{"name":"permanent-error-test","namespace":"system"},"error":"error validating monitor: 400 Bad Request: {\"errors\":[\"The value provided for parameter 'query' is invalid\"]}","stacktrace":"github.com/DataDog/datadog-operator/internal/controller/datadogmonitor.(*Reconciler).internalReconcile\n\t/workspace/internal/controller/datadogmonitor/controller.go:257\ngithub.com/DataDog/datadog-operator/internal/controller/datadogmonitor.(*Reconciler).Reconcile\n\t/workspace/internal/controller/datadogmonitor/controller.go:117\ngithub.com/DataDog/datadog-operator/internal/controller.(*DatadogMonitorReconciler).Reconcile\n\t/workspace/internal/controller/datadogmonitor_controller.go:52\nsigs.k8s.io/controller-runtime/pkg/reconcile.(*objectReconcilerAdapter[...]).Reconcile\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.3/pkg/reconcile/reconcile.go:169\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Reconcile\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.3/pkg/internal/controller/controller.go:222\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.3/pkg/internal/controller/controller.go:479\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.3/pkg/internal/controller/controller.go:438\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func1.1\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.3/pkg/internal/controller/controller.go:313"}
```

### Checklist


internal/controller/datadogslo/slo.go
```diff
@@ -157,26 +168,29 @@ func deleteSLO(auth context.Context, client *datadogV1.ServiceLevelObjectivesApi
 		Force: &force,
 	}
 	_, localVarHTTPResponse, err := client.DeleteSLO(auth, sloID, optionalParams)
+	if localVarHTTPResponse != nil {
```
levan-m, line 172: not introduced by PR but If we do nil check here, we should do it on 175, 177 as well.

internal/controller/datadoggenericresource/dashboards.go
```diff
@@ -51,47 +51,59 @@ func (h *DashboardHandler) refreshState(_ context.Context, _ *v1alpha1.DatadogGe
...
+		return datadogV1.Dashboard{}, translateClientError(err, httpResp, "error getting dashboard")
 	}
 	return dashboard, nil
 }
 
 func createDashboard(auth context.Context, client *datadogV1.DashboardsApi, instance *v1alpha1.DatadogGenericResource) (datadogV1.Dashboard, error) {
 	dashboardCreateData := &datadogV1.Dashboard{}
 	if err := json.Unmarshal([]byte(instance.Spec.JsonSpec), dashboardCreateData); err != nil {
-		return datadogV1.Dashboard{}, translateClientError(err, "error unmarshalling dashboard spec")
+		return datadogV1.Dashboard{}, translateClientError(err, nil, "error unmarshalling dashboard spec")
```
levan-m: we are using `translateClientError` for unmarshalling error on user input. What would be the code for such error and how it will be handled?

pkg/controller/utils/apierror.go
```diff
@@ -0,0 +1,44 @@
...
+func (e *APIError) Unwrap() error { return e.err }
+
+// IsPermanentAPIError reports whether err is a 4xx (client-side, non-retryable)
+// error, excluding 429 which is worth retrying after a backoff.
+func IsPermanentAPIError(err error) bool {
+	var apiErr *APIError
+	if !errors.As(err, &apiErr) {
+		return false
+	}
+	return apiErr.StatusCode >= 400 && apiErr.StatusCode < 500 && apiErr.StatusCode != http.StatusTooManyRequests
```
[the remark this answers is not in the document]
levan-m: I'm inclined to leave as is. auth failure would more likely be resolved by api/app key update which needs operator restart anyway.

internal/controller/datadogslo/slo.go
gabedos: 3b20f17e37f98845155bf7255d1a50cc5c38c421 added this in because CI was complaining about `body.Close` lint errors

This one condition should be enough. If there is an actual `localVarHTTPResponse ` then we defer the body close so it gets picked up in 175 and 177 exits. If `localVarHTTPResponse` is nil then there's nothing to close.

levan-m: I meant checking null before accesssing `localVarHTTPResponse.StatusCode`

pkg/controller/utils/apierror.go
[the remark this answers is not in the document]
gabedos: 🤔 401 Unauthorized & 403 Forbidden _could_ be added into the condition. If the user were to change the permissions of their keys within the Datadog App would an Operator restart actually be needed since it's still the same key? I'm not familiar with how this works. Still, I think it _could_ be fine to fallback to 1hr retry in that case too.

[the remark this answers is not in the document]
levan-m: yea agree, 1h for such errors is fine.

internal/controller/datadoggenericresource/dashboards.go
gabedos: 🤔 Right now a "bad" unmarshal would still be picked up as a transient error since there's no http status code. Since this technically is a client-side issue of providing bad data, I think it could be fair to surface a 400 Bad Request error here. Pushing up a commit in a sec with this approach.

gabedos: [730ffe9](https://github.com/DataDog/datadog-operator/pull/3429/commits/730ffe9cd92354e635bc0bdf6d336a6ea5ce62a3)

levan-m (approved):

PR #3471, opened 2026-09-16
fix(acr): enable ADP check source when ACR is enable

fabbing:
### What does this PR do?

Enable the ADP check source when ACR is enable.

### Motivation

ACR forward checks events to ADP, without the ADP check source they aren't processed.

### Minimum Agent Versions

* Agent: not released yet (dev build)

### Checklist


internal/controller/datadogagent/feature/checkrunner/feature_test.go
```diff
@@ -80,6 +86,11 @@ func Test_checkRunnerFeature(t *testing.T) {
 					}
 					assert.NotContains(t, crEnvVars, checkRunnerEnabledEnvVar, "DD_CHECK_RUNNER_ENABLED is a core agent setting and should not be set on the check runner")
 
+					// ADP must serve the Checks IPC endpoint ACR is pointed at.
+					adpEnvVars := mgr.EnvVarMgr.EnvVarsByC[apicommon.AgentDataPlaneContainerName]
+					assert.Contains(t, adpEnvVars, dataPlaneChecksEnabledEnvVar, "DD_DATA_PLANE_CHECKS_ENABLED should be set on the agent data plane so it serves the Checks IPC source")
+					assert.NotContains(t, agentEnvVars, dataPlaneChecksEnabledEnvVar, "DD_DATA_PLANE_CHECKS_ENABLED is an ADP setting and should not be set on the core agent")
+
```
tobz: We actually _don't_ want to do this: the goal is that all configuration flows through the Core Agent, so we should set these configurations on the Core Agent directly.

tobz (changes requested):
Left a note about where ADP configurations should be placed, container-wise, to ensure the proper flow of configuration data.

internal/controller/datadogagent/feature/checkrunner/feature_test.go
fabbing: Ah, good point! Addressed with: https://github.com/DataDog/datadog-operator/pull/3471/commits/dec4ce06cf6c72776d2a713e2e56b1392b9a8f77.
Then I got bitten by the config schema because it didn't know the keys!

internal/controller/datadogagent/feature/checkrunner/feature.go
```diff
@@ -55,6 +58,10 @@ func (f *checkRunnerFeature) Configure(dda metav1.Object, _ *v2alpha1.DatadogAge
 			IsRequired: &f.enabled,
 			Containers: []apicommon.AgentContainerName{apicommon.AgentCheckRunnerContainerName},
 		}
+
+		if !featureutils.IsDataPlaneEnabled(dda, ddaSpec, f.defaultDataPlaneEnabled) {
+			f.logger.Info("The checkrunner feature requires the dataPlane feature.")
+		}
```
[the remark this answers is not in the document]
tobz: Bot seems right: this should be a hard failure/error if the data plane feature is not enabled.

internal/controller/datadogagent/feature/checkrunner/feature.go
```diff
@@ -88,38 +94,43 @@ func (f *checkRunnerFeature) ManageNodeAgent(managers feature.PodTemplateManager
...
 		Name:  ddCheckRunnerEndpointsIPCEndpoint,
 		Value: "http://localhost:5105",
 	})
+
+	// ADP configuration
+	// Enable Checks IPC source (the endpoint ACR sends events to)
+	managers.EnvVar().AddEnvVarToContainer(coreAgentContainer, &corev1.EnvVar{
+		Name:  common.DDDataPlaneEnabled,
+		Value: "true",
+	})
```
tobz: After making the change to require the data plane feature to be enabled as a dependency of this feature, we won't need to do this anymore.

PR #3479, opened 2026-09-17
feat: add NodeSelector for host-profiler deployments

wehzzz:
### What does this PR do?
Added a `nodeSelector` to target `linux` nodes, as `linux` is the only os supported by the host-profiler.

### Motivation
Follows up on this review comment: https://github.com/DataDog/opentelemetry-examples/pull/372#discussion_r4028565674

### Additional Notes

### Minimum Agent Versions

* Agent: vX.Y.Z
* Cluster Agent: vX.Y.Z

### Describe your test plan

### Checklist


Gandem (approved):

khewonc (approved):

tbavelier:
@wehzzz I'd rather not have host-profiler reach into the shared component reconciler for this.

`SetLinuxNodeSelector` is called from `datadogagent.ComponentRegistry.reconcileComponent`, which is not the path that builds workloads. The DatadogAgent controller only creates `DatadogAgentInternal`s. The live component registry is in `datadogagentinternal`, and even that one only reconciles the Cluster Agent, Cluster Checks Runner, and OTel gateway Deployments. Profile DDAIs skip it entirely. The node Agent — the pod that actually runs host-profiler — is built later in `reconcileV2Agent`.

So this either no-ops, or if it were moved onto the live registry it would stamp `kubernetes.io/os=linux` onto Deployments that do not run host-profiler, and still miss the DaemonSet.

Windows profiles are already handled on that DaemonSet path, and this change does not affect them:

- On `provider=windows`, host-profiler is `Excluded`, so `ManageNodeAgent` is skipped.
- `ApplyWindowsPodTransformation` then strips the container (it is not in `windowsAllowedContainerNames`) and sets `kubernetes.io/os=windows` last.
- Windows nodes also carry the automatic `node.kubernetes.io/os=windows:NoSchedule` taint. The default node-agent DaemonSet does not tolerate it — only the Windows profile path adds that toleration — so the pod would not land on those nodes anyway.

The right place is `ManageNodeAgent`, which only runs for the node-agent pod that includes the container. Set `kubernetes.io/os=linux` there and leave other selector keys intact. Windows profiles never call that hook, so they keep `kubernetes.io/os=windows`.

A node-agent `nodeSelector` override can still replace that key afterward. That is expected: overrides win over feature defaults, so we should not re-apply the selector after overrides to beat that.

tbavelier (changes requested):
See previous comment and suggested approach https://github.com/DataDog/datadog-operator/pull/3508

wehzzz:
Thanks @tbavelier for the explanation and the clarification!

PR #3486, opened 2026-09-18
[CONS-11950] Apply global.registry to the managed CSI driver image

L3n41c:
### What does this PR do?

Makes `spec.global.registry` apply to the Datadog CSI driver image that the operator manages.

- Adds `spec.registry` to `DatadogCSIDriverSpec`, propagated by the DatadogAgent controller from the (already defaulted) `spec.global.registry`.
- Adds `spec.global.csi.image` (a dedicated `CSIImageConfig`: `name`, `tag`, `pullPolicy`, `pullSecrets`), propagated to `spec.csiDriverImage`, so those work while keeping `autoManage: true`. It deliberately does not reuse `AgentImageConfig`, whose `jmxEnabled` only applies to the Agent image — there is no `csi-driver:<tag>-jmx`, so offering the knob could only resolve to a tag that does not exist.
- Wires `imagePullPolicy` from both image specs into the DaemonSet containers. The CRD accepted the field but nothing ever read it.
- Includes `registrarImage.pullSecrets` in the pod's `imagePullSecrets`. Only the driver's secrets were used, so a relocated registrar could not authenticate.
- Keeps GKE Autopilot on GCR in `resolveCSIDriverImage`, using the existing `images.IsGCRRegistry`, so moving the default off hardcoded GCR does not break the Autopilot WorkloadAllowlist.

### Motivation

[CONS-11950](https://datadoghq.atlassian.net/browse/CONS-11950) (FR: [FRCONS-1268](https://datadoghq.atlassian.net/browse/FRCONS-1268)).

`spec.global.registry` is the documented single place a customer points a `DatadogAgent` at another container registry, and it already redirects every Agent image plus the Single Step Instrumentation injection images. The CSI driver ignored it: `resolveCSIDriverImage` hardcoded `gcr.io/datadoghq`. A customer on a private or mirrored registry who set `global.csi.enabled: true` got the CSI DaemonSet in `ImagePullBackOff` against an egress destination their environment blocks by design, with nothing in the docs explaining why.

The only workaround was to opt out of operator management entirely (`autoManage: false`) and hand-own the `DatadogCSIDriver` CR with full `registry/name:tag` image names, because the operator reverts any image edit on the CR it manages.

### Additional Notes

**Reading this diff.** It is ~760 added lines, but only **132 are hand-written source**. The split:

| | Lines |
|---|---|
| Generated (`config/crd/**`, `zz_generated.*`, `docs/configuration*.md`) | ~400 |
| Tests | ~230 |
| Hand-written source | ~132 |

Most of the generated bulk is one 48-line CRD block for `global.csi.image`, replicated across 3 CRDs x 2 formats because `GlobalConfig` is embedded in `DatadogAgent`, `DatadogAgentInternal` and `DatadogAgentProfile`. Reviewing `internal/`, `api/**/*_types.go` and the tests covers everything of substance; the rest is `make generate && make manifests` output.

The `global.registry` field documentation is updated to say it covers Agent images, SSI injection images and the CSI driver image.

### Minimum Agent Versions

None. This only changes which registry the operator resolves images from.

* Agent: n/a
* Cluster Agent: n/a

### Describe your test plan

Unit tests (`go test ./...`) and `make lint` pass. New coverage:

- `internal/controller/datadogcsidriver`: `TestResolveImages_Registry` tables the registry/override precedence (registry only, empty registry, tag-only override, full `registry/name:tag` override winning, registrar unaffected, and the three Autopilot rows); `TestBuildDaemonSet_ImagePullPolicy` and `TestBuildDaemonSet_RegistrarImagePullSecrets` pin the two newly-honored fields, the latter also asserting the `DD_APM_REGISTRY_AUTH_*` fallback still uses driver secrets only.
- `internal/controller/datadogagent`: `TestReconcileDatadogCSIDriver_ImageConfigPropagated` and `_ImageConfigIsCopied` cover propagation and that the built CR does not alias the DDA.

Manual, with the operator started with `--datadogCSIDriverEnabled=true`:

1. `DatadogAgent` with `global.registry: <mirror>` and `global.csi.enabled: true` → `kubectl get ddcsi <name> -o jsonpath='{.spec.registry}'` returns the mirror, and the DaemonSet shows `<mirror>/csi-driver:1.4.0` with the registrar unchanged on `registry.k8s.io/sig-storage`.
2. No `global.registry` → `registry.datadoghq.com/csi-driver:1.4.0`.
3. Set `global.csi.image.pullSecrets` → the pod template's `imagePullSecrets` appears.
4. Change `global.registry` on the DDA → the CR and the DaemonSet converge on the next reconcile.

### Checklist

[CONS-11950]: https://datadoghq.atlassian.net/browse/CONS-11950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brett0000FF (approved):

clamoriniere (approved):

internal/controller/datadogagent/ddcsi.go
```diff
@@ -156,6 +168,21 @@ func (r *Reconciler) buildDesiredDatadogCSIDriver(instance *v2alpha1.DatadogAgen
 	return ddcsi, nil
 }
 
+// csiDriverImageConfig maps the DDA's CSI image configuration onto the AgentImageConfig the
+// DatadogCSIDriver spec takes, leaving JMXEnabled unset (see CSIImageConfig).
+func csiDriverImageConfig(image *v2alpha1.CSIImageConfig) *v2alpha1.AgentImageConfig {
+	imageConfig := &v2alpha1.AgentImageConfig{
+		Name:       image.Name,
+		Tag:        image.Tag,
+		PullPolicy: image.PullPolicy,
```
tbavelier, line 177: `PullPolicy` is a pointer, and this assignment shares it with `image`. `Name` and `Tag` are strings, and `PullSecrets` is copied into a new slice just below, so the built `DatadogCSIDriver` is otherwise a snapshot of the DDA spec.

`TestReconcileDatadogCSIDriver_ImageConfigIsCopied` already relies on that snapshot: it mutates the DDA after `buildDesiredDatadogCSIDriver` and expects the built object to stay unchanged. The same mutation through `*image.PullPolicy` would change `ddcsi.Spec.CSIDriverImage.PullPolicy` too. Copying the value (`ptr.To(*image.PullPolicy)` when non-nil) would close that.

In the reconcile path this does not matter today. Nothing writes through this pointer on either object: the DDA value is only read while building the driver spec, and the CSI controller only reads it to set the container `imagePullPolicy` (a separate value field). The desired object is created or patched immediately, so the alias never changes what gets applied.

tbavelier (approved):
A minor nit based on the test construction, otherwise, LGTM!

internal/controller/datadogagent/ddcsi.go
L3n41c: Good catch, thanks — fixed in d464e1f0. `csiDriverImageConfig` now goes through `CSIImageConfig.DeepCopy()` (matching how `Name`/`Tag`/`PullSecrets` were already handled) instead of hand-copying each field, so `PullPolicy` is detached from the DDA spec too. Extended `TestReconcileDatadogCSIDriver_ImageConfigIsCopied` to mutate `PullPolicy` and `PullSecrets` after construction alongside the existing `Tag` mutation — it fails without the fix.

tbavelier (approved):

